PR #11, opened 2023-03-01
build a SILO Docker image

### LINKED ISSUE #3: Build Docker containers
Build and publish a docker container that contains a ready-to-use SILO instance

fengelniederhammer:
resolves #3 

A possible next step would be to separate the Dockerfile into a build and a runtime container


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  dockerImage:
+    name: 'Build Docker image'
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' }}
```
chaoran-chen, line 11: Why should the workflow be executed on both push and pull_request (lines 4-5) but then, the job only runs on push?

chaoran-chen:
The docker image build works well on my machine :)

.github/workflows/ci.yml
fengelniederhammer: Because this will contain additional jobs for the tests etc. (at least that's what I had in mind.

Issue #16, opened 2023-03-02
Implement reading database data in SILO API


fengelniederhammer:
As a first step: When starting the API, read the data, process it, and load it into the database. This has to be split up in a further step into a preprocessing pipeline and reading a serialized state.

When the relevant steps are migrated, delete the current [`main.cpp`](https://github.com/GenSpectrum/LAPIS-SILO/blob/6908c7f3a6ac85008af3606d0222ea58759ce954/src/main.cpp) and only leave the api main method. (Make sure to remove the dependency on readline)

Issue #15, opened 2023-03-02
Make sure that failed `assert`s don't terminate the API


fengelniederhammer:
Currently the code contains a lot of `assert`s. Those terminate the API upon failure. E.g. reproduce with 

```
docker run -p 8080:8080 silo
curl --request POST --data '{
      "action": {
        "type": "Aggregated"
      },
      "filter": {}
    }' http://localhost:8080/query
```
yields "Assertion failed: js.HasMember("type") (/src/src/query_engine/query_engine.cpp: to_ex: 17)" and the API stops.

-> Make `assert` not terminate or throw exceptions instead.

Issue #14, opened 2023-03-02
Harden the SILO api


fengelniederhammer:
Make the SILO API more resistant to errors:
- /query should only accept POST requests.
- /info should only accept GET requests.
- Make sure that there is proper error handling in place: i.e. if an exception is thrown, the application doesn't terminate, but responds with an appropriate HTTP repsonse.
- The response should always be a valid JSON in a defined format.

PR #17, opened 2023-03-02
implement a Rest interface for SILO

### LINKED ISSUE #2: Expose SILO functionality via a REST interface
Implement a REST interface that one can send SILO queries via a HTTP request to.

@chaoran-chen @Taepper How should we integrate the "update data" step? Should we implement another endpoint that triggers loading new data? Or is there another way to access the data used by a running SILO instance?

fengelniederhammer:
resolves #2 

This is more an MVP than a fully functional solution. It's possible to communicate with SILO via HTTP requests, but:
- handling JSON is quite exhausting in C++. We should think of something how to simplify serializing structs. I didn't find anything premade yet. One always has to build JSON objects by hand. Luckily Poco can serialize those objects into JSON then. #21 
- The interface needs to be hardened: /query should only accept POST requests. If something goes wrong, the application will terminate. #14
- We should not use `assert` as it terminates the application on failure #15
- The current `main.cpp` should be deleted and the relevant content should be migrated to `--processData` of the `api.cpp`. #16
- We should decide what we want to do with the "info_d" command #18 
- We should implement proper logging in a file instead of to stdout #19
- Poco offers quite some options to configure the App (including loading the configuration from a file). We should check whether we need to tweak the App a bit by using more threads and such as [in this example](https://github.com/pocoproject/poco/blob/feb1e24cfdfe0c1e1b636eee8b63f98d95c9f817/Net/samples/HTTPTimeServer/src/HTTPTimeServer.cpp#L189)

Also finding a suitable library for an HTTP interface was quite hard. We decided against boost, since it's too low level. There are a lot of outdated and unmaintained frameworks around, and there are some alternatives in C which also seem very low level. Poco seemed to be the only viable alternative

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -97,17 +97,6 @@ int handle_command(Database& db, std::vector<std::string> args) {
       }
       auto sequence_out = args.size() > 3 ? std::ofstream(args[3]) : std::ofstream(default_sequence_input + ".repair");
       prune_sequences(meta_input, sequence_input.get_is(), sequence_out);
-   } else if ("info" == args[0]) {
-      if (args.size() > 1) {
-         std::ofstream out(args[1]);
-         if (!out) {
-            std::cout << "Could not open outfile " << args[1] << endl;
-            return 0;
-         }
-         db.db_info(out);
-      } else {
-         db.db_info(std::cout);
-      }
    } else if ("info_d" == args[0]) {
       if (args.size() > 1) {
          std::ofstream out(args[1]);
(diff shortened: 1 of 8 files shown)


fengelniederhammer:
- [ ] clean up commits before merge

src/main.cpp
```diff
@@ -104,9 +104,9 @@ int handle_command(Database& db, std::vector<std::string> args) {
             std::cout << "Could not open outfile " << args[1] << endl;
             return 0;
          }
-         db.db_info(out);
+//         db.get_db_info(out);
```
fengelniederhammer, line 107: - [ ] Check, probably delete

fengelniederhammer:
Maybe for additional reference - we took inspiration from those examples:
https://github.com/edson-a-soares/poco_restful_webservice
https://github.com/pocoproject/poco/blob/master/Net/samples/HTTPFormServer/src/HTTPFormServer.cpp
https://github.com/pocoproject/poco/blob/master/Net/samples/HTTPTimeServer/src/HTTPTimeServer.cpp

JonasKellerer (approved):

PR #24, opened 2023-03-06
use nlohmann_json to serialize the query response

### LINKED ISSUE #21: Simplify serializing/deserializing JSON
We can use https://github.com/nlohmann/json instead of Poco::JSON. That should simply handling JSON.

fengelniederhammer:
resolves #21 

I have a couple of questions on which I'd like to have early feedback before fully implementing it, so here is a draft PR.

* Do we really want to have a single endpoint to which we send all queries? We lose typing of request and response: The query action is one of (Aggregated, Mutations, ...) and the response is one of (AggregatedResponse, MutationsResponse, ...). The interface can't communicate properly that we know exactly which of the responses you get for which request. The interface instead states something like "send any query and you might get any response". 
E.g. the interface technically doesn't forbid to return a mutations response on a aggregated query. By implementing separate endpoints for each query, we rescue the mapping.
* The response currently contains performance measures. Do we want to keep it that way? That makes the tests harder to write as we can't simply compare them to a hardcoded expectation (see failing end to end test).
```
  {
    "actionTime": 9,
    "filterTime": 0,
    "parseTime": 12,
    "returnMessage": {
      "count": 0
    }
  }
```
* Are you ok with renaming the fields of the structs such that they comply the usual camelCase convention in JSON? This simplifies mapping to JSON, since we can use that macro from nlohman_json to generate functions for the mapping.


fengelniederhammer:
Answers:
* Regarding typing and multiple endpoints: We sacrifice typing for extendability. We keep a single endpoint.
* Performance measures are useful (at least in the beginning) and we'd like to keep them.

Taepper (commented):
Looks great! You are really moving along quickly!

fengelniederhammer:
This is now ready for review. Unfortunately it was relatively hard to write unit tests for the request handlers, we didn't manage to mock the request and response, so we left this for a later task.

Also as a follow up, we should restructure the code to better distinguish between silo and it's web api.

Taepper (approved):

src/query_engine/query_engine.cpp
```diff
@@ -822,48 +824,45 @@ silo::result_s silo::execute_query(const silo::Database& db, const std::string&
...
-               ret.return_message += "{\"mutation\":\"";
-               ret.return_message += s.mut_from;
-               ret.return_message += std::to_string(s.position);
-               ret.return_message += s.mut_to;
-               ret.return_message += "\",\"proportion\":" + std::to_string(s.proportion) +
-                  ",\"count\":" + std::to_string(s.count) + "},";
-            }
+
+            std::vector<response::MutationProportion> mutationProportions(mutations.size());
+            std::transform(mutations.begin(), mutations.end(), mutationProportions.begin(), [](mutation_proportion mutation_proportion) {
```
JonasKellerer, line 852: Could we rename the struct mutation_proportion? (Maybe to MutationProportion) This looks like errors are down to be happening, because it is unclear if the variable or the struct function is called.

PR #29, opened 2023-03-07
reduce size of Docker images

### LINKED ISSUE #12: Reduce the size of the Docker images
By including the whole libraries in the image, the container is about 500 MB in size. Especially boost is probably big.

-> Split into a build time and a runtime container.

fengelniederhammer:
resolves #12 

previously: ~500 MB, now ~40MB
Also use Conan in the build stage of the Dockerfile and statically link libraries as far as possible. Only TBB strongly advices against statically linking it, so we kept the dynamic link there.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,8 @@
 *
 !CMakeLists.txt
+!build_with_conan.sh
+!conanprofile.docker
+!conanfile.txt
 !src/
 !include/
 !testBaseData/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event_name == 'push' }}
     steps:
-      - uses: actions/checkout@v3
-      - name: Docker metadata
+      -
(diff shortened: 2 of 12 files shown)


chaoran-chen:
Cool! Am I understanding it correctly that we now have a basically self-sustained and small (<40 MB) executable and that TBB is the only package that needs to be installed separately?

JonasKellerer:
Yes. That is the idea. Boost is so large that we would not want to have it in the image.

chaoran-chen:
And by statically linking boost, it gets smaller? Are only the functions of boost that we use embedded or how does it work? (sorry for the very naive questions)

JonasKellerer:
In general terms yes. The linker can throw out unused symbols that are not needed, which are many in the case for boost. However, if there are many transitive dependencies this could also make the image bigger, because all of them have to be included in the artifact, which is gladly not the case for us.

fengelniederhammer:
We also statically linked the C++ standard library and a gcc library. I've read that this might yield portability issues when executing it on a different OS. But as long as we just shift the executable between Docker images, we figured that this is the best solution.

JonasKellerer (approved):

Issue #27, opened 2023-03-07
Naming convention


JonasKellerer:
Add naming convention, naming convention enforcement and refactor code to comply to the convention.

PR #34, opened 2023-03-09
Add formatter and corresponding ci

JonasKellerer:
Part of #27


src/example.test.cpp
```diff
@@ -1,18 +1,15 @@
...
 class ClassToTest {
-   public:
-   virtual void functionToMock() {
-   }
-   virtual void callMockedFunction() {
-      functionToMock();
-   }
+  public:
+   virtual void functionToMock() {}
+   virtual void callMockedFunction() { functionToMock(); }
```
fengelniederhammer, line 8: Maybe not so important here, but do we really want to allow method declaration and implementation on the same line?

src/query_engine/query_engine.cpp
```diff
@@ -86,18 +102,20 @@ std::unique_ptr<BoolExpression> parse_expression(const Database& db, const rapid
...
       unsigned pos = js["position"].GetUint();
       char ref_symbol = db.global_reference[0].at(pos);
       /// this <= is correct! the negation would flip the exact bit from -1 to +1 and vice versa
-      if (exact <= 0) { /// NucEqEx
-         return std::make_unique<NegEx>(std::make_unique<NucEqEx>(pos, silo::to_symbol(ref_symbol)));
-      } else { /// NucMbEx
-         return std::make_unique<NegEx>(std::make_unique<NucMbEx>(pos, silo::to_symbol(ref_symbol)));
+      if (exact <= 0) {  /// NucEqEx
+         return std::make_unique<NegEx>(std::make_unique<NucEqEx>(pos, silo::to_symbol(ref_symbol))
+         );
```
fengelniederhammer, line 115: This looks odd. Can we somehow make it look like this?
```suggestion
         return std::make_unique<NegEx>(
            std::make_unique<NucEqEx>(pos, silo::to_symbol(ref_symbol))
         );
```

include/silo/database.h
```diff
@@ -64,30 +64,28 @@ class DatabasePartition {
...
 
    void finalizeBuild(const Dictionary& dict);
 };
 
 struct PreprocessingConfig;
 
 class Database {
-   public:
-   const std::string wd; // working directory
+  public:
```
fengelniederhammer, line 80: Do we really want to have only two spaces indent here?

fengelniederhammer (approved):

src/example.test.cpp
JonasKellerer: In general, yes. But we want to have the declaration in the header files and the implementation in the cpp files, where this is done automatically. Thus this should not be an issue.
The example.test.cpp can be deleted later on, when we dont need it anymore for easy reference.

src/query_engine/query_engine.cpp
JonasKellerer: Good idea. However, with the clang-tidy pullrequest, this should not occur at this point due to renaming.

PR #35, opened 2023-03-14
Naming convention

### LINKED ISSUE #27: Naming convention

JonasKellerer:
Resolves #27


chaoran-chen:
image

The build can take 20 minutes? Not as part of this PR but do you think that this can be significantly improved (e.g. by some caching)?

src/query_engine/query_simplification.cpp
```diff
@@ -176,34 +259,59 @@ std::unique_ptr<BoolExpression> OrEx::simplify(const Database& db, const Databas
       return ret;
    }
 }
+OrExpression::OrExpression() {}
```
fengelniederhammer: Why are there implementations for `OrExpression` in this file? (same for `NOfExpression` below).

src/query_engine/query_engine.cpp
```diff
@@ -1040,32 +1036,29 @@ silo::QueryResult silo::execute_query(
...
-         } else {
-            query_result.queryResult = response::ErrorResult{
-               "Unknown action", std::string(action_type) + " is not a valid action"};
+         std::vector<std::string> group_by_fields;
+         for (const auto& field : action["groupByFields"].GetArray()) {
+            group_by_fields.emplace_back(field.GetString());
          }
+
+         query_result.queryResult = response::ErrorResult{
+            "Unknown action", std::string(action_type) + " is not a valid action"};
```
fengelniederhammer, line 1085: Let's rather return a "groupByFields is not properly implemented yet" exception instead. Your change changes the logic from "silently ignore valid cases" to "everything is invalid".

src/storage/Dictionary.cpp
```diff
@@ -6,14 +6,14 @@
 
 using namespace silo;
 
-void Dictionary::update_dict(
-   std::istream& meta_in,
+void Dictionary::updateDictionary(
+   std::istream& dictionary_file,
```
fengelniederhammer, line 10: I think it actually is a metadata file and not a dictionary file, isn't it?

src/database.cpp
```diff
@@ -539,47 +544,47 @@ unsigned silo::processMeta(
...
          }
       }
    }
 }
 
-void silo::Database::save(const std::string& save_dir) {
+[[maybe_unused]] void silo::Database::saveDatabaseState(const std::string& save_dir) {
    if (!partition_descriptor) {
-      std::cerr << "Cannot save db without partition_descriptor." << std::endl;
+      std::cerr << "Cannot saveDatabaseState db without partition_descriptor." << std::endl;
```
fengelniederhammer, line 645: revert

.github/workflows/ci.yml
```diff
@@ -28,25 +28,30 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - uses: cpp-linter/cpp-linter-action@v2
-        id: linter
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
```
fengelniederhammer, line 39: Why do we need to login here?

.github/workflows/ci.yml
```diff
@@ -28,25 +28,30 @@ jobs:
...
-        run: echo "Some files failed the linting checks!"
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build linter image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          push: false
+          load: true
```
fengelniederhammer, line 49: I guess we don't need to load the image, do we? Basically we don't need the image at all, we just want to build inside it?

Dockerfile
```diff
@@ -17,8 +18,7 @@ RUN --mount=type=cache,target=/root/.conan conan install . --build=missing --pro
 
 RUN  \
     --mount=type=cache,target=/root/.conan \
-    --mount=type=cache,target=build \
-    --mount=type=cache,target=build \
+#    --mount=type=cache,target=build \
```
fengelniederhammer, line 21: Reenable this line

Dockerfile
```diff
@@ -5,7 +5,8 @@ RUN apk update && apk add --no-cache py3-pip \
     cmake=3.24.3-r0 \
     linux-headers=5.19.5-r0 \
     boost-build=1.79.0-r0 \
-    libtbb=2021.7.0-r0
+    libtbb=2021.7.0-r0 \
+    clang15-extra-tools=15.0.7-r0
```
fengelniederhammer, line 9: Do we need these here?

fengelniederhammer (changes requested):
A couple of minor things, otherwise it looks good :)

And we should write down somewhere why we use an additional Dockerfile for the linter.

README.md
```diff
@@ -36,12 +38,27 @@ docker run -i silo
...
+```
+
+For linting we use clang-tidy. The config is stored in `.clang-tidy`. It will run automatically with the build process
+and will throw errors accordingly. However it is rather slow. If you only want a fast build use
+
+```shell
+./build_with_conan build_without_clang_tidy
+```
+
+When pushing to github, a separate dockerimage will be built, which runs the formatter. (This is a workaround, because
```
fengelniederhammer, line 63:
```suggestion
When pushing to github, a separate Docker image will be built, which runs the formatter. (This is a workaround, because
```

fengelniederhammer (approved):

JonasKellerer:
> image
> 
> The build can take 20 minutes? Not as part of this PR but do you think that this can be significantly improved (e.g. by some caching)?

Yes and no. The linting takes time, because he needs to analyze the code, with all includes. This we can hopefully reduce by reducing the size of the files and by cleaning up the imports in ticket #40 , but it will still take a long time. However, the build process of the image and the install of the packages, might be cacheable. We'll add a new ticket to the board to investigate.

PR #42, opened 2023-03-15
refactor: properly name the filter types

fengelniederhammer:


endToEndTests/test/queries/nOf_2of3_aggregated.json
```diff
@@ -10,17 +10,17 @@
       "matchExactly": false,
       "children": [
         {
-          "type": "NucEq",
+          "type": "NucleotideEquals",
           "position": 241,
           "value": "T"
```
JonasKellerer, line 15: Maybe "symbol" could be a better name?

JonasKellerer (approved):

chaoran-chen (commented):
I am not sure if I find this better. The full names are quite long and `NucMut` is easier to remember for me than `HasNucleotideMutation`. 

@Taepper, what do you think?

fengelniederhammer:
I don't exactly see the point why we should have short, obfuscated names that one has to look up in the documentation. Descriptive names are a key feature for understandable software.
* Long names are a lot more descriptive by themselves and can be understood without reading the docs. To me as a user, it was not clear what `NucEq` actually does as a filter. I only understood it by reading the code.
* Most queries will not be written by hand, i.e. length doesn't matter.
* At least for me, full names are actually easier to remember. If you type a wrong name, SILO should respond with a proper errror message telling you what the allowed values are. I find "remembering" not a problem, because you can simply look it up in the docs or hope for good error messages.

chaoran-chen:
Good points. I don't have a strong opinion and am happy with the change.

One thing regarding `Negated`: should we call it `Not` instead? Because we also write `And` and `Or` (and not something like `Conjunction` and `Disjunction`)

fengelniederhammer:
`Not` is better, I wasn't really happy with `Negated` anyway :)

fengelniederhammer:
@Taepper if you have a different opinion, we can still change it later. We'll merge this so that it doesn't block us.

PR #43, opened 2023-03-15
build: use Conan 2.0

fengelniederhammer:
Somehow we suddenly had problems building the Docker images locally with Conan 1, so we decided to migrate to Conan 2. The previously blocking issue (oneTBB could not be built with Conan 2) was resolved in the meantime.


chaoran-chen (approved):

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ endif ()
 
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
-set(CMAKE_CXX_FLAGS "-Wall -march=native")
+set(CMAKE_CXX_FLAGS "-Wall")
```
JonasKellerer, line 10: i would keep this in, so the compiler uses the native architecture for building

Dockerfile
```diff
@@ -7,16 +7,17 @@ RUN apk update && apk add --no-cache py3-pip \
     boost-build=1.79.0-r0 \
     libtbb=2021.7.0-r0
 
-RUN pip install conan==1.59.0
+RUN pip install conan==2.0.1
```
JonasKellerer, line 10: 2.0.2 is the latest version, if we want to use that

Dockerfile_linter
```diff
@@ -5,15 +5,16 @@ RUN apt update && apt install -y \
     python3-pip=22.0.2+dfsg-1ubuntu0.2 \
     clang-tidy=1:14.0-55~exp2
 
-RUN pip install conan==1.59.0
+RUN pip install conan==2.0.1
```
JonasKellerer, line 8: Same as above, 2.0.2

JonasKellerer (approved):

PR #44, opened 2023-03-16
feat: add matchAll filter

JonasKellerer:
used for https://github.com/GenSpectrum/LAPIS/issues/178 for an empty filter

### DIFF
--- a/endToEndTests/test/queries/invalidAction.json
+++ b/endToEndTests/test/queries/invalidAction.json
@@ -4,7 +4,7 @@
     "action": {
       "type": "invalid action"
     },
-    "filter": {
+    "filterExpression": {
       "type": "N-Of",
       "numberOfMatchers": 2,
       "matchExactly": false,
--- a/endToEndTests/test/queries/invalidMutationsMinProportion.json
+++ b/endToEndTests/test/queries/invalidMutationsMinProportion.json
@@ -5,7 +5,7 @@
       "type": "Mutations",
       "minProportion": -0.5
     },
-    "filter": {
+    "filterExpression": {
       "type": "N-Of",
       "numberOfMatchers": 2,
       "matchExactly": false,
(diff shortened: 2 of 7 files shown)


fengelniederhammer (approved):

chaoran-chen:
At the moment, could we use an empty `And` instead?

fengelniederhammer:
Yes, 
```
{
    "type": "N-Of",
    "numberOfMatchers": 0,
    "matchExactly": false,
    "children": []
}
```
also works. But both solutions seemed rather hacky. Also using an empty `And` intuitively seems to behave wrong, because I would have expected that it only is true, if all children match, and is false, if no child matches (would be the case here).

Taepper:
I would prefer the expression to be named 'True' and then also add a complementary 'False'.
Otherwise I see no problem with instead going with an empty And.

fengelniederhammer:
> expression to be named 'True'

Maybe we should discuss naming here ;) I think what you have in mind is boolean expressions. We have two conflicting namings here:
* In the JSON, the field is called "filter". That's the user's perspective: Sequences are filtered by the condition that one provides. This is what the name `MatchAll` reflects.
* In the code, everything is called "expression", more specifically: the common parent class is boolean expression. This indicates that they evaluate to true/false. This is what the name `True` reflects.

I think we should decide for one of the options: Either call everything a "filter" or an "expression", but we should not mix both. This includes either renaming the field in the JSON to "expression" or renaming `BoolExpression` and all child classes to `XyzFilter` (because those classes represent the JSON structure, so they should carry similar names).

I think I would prefer filter:
* It represents the user's perspective.
* The filters are not really "boolean": There is a "maybe", i.e. ternary logic. I know that it's simplified away, but actually it exists in the SILO query language.
* They don't evaluate to true/false, but to a list of sequences that fulfill the condition. This is very naturally the concept of a "filter".

But I'm open to hear arguments for calling it "expression"s ;)

Taepper:
An additional step will be added that transforms the boolean expressions to set-based operators. In the prototype this was not present for simplicity. The general approach will parse the json as boolean expressions.
The field is called filter, as we use the boolean expression to filter the sequences.

In SQL one also writes `true` in the where-clause. Note that the user generally does not write this. I expect the matchAll is mostly for testing purposes.

Taepper (approved):

Issue #40, opened 2023-03-14
SILO: Initial clean up


fengelniederhammer:
Bundling a couple of necessary refactorings before doing any bigger changes:

- [x] #37 
- [x] #38
- [x] #39
- [x] #41

PR #50, opened 2023-03-17
rearrange includes

JonasKellerer:
rearrange includes, try to remove transitive includes and insert used includes in implementation

part of #40


include/silo/query_engine/query_result.h
```diff
@@ -18,6 +21,13 @@ struct ErrorResult {
    std::string error;
    std::string message;
 };
+
+struct QueryResult {
```
fengelniederhammer, line 25: Good catch :+1:

fengelniederhammer (approved):

Issue #19, opened 2023-03-02
Implement Logging


fengelniederhammer:
Use Poco Logging instead of stdout for logging

JonasKellerer:
Maybe better [spdlog](https://github.com/gabime/spdlog)

PR #51, opened 2023-03-21
Use spdlog for logging

fengelniederhammer:
This is a part of #19. The performance log statements are not yet migrated to spdlog, they are still written to std::cout. This will be done in a separate step. I would propose to log those log messages into a separate log file.

### DIFF
--- a/include/silo/storage/sequence_store.h
+++ b/include/silo/storage/sequence_store.h
@@ -2,6 +2,7 @@
 #ifndef SILO_SEQUENCE_STORE_H
 #define SILO_SEQUENCE_STORE_H
 
+#include <spdlog/spdlog.h>
 #include <array>
 #include <boost/serialization/array.hpp>
 #include <roaring/roaring.hh>
@@ -27,6 +28,12 @@ struct Position {
    uint32_t flipped_bitmap = REFERENCE_BITMAP_IS_FLIPPED;
 
    bool nucleotide_symbol_n_indexed = false;
+};
+
+struct SequenceStoreInfo {
+   uint32_t sequence_count;
+   uint64_t size;
+   size_t n_bitmaps_size;
 };
 
 class SequenceStore {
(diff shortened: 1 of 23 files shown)


include/silo/storage/sequence_store.h
```diff
@@ -77,4 +84,10 @@ class SequenceStore {
 
 }  // namespace silo
 
+template <>
+struct fmt::formatter<silo::SequenceStoreInfo> : fmt::formatter<std::string> {
```
JonasKellerer, line 88: My Clion says, that we can mark this as `maybe_unused`. I like that because it tells us later on, that there is a purpose behind this entry. :)

src/silo/storage/sequence_store.cpp
```diff
@@ -1,19 +1,27 @@
...
 #include "silo/common/silo_symbols.h"
-#include "silo/database.h"
+
+auto fmt::formatter<silo::SequenceStoreInfo>::format(
+   silo::SequenceStoreInfo sequence_store_info,
+   fmt::format_context& ctx
+) -> decltype(ctx.out()) {
+   return format_to(
+      ctx.out(), "SequenceStoreInfo[sequence count: {}, size: {}, N bitmaps size: {}]",
+      sequence_store_info.sequence_count, sequence_store_info.size,
```
JonasKellerer, line 21: Maybe I'll have a look at the formatter. In my eyes, this should be on separate lines :)

src/silo_api/api.cpp
```diff
@@ -137,7 +139,37 @@ class SiloServer : public Poco::Util::ServerApplication {
...
 };
 
+static const int MAX_FILES_7 = 7;
+static const int AT_MIDNIGHT = 0;
+static const int AT_0_MINUTES = 0;
+static const bool DONT_TRUNCATE = false;
+
+static const std::chrono::duration<int64_t> FIVE_SECONDS = std::chrono::seconds(5);
+
+void setupLogger() {
```
JonasKellerer, line 149: Could this live in its own file(s)? I would not suspect it to be here.

src/silo/storage/sequence_store.cpp
fengelniederhammer: Yes, that would suit my eyes better, too ;) There are a couple of weird line breaks in the code now.

PR #56, opened 2023-03-23
Harden the API

### LINKED ISSUE #14: Harden the SILO api

fengelniederhammer:
resolves #14


include/silo_api/info_handler.h
```diff
@@ -1,26 +1,25 @@
 #ifndef SILO_INFOHANDLER_H
 #define SILO_INFOHANDLER_H
 
-#include <Poco/Net/HTTPRequestHandler.h>
 #include <Poco/Net/HTTPServerRequest.h>
 #include <Poco/Net/HTTPServerResponse.h>
 
+#include "silo_api/request_handler.h"
```
JonasKellerer, line 7: We could make this a forward declaration `class RestResource;`. This makes the build faster, and it is clearer, where `RestResource` comes from.

src/silo_api/request_handler.test.cpp
```diff
@@ -0,0 +1,74 @@
...
+  public:
+   MOCK_METHOD(
+      void,
+      handleRequest,
+      (Poco::Net::HTTPServerRequest & request, Poco::Net::HTTPServerResponse& response),
+      ()
+   );
+};
+
+TEST(ErrorRequestHandler, handlesRuntimeErrors) {
```
JonasKellerer, line 17: In other test cases we used snake case. Which do we prefer? Google says, that CamelCase would be better, because of strange leading underscore stuff. We should add this to the README

src/silo_api/request_handler.test.cpp
```diff
@@ -0,0 +1,74 @@
...
+   auto under_test = silo_api::ErrorRequestHandler(wrapped_handler_mock);
+
+   ON_CALL(*wrapped_handler_mock, handleRequest)
+      .WillByDefault(testing::Throw(std::runtime_error("my error message")));
+
+   silo_api::test::MockResponse response;
+   silo_api::test::MockRequest request(response);
+   under_test.handleRequest(request, response);
+
+   EXPECT_EQ(response.getStatus(), 500);
```
JonasKellerer, line 29: Maybe we could use [pocos responses](https://docs.pocoproject.org/current/Poco.Net.HTTPResponse.html)?

src/silo_api/request_handler.test.cpp
fengelniederhammer: The docs say that one should not use `_` in test names: https://github.com/google/googletest/blob/main/docs/faq.md#why-should-test-suite-names-and-test-names-not-contain-underscore

include/silo_api/info_handler.h
fengelniederhammer: I didn't manage to make it work since `InfoHandler` inherits from `RestResource`, i.e. The this header file needs to know about the structure of `RestResource`.

include/silo_api/request_handler.h
```diff
@@ -6,8 +6,17 @@
...
+#include <nlohmann/json.hpp>
 
 namespace silo_api {
+
+struct ErrorResponse {
+   std::string error;
+   std::string message;
+};
+
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(ErrorResponse, error, message);
```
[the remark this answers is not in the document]
fengelniederhammer: Doesn't work, because the `InfoHandler` needs to know that the generated functions exist. The alternative would be to only declare the functions here and include the macro in the .cpp file, but that results in more code...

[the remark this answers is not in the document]
JonasKellerer: If there will be more, we can consider doing this. For now I would suggest leaving it like this.

src/silo_api/request_handler.test.cpp
JonasKellerer: I'll have a look how to disable the clang-tidy functionality for this. Then I will change the test names accordingly. Maybe in a separate pull request.

include/silo_api/info_handler.h
JonasKellerer: Sadly, me neither... 
I've pushed a commit, where each handler is put in its own file. Thus, `RestResource` can be found more easily. We can also think about putting the request handlers in their own folder. What do you think?

fengelniederhammer: Sounds reasonable, but for now IMO the `/silo_api` is already the separate folder. If the api layer grows further, we can still split it.

JonasKellerer (approved):

PR #58, opened 2023-03-24
fix: throw exceptions instead of terminating the program via the assert macro

### LINKED ISSUE #15: Make sure that failed `assert`s don't terminate the API

fengelniederhammer:
resolves #15

Appreciate early feedback. I propose to solve this via a custom macro in which we can provide a custom error message. Using a macro has the advantage over a function, that we don't need to allocate the error message string if the check does not fail.

### DIFF
--- a/src/silo/query_engine/query_engine.cpp
+++ b/src/silo/query_engine/query_engine.cpp
@@ -1,7 +1,6 @@
 #include "silo/query_engine/query_engine.h"
 
 #include <tbb/parallel_for.h>
-#include <cassert>
 #include <memory>
 #include <roaring/roaring.hh>
 #include <string>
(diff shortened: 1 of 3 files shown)


src/silo/query_engine/query_engine.cpp
```diff
@@ -26,6 +26,11 @@
 #include "silo/database.h"
 #include "silo/query_engine/query_result.h"
 
+#define CHECK_SILO_QUERY(condition, message)    \
```
JonasKellerer, line 29: Do we want to extend the macro, that it can also work with other exceptions (namely extending it with one more argument), or do we want to keep it specific to silo queries?

JonasKellerer:
I like it. :) Makes it also easy to switch out, if we want to change it.

fengelniederhammer (commented):
Looks good to me (can't approve it myself).

JonasKellerer (approved):

PR #75, opened 2023-04-02
Use sorted files when building database.

### LINKED ISSUE #67: Bug: Building Database with Sorted Data
Currently the sorted data is not used for the actual building of the database.
In the partitioned folder, the sorted files receive the infix `_sorted` (between filename and ending).
This will be made configurable and the correct files will be used for building.

Taepper:
Resolves #67


fengelniederhammer (changes requested):
[prepare_dataset.cpp](https://github.com/GenSpectrum/LAPIS-SILO/blob/726e0948d403db26f5dea19bc3da9914c5d93af2/src/silo/prepare_dataset.cpp#L435-L436) also contains `"_sorted"`. As far as I can judge, this is connected to this change here?

Doesn't it make more sense to put those files into a separate directory instead of appending something to the file name? In the target state of the preprocessing pipeline we won't need those sorted files I guess, so let's maybe try to mess less with the input directory?

PR #80, opened 2023-04-03
fix: split PerfEvent into perf_event and block_timer

### LINKED ISSUE #68: Extract BlockTimer from PerfEvent
PerfEvent.hpp unnecessarily includes the BlockTimer struct. An error within preprocessor macros also leads to a failed build on OS other than Linux.

Taepper:
Resolves #68


include/external/block_timer.h
```diff
@@ -0,0 +1,30 @@
+//
+// Created by Alexander Taepper on 02.04.23.
+//
```
JonasKellerer, line 3: I think Clion creates these automatically, but I think we don't need those, because we can look at git blame if we want to find this out. So I would suggest deleting this comment.

include/external/block_timer.h
```diff
@@ -0,0 +1,30 @@
...
+   explicit BlockTimer(output_t& ref)
+       : output(ref),
+         start(Clock::now()) {}
+
+   ~BlockTimer() {
+      auto end = Clock::now();
+      output = std::chrono::duration_cast<Unit>(end - start).count();
+   }
+
+   output_t until_now() { return std::chrono::duration_cast<Unit>(Clock::now() - start).count(); }
```
JonasKellerer, line 27: Is this function needed by PerfEvent? If not I would suggest renaming it to camelCase.

include/external/block_timer.h
```diff
@@ -0,0 +1,30 @@
...
+
+template <typename Unit = std::chrono::microseconds, typename Clock = std::chrono::steady_clock>
+struct [[nodiscard]] BlockTimer {
+   using time_point_t = typename Clock::time_point;
+   using output_t = typename Clock::rep;
+
+   output_t& output;
+   time_point_t start;
+
+   explicit BlockTimer(output_t& ref)
```
JonasKellerer, line 18: What is `ref`? From the name, I can not find out its meaning. Same with `output`.

JonasKellerer:
Dumb question, but do we need `perf_event.h` still in the code? I could not find any file including it. But maybe I have not understood the idea behind the timer. Maybe you could explain it to me at some time.

Additionally I would suggest putting `block_timer.h` in another folder, maybe in `common`? I would use the `external` folder only for headers, we include from external sources, which we can not add through conan.

Taepper:
> Dumb question, but do we need `perf_event.h` still in the code? I could not find any file including it. But maybe I have not understood the idea behind the timer. Maybe you could explain it to me at some time.

I used PerfEvent when performance tweaking. In case we use it again, I would leave it in for now.
BlockTimer is completely independent of PerfEvent

JonasKellerer (approved):

Issue #61, opened 2023-03-30
Make sure that nucleotide positions start at 1


fengelniederhammer:
See https://github.com/GenSpectrum/LAPIS/pull/158#discussion_r1152912906

Check and eventually fix.

Taepper:
I will push the necessary change once we have less open pull requests

fengelniederhammer:
@Taepper has this been fixed with https://github.com/GenSpectrum/LAPIS-SILO/pull/141?

Taepper:
Yes, it did, I forgot to close it. For completeness I will commit a PR that shows that the position is now in all cases (including the mutation distribution) 1-indexed

PR #91, opened 2023-04-21
fix: do not ignore flipped bitmap in SequenceStore::getFlippedBitmapF…

danielgrittner:
…romAmbiguousSymbol

This PR fixes the following two bugs:

1. `COUNT_AMBIGUOUS_SYMBOLS` should be 7 and not 8 in `SequenceStore::getFlippedBitmapFromAmbiguousSymbol` and `SequenceStore::getBitmapFromAmbiguousSymbol` because the arrays only have 7 entries.

2. In `SequenceStore::getFlippedBitmapFromAmbiguousSymbol`, the result of `roaring::api::roaring_bitmap_flip(&bitmap_to_flip->roaring, 0, sequence_count)` was ignored. Hence, the behavior of `getFlippedBitmapFromAmbiguousSymbol` was exactly identical to `getBitmapFromAmbiguousSymbol`. As far as I understood it from the code, the intended behavior is to only flip the bitmap corresponding to the exact value of the nucleotide symbol.

### DIFF
--- a/include/silo/storage/sequence_store.h
+++ b/include/silo/storage/sequence_store.h
@@ -26,7 +26,7 @@ struct Position {
       archive& nucleotide_symbol_n_indexed;
    }
 
-   std::array<roaring::Roaring, SYMBOL_COUNT> bitmaps;
+   std::array<roaring::Roaring, SYMBOL_COUNT> bitmaps{};
    std::optional<NUCLEOTIDE_SYMBOL> symbol_whose_bitmap_is_flipped = std::nullopt;
    bool nucleotide_symbol_n_indexed = false;
 };
@@ -39,7 +39,7 @@ struct SequenceStoreInfo {
 
 class SequenceStore {
   private:
-   unsigned sequence_count;
+   unsigned sequence_count{};
 
   public:
    friend class boost::serialization::access;
@@ -51,8 +51,8 @@ class SequenceStore {
       archive& nucleotide_symbol_n_bitmaps;
    }
 
-   std::array<Position, GENOME_LENGTH> positions;
-   std::vector<roaring::Roaring> nucleotide_symbol_n_bitmaps;
+   std::array<Position, GENOME_LENGTH> positions{};
+   std::vector<roaring::Roaring> nucleotide_symbol_n_bitmaps{};
 
    SequenceStore();
 
--- a/src/silo/storage/sequence_store.test.cpp
+++ b/src/silo/storage/sequence_store.test.cpp
@@ -1,0 +1,47 @@
+#include "silo/storage/sequence_store.h"
+
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <string>
+
+std::unique_ptr<silo::SequenceStore> setupSequenceStore() {
+   std::unique_ptr<silo::SequenceStore> sequence_store = std::make_unique<silo::SequenceStore>();
+
+   // Construct test genomes.
+   std::string test_genome1 = "A";
+   test_genome1.reserve(silo::GENOME_LENGTH);
+   for (unsigned i = 1; i < silo::GENOME_LENGTH; ++i) {
+      test_genome1.push_back('A');
+   }
+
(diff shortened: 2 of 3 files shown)


src/silo/storage/sequence_store.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+   sequence_store->interpret({std::move(test_genome1), std::move(test_genome2)});
+
+   return sequence_store;
+}
+
+TEST(SequenceStore, shouldReturnCorrectBitmapForAmbiguousSymbol) {
+  auto sequence_store = setupSequenceStore();
+
+  // Note: 1-indexed!
+  auto* bitmap2 = sequence_store->getBitmapFromAmbiguousSymbol(1, silo::NUCLEOTIDE_SYMBOL::A);
```
JonasKellerer, line 33: I think this can be called 'bitmap', same as below :)

include/silo/storage/sequence_store.h
```diff
@@ -39,7 +39,7 @@ struct SequenceStoreInfo {
 
 class SequenceStore {
   private:
-   unsigned sequence_count;
+   unsigned sequence_count{};
```
JonasKellerer, line 42: Personally, I am not happy to initialize the `sequence_count` here, but since we do not have a real constructor, I think it is fine.

include/silo/storage/sequence_store.h
```diff
@@ -51,8 +51,8 @@ class SequenceStore {
       archive& nucleotide_symbol_n_bitmaps;
    }
 
-   std::array<Position, GENOME_LENGTH> positions;
-   std::vector<roaring::Roaring> nucleotide_symbol_n_bitmaps;
+   std::array<Position, GENOME_LENGTH> positions{};
+   std::vector<roaring::Roaring> nucleotide_symbol_n_bitmaps{};
```
JonasKellerer, line 55: Same as above. :)

src/silo/storage/sequence_store.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+   std::unique_ptr<silo::SequenceStore> sequence_store = std::make_unique<silo::SequenceStore>();
+
+   // Construct test genomes.
+   std::string test_genome1 = "A";
+   test_genome1.reserve(silo::GENOME_LENGTH);
+   for (unsigned i = 1; i < silo::GENOME_LENGTH; ++i) {
+      test_genome1.push_back('A');
+   }
+
+   std::string test_genome2 = "C";
```
JonasKellerer, line 18: I think, we can use this also in other places, so I would suggest making it a function, and therefore getting rid of the comment as well.

src/silo/storage/sequence_store.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+   sequence_store->interpret({std::move(test_genome1), std::move(test_genome2)});
+
+   return sequence_store;
+}
+
+TEST(SequenceStore, shouldReturnCorrectBitmapForAmbiguousSymbol) {
+  auto sequence_store = setupSequenceStore();
+
+  // Note: 1-indexed!
```
JonasKellerer, line 32: Is this comment necessary, or does it belong to #61 ?

JonasKellerer (commented):
Nice find. :) I just wanted to add similar fixes, as I was working on the Maybe query.

src/silo/storage/sequence_store.test.cpp
danielgrittner: Well, this comment was rather for myself because I didn't know in the beginning that genomes are 1-indexed 😅

include/silo/storage/sequence_store.h
danielgrittner: Left it as it is since there is no real constructor atm.

src/silo/storage/sequence_store.test.cpp
danielgrittner: Good idea!

Issue #72, opened 2023-04-02
Parse Primary Key as String


Taepper:
The EPI_ISL is currently expected in a very specific format where an integer is extracted. This was not used in any optimizations and should be changed back to simply parsing that tsv column as a string to be more flexible for different primary key formats.

PR #94, opened 2023-05-02
Refactor: Primary key is no longer parsed as prefixed integer

Taepper:
To generalize the implementation to datasets with differently formatted primary keys, we no longer parse its specific format


src/silo/prepare_dataset.cpp
```diff
@@ -20,28 +20,24 @@
 ) {
    SPDLOG_INFO("Pruning metadata");
 
-   std::unordered_set<uint64_t> epi_isl_ids;
+   std::unordered_set<std::string> epi_isl_set;
```
fengelniederhammer, line 23:
```suggestion
   std::unordered_set<std::string> found_primary_keys;
```
I usually don't see any value in adding the type (like `set` or `map`) to a variable name.

Also this should be more general. I think we don't want to stick to "epi_isls" anymore. From my understanding, FASTA files consist of headers and the sequences. The header looks like ">identifier". We expect that the identifier is our primary key, isn't it? We should reflect this is variable naming here.

src/silo/storage/metadata_store.cpp
```diff
@@ -74,14 +70,14 @@ unsigned MetadataStore::fill(
 
 // TODO(someone): clean up and specify inputs
 void MetadataStore::inputSequenceMeta(
-   uint64_t epi_isl_number,  // NOLINT
+   std::string primary_key,  // NOLINT
```
fengelniederhammer, line 73: Do we still need the `// NOLINT`?

src/silo/storage/sequence_store.cpp
```diff
@@ -34,14 +34,20 @@ unsigned silo::SequenceStore::fill(std::istream& input_file) {
...
          break;
       }
       if (!getline(input_file, genome)) {
          break;
       }
+
+      if (epi_isl_with_prefix.at(0) != '>') {
+         throw silo::PreprocessingException("EPI ISL prefix > is missing.");
+      }
+      std::string const epi_isl = epi_isl_with_prefix.substr(1);
```
fengelniederhammer, line 49: The very same logic is duplicated in a couple of places. Since we're already on it, how about extracting a `SequenceFileReader` or similar that encapsulates this?

Taepper: Good idea, I am on it. How about "fasta"-reader instead?

fengelniederhammer: Also sounds good. :)

src/silo/common/fasta_reader.cpp
```diff
@@ -0,0 +1,43 @@
...
+    : in_file(in_file_name) {}
+
+bool silo::FastaReader::nextKey(std::string& key) {
+   std::string key_with_prefix;
+   if (!getline(in_file.getInputStream(), key_with_prefix)) {
+      return false;
+   }
+
+   if (key_with_prefix.at(0) != '>') {
+      return false;
```
fengelniederhammer, line 13: I think we should throw exceptions instead.

Can't we simply return the key instead of populating a passed reference?

include/silo/common/fasta_reader.h
```diff
@@ -0,0 +1,24 @@
...
+   silo::InputStreamWrapper in_file;
+
+  public:
+   explicit FastaReader(std::string in_file_name);
+
+   bool nextKey(std::string& key);
+
+   bool next(std::string& key, std::string& genome);
+
+   void reset();
```
fengelniederhammer, line 20: Do we need that method?

src/silo/common/fasta_reader.cpp
Taepper: I initially had it like that, but this would copy too much data for the genomes

Taepper: Exception is a good idea, what would you suggest, since this is now no longer in `silo/preprocesssing` but in `silo/common`? How about `silo/common/FastaFormatException`

include/silo/common/fasta_reader.h
Taepper: Yes, the sorting method needs to read the file twice. I found this better than reopening the file a second time

src/silo/common/fasta_reader.cpp
Taepper: > I initially had it like that, but this would copy too much data for the genomes

And I wanted to have the key/genome consistent

src/silo/common/fasta_reader.cpp
fengelniederhammer, line 1: Please add a unit test

src/silo/common/fasta_reader.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+bool silo::FastaReader::next(std::string& key, std::string& genome) {
+   std::string key_with_prefix;
+   if (!getline(in_file.getInputStream(), key_with_prefix)) {
+      return false;
+   }
+
+   if (key_with_prefix.at(0) != '>') {
+      throw FastaFormatException("Fasta key prefix '>' missing for key: " + key_with_prefix);
+   }
```
fengelniederhammer, line 32: Can't we reuse that?
```suggestion
   nextKey(key);
```

src/silo/prepare_dataset.cpp
```diff
@@ -345,70 +287,60 @@ void sortChunk(
       // Reset gpointer, read file again, putting every genome at the correct position.
       // Write file to ostream
 
-      struct EPIDate {
-         uint64_t epi;
+      struct KeyDatePair {
```
fengelniederhammer, line 290: It's kinda funny that something called `Pair` has three members ;)

fengelniederhammer (approved):

fengelniederhammer (changes requested):
A couple of minor things :)

src/silo/prepare_dataset.cpp
Taepper: I thought about that too, but figured that the position in the file is more "meta info". Also not fully happy with the name, but KeyDate would definitely be worse

fengelniederhammer: Ok, if you think it fits, we can leave it that way. At least it already suggests, that the third field is maybe not that important.

src/silo/common/fasta_reader.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+   key = key_with_prefix.substr(1);
+
+   in_file.getInputStream().ignore(LONG_MAX, '\n');
+   return true;
+}
+
+bool silo::FastaReader::next(std::string& key, std::string& genome) {
+   nextKey(key);
```
fengelniederhammer, line 25: I think it's unfortunately not that easy. Don't we need a `return false` somehow?

Taepper: Hmm, after consideration I would leave this change out. The repetition is minimal and we would need to mutate the key variable earlier, which is not wanted in case we throw an exception and the nextKey function would need to be passed an additional parameter, to consider the case whether the genome is skipped or not.

src/silo/common/fasta_reader.test.cpp
```diff
@@ -0,0 +1,375 @@
...
+   std::string genome;
+   EXPECT_TRUE(in_file.nextKey(key));
+   EXPECT_EQ(key, "EPI_ISL_1408408");
+
+   EXPECT_TRUE(in_file.next(key, genome));
+   EXPECT_EQ(key, "EPI_ISL_1749899");
+   EXPECT_EQ(
+      genome,
+      "------------------------------"
+      "ACCAACCAACTTTCGATCTCTTGTAGATCTGTTCTCTAAACGAACTTTAAAATCTGTGTGGCTGTCACTCGGCTGCATGCTTAGTGCACTCA"
```
fengelniederhammer, line 29: We could as well test this with an artificial sequence file that has a shorter genome. This is a unit test, so we only need data that has a correct structure.

src/silo/common/fasta_reader.test.cpp
```diff
@@ -0,0 +1,375 @@
+#include <gtest/gtest.h>
+#include <string>
+#include <unordered_map>
+
+#include "silo/common/fasta_format_exception.h"
+#include "silo/common/fasta_reader.h"
+#include "silo/preprocessing/preprocessing_config.h"
+
+TEST(FastaReader, shouldReadFastaFile) {
+   const silo::InputDirectory input_directory{"./testBaseData/"};
```
fengelniederhammer, line 10: Why do we need all those other files here?

src/silo/common/fasta_reader.test.cpp
```diff
@@ -0,0 +1,375 @@
...
+      "TAGTTAACTTTAATCTCACATAGCAATCTTTAATCAGTGTGTAACATTAGGGAGGACTTGAAAGAGCCACCACATTTTCACCGAGGCCACGC"
+      "GGAGTACGATCGAGTGTACAGTGAACAATGCTAGGGAGAGCTGCCTATATGGAAGAGCCCTAATGTGTAAAATTAATTTTAGTAGTGCTATC"
+      "CCCATGTGATTTTAATAGCTTCTTAG------------------------------------------"
+   );
+
+   EXPECT_TRUE(in_file.nextKey(key));
+   EXPECT_EQ(key, "EPI_ISL_2016901");
+}
+
+TEST(FastaReader, shouldFailWrongFormat) {
```
fengelniederhammer, line 362: I'd suggest to make this easier to read:
```suggestion
TEST(FastaReader, givenDataInWrongFormatThenShouldThrowAnException) {
```
Actually it would be even nicer to say what's wrong about the format (e.g. given data with missing > in header)

src/silo/common/fasta_reader.test.cpp
fengelniederhammer, line 1: This test is missing at least two cases:
* We reached the end of the file
* The last line of the file is a header (without a sequence)

Maybe we even want to redesign the test (or even the class) such that we can write the file content into the test instead of individual files? Maybe the class should accept a stream in the constructor, so that we can use a string stream as input in the test? Or we could generate temporary files?

src/silo/common/fasta_reader.cpp
fengelniederhammer: How about that?
```
bool populateKey(std::string& key) {
   std::string key_with_prefix;
   if (!getline(in_file.getInputStream(), key_with_prefix)) {
      return false;
   }

   if (key_with_prefix.at(0) != '>') {
      throw FastaFormatException("Fasta key prefix '>' missing for key: " + key_with_prefix);
   }

   key = key_with_prefix.substr(1);
   return true;
}

bool silo::FastaReader::nextKey(std::string& key) {
   auto key_was_read = populateKey(key);

   in_file.getInputStream().ignore(LONG_MAX, '\n');
   return key_was_read;
}

bool silo::FastaReader::next(std::string& key, std::string& genome) {
   auto key_was_read = populateKey(key);
   if (!key_was_read) {
      return false;
   }

   if (!getline(in_file.getInputStream(), genome)) {
      throw FastaFormatException(
         "Missing genome sequence in line following key: " + key_with_prefix
      );
   }

   return true;
}
```
* Reading the genome or not is solved by delegation instead of by an additional parameter
* Why does it matter whether the key is mutated or not in case of an exception? If it's really such a big deal, then `populateKey` could return an optional string instead of populating the `key` reference.

fengelniederhammer:
This resolves #72, doesn't it?

include/silo/common/fasta_format_exception.h
```diff
@@ -0,0 +1,19 @@
+//
+// Created by Alexander Taepper on 02.05.23.
+//
```
JonasKellerer, line 3: Delete this comment :)

include/silo/common/fasta_reader.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include "silo/common/input_stream_wrapper.h"
+
+namespace silo {
+class FastaReader {
+  private:
+   silo::InputStreamWrapper in_file;
+
+  public:
+   explicit FastaReader(const std::string& in_file_name);
```
JonasKellerer, line 14: Can we use the filesystem::path here instead of string? This would make a test, if the path exists a little easier. Maybe also add the exception in the cpp if it does not exist, or can we cope with a not provided file?

JonasKellerer: Maybe we could also add this into the InputStreamWrapper, or is it nested to deeply?

src/silo/storage/sequence_store.cpp
```diff
@@ -1,5 +1,6 @@
 #include "silo/storage/sequence_store.h"
 
+#include <silo/common/fasta_reader.h>
```
JonasKellerer, line 3: Use quotes here for the input, since it comes from our own code.

JonasKellerer (commented):
Nice. I like it :)

include/silo/prepare_dataset.h
```diff
@@ -5,6 +5,8 @@
 #include <string>
 #include <unordered_map>
 
+#include "silo/common/fasta_reader.h"
```
fengelniederhammer, line 8: We could probably use a forward declaration here, can't we?

src/silo/common/fasta_reader.test.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "silo/common/fasta_format_exception.h"
+#include "silo/common/fasta_reader.h"
+#include "silo/preprocessing/preprocessing_config.h"
+
+TEST(FastaReader, shouldReadFastaFile) {
+   const silo::InputDirectory input_directory{"./testBaseData/fastaFiles/"};
+   const silo::SequenceFilename sequence_filename{"test.fasta"};
+   const std::filesystem::path file_path =
+      silo::createPath(input_directory.directory, sequence_filename.filename);
+   silo::FastaReader in_file(file_path);
```
fengelniederhammer, line 14: Nitpicking: Strictly speaking, this is not really a file. I'd either call it `under_test` or `fasta_reader`.

src/silo/common/fasta_reader.test.cpp
```diff
@@ -0,0 +1,71 @@
...
+   const std::filesystem::path file_path =
+      silo::createPath(input_directory.directory, sequence_filename.filename);
+   silo::FastaReader wrong_in_file(file_path);
+
+   std::string key;
+   std::string genome;
+   EXPECT_THROW(wrong_in_file.next(key, genome), silo::FastaFormatException);
+}
+
+TEST(FastaReader, givenDataInWrongFormatThenShouldThrowAnException2) {
```
fengelniederhammer, line 57: Let's give this test a proper name, maybe something like:
```suggestion
TEST(FastaReader, givenDataInWithMissingGenomeThenShouldThrowAnException) {
```

src/silo/prepare_dataset.cpp
```diff
@@ -15,33 +16,19 @@
...
+   silo::FastaReader& sequences_in,
    std::ostream& metadata_out
 ) {
    SPDLOG_INFO("Pruning metadata");
 
-   std::unordered_set<uint64_t> epi_isl_ids;
+   std::unordered_set<std::string> found_primary_keys;
    uint32_t found_sequences_count = 0;
    uint32_t found_metadata_count = 0;
    {
```
fengelniederhammer, line 27: Do we need that scope?

fengelniederhammer (approved):
A couple of minor things that are maybe not that important either. Otherwise it looks good to me :)

src/silo/prepare_dataset.cpp
Taepper: What do you mean?

Taepper: Ah, I removed one curly brace pair that restricted the scope, but left the other one in, because the variable name key would otherwise clash. I would therefore wait for the refactoring of the metadata reading before deciding on this refactor. I would not want to reuse the variable, which would make it more confusing and neither complicate the names too much, because both keys in the metadata and the sequence file semantically represent the same thing

fengelniederhammer: Makes sense. :)

Issue #81, opened 2023-04-05
Generalize reading metadata


fengelniederhammer:
Implement reading arbitrary columns from the metadata files. Currently, only hardcoded column names in a hardcoded order can be read. Validate the columns against configured column names (Those can be hardcoded if reading the actual config is not implemented yet.)

PR #95, opened 2023-05-02
Read the config

### LINKED ISSUE #6: Read the config
Enable SILO to read the pathogen config.

Use the config when reading the input files to validate them:
* Metadata contains fields specified in the config.
* Configured primary key is unique.

JonasKellerer:
Resolves #6

The config is only read. Usage will be implemented with #81 or after.

### DIFF
--- a/include/silo/config/config_exception.h
+++ b/include/silo/config/config_exception.h
@@ -3,13 +3,13 @@
 
 #include <iostream>
 
-namespace silo {
+namespace silo::config {
 
 class ConfigException : public std::runtime_error {
   public:
    explicit ConfigException(const std::string& error_message);
 };
 
-}  // namespace silo
+}  // namespace silo::config
 
 #endif  // SILO_INCLUDE_SILO_CONFIG_CONFIG_EXCEPTION_H_
--- a/include/silo/config/config_exception.h
+++ b/include/silo/config/config_exception.h
@@ -1,0 +1,15 @@
+#ifndef SILO_INCLUDE_SILO_CONFIG_CONFIG_EXCEPTION_H_
+#define SILO_INCLUDE_SILO_CONFIG_CONFIG_EXCEPTION_H_
+
+#include <iostream>
+
+namespace silo {
+
+class ConfigException : public std::runtime_error {
+  public:
+   explicit ConfigException(const std::string& error_message);
+};
+
+}  // namespace silo
+
+#endif  // SILO_INCLUDE_SILO_CONFIG_CONFIG_EXCEPTION_H_
--- a/src/silo/config/config_repository.cpp
+++ b/src/silo/config/config_repository.cpp
@@ -6,13 +6,13 @@
 #include "silo/config/config_exception.h"
 #include "silo/config/database_config_reader.h"
 
-namespace silo {
+namespace silo::config {
 
-ConfigRepository::ConfigRepository(const silo::DatabaseConfigReader& reader)
+ConfigRepository::ConfigRepository(const DatabaseConfigReader& reader)
     : reader_(reader) {}
 
 DatabaseConfig ConfigRepository::readConfig(const std::filesystem::path& config_path) const {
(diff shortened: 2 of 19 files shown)


include/silo/config/config_exception.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SILO_INCLUDE_SILO_CONFIG_CONFIG_EXCEPTION_H_
+#define SILO_INCLUDE_SILO_CONFIG_CONFIG_EXCEPTION_H_
+
+#include <iostream>
+
+namespace silo {
```
fengelniederhammer, line 6: Maybe put those in a separate namespace?
```suggestion
namespace silo::config {
```

src/silo/config/database_config_reader.cpp
```diff
@@ -0,0 +1,58 @@
...
+   }
+
+   throw silo::ConfigException("Unknown metadata type: " + type);
+}
+
+template <>
+struct convert<silo::DatabaseMetadata> {
+   static bool decode(const Node& node, silo::DatabaseMetadata& metadata) {
+      metadata.name = node["name"].as<std::string>();
+      metadata.type = stringToMetadataType(node["type"].as<std::string>());
```
fengelniederhammer, line 47: Is it possible to implement this as a static method?
```suggestion
      metadata.type = silo::DatabaseMetadataType::from(node["type"].as<std::string>());
```

src/silo/config/database_config_reader.test.cpp
```diff
@@ -0,0 +1,50 @@
+#include "silo/config/database_config_reader.h"
+#include "silo/config/config_exception.h"
+
+#include <gtest/gtest.h>
+
+TEST(DatabaseConfigReader, shouldReadConfigWithoutErrors) {
+   ASSERT_NO_THROW(silo::DatabaseConfigReader().readConfig("testBaseData/test_database_config.yaml")
```
fengelniederhammer, line 7: I think we can delete this test in favor of the second test?

src/silo/config/database_config_reader.test.cpp
fengelniederhammer, line 1: Do we maybe want to test the case where we read a file that has an invalid structure? I.e. `metadata` is missing or similar?

src/silo/config/config_repository.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace silo {
+
+ConfigRepository::ConfigRepository(const silo::DatabaseConfigReader& reader)
+    : reader_(reader) {}
+
+DatabaseConfig ConfigRepository::readConfig(const std::filesystem::path& config_path) const {
+   const auto config = reader_.readConfig(config_path);
+   validateConfig(config);
+   return config;
```
fengelniederhammer, line 17: "Clang-Tidy: Constness of 'config' prevents automatic move "

src/silo/config/config_repository.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void ConfigRepository::validateConfig(const DatabaseConfig& config) const {
+   std::map<std::string, DatabaseMetadataType> metadata_map;
+   for (const auto& metadata : config.schema.metadata) {
+      if (metadata_map.find(metadata.name) != metadata_map.end()) {
+         throw ConfigException("Metadata name is not unique: " + metadata.name);
```
fengelniederhammer, line 24: I think that makes it a bit clearer what the problem was:
```suggestion
         throw ConfigException("Metadata name is defined twice in the config:  " + metadata.name);
```

fengelniederhammer (approved):

src/silo/config/database_config_reader.cpp
JonasKellerer: Sadly, I did not find a way to attach it to the enum class. But since we have our namespace I think it is fine to have a function `toDatabaseMetadataType`.

Issue #82, opened 2023-04-05
Generalize the metadata store to hold arbitrary columns


fengelniederhammer:
This is a follow up to #81. After reading arbitrary columns from the metadata files, they should also be stored and queryable.

- [x] #99 
- [x] #100 
- [x] #101 
- [x] #103 
- [x] #113 
- [x] #114

PR #96, opened 2023-05-02
Generalize the metadata reader

### LINKED ISSUE #81: Generalize reading metadata

fengelniederhammer:
resolves #81 

This is a sketch how we could generalize reader the metadata files. The metadata reader will probably need another method that yields an iterator or similar over the individual, full rows.


JonasKellerer (commented):
I like it. I think some things are leftover for #82

JonasKellerer (approved):

Issue #100, opened 2023-05-04
Implement classes for metadata columns


fengelniederhammer:
Write an abstraction for columns:
* columns can be indexed
* columns can be not indexed
In both cases, we need a method to filter the stored values by a specific value: `Roaring Column::filter<v>(v value)`.

We need columns that can store string, pango_lineages and dates.

Also implement the DateBetween filter for date columns.

Issue #99, opened 2023-05-04
Generalize the Dictionary to hold arbitrary columns


fengelniederhammer:
Don't hardcode the column names in the `Dictionary`. Make it possible to store arbitrary columns (string, date and pango_lineage).

PR #102, opened 2023-05-04
feat: implement dummy method to lookup in dictionary

fengelniederhammer:
This is a preparation so that we can work on #99 and #100 in parallel.


JonasKellerer (approved):

fengelniederhammer:
Irrelevant due to #105

PR #104, opened 2023-05-10
Metadata columns

### LINKED ISSUE #100: Implement classes for metadata columns

fengelniederhammer:
resolves #100 

This is still missing a pango lineage column, but that can easily be implemented later and we still lack a proper type definition for pango lineages.


include/silo/storage/raw_base_column.h
```diff
@@ -0,0 +1,26 @@
...
+namespace silo::storage {
+
+template <typename T>
+class RawBaseColumn {
+  private:
+   std::string column_name;
+   std::vector<T> values;
+
+  public:
+   RawBaseColumn(std::string column_name, std::vector<T> values);
```
JonasKellerer, line 19: Same here with the const ref.

src/silo/storage/date_column.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+   return final_bitmap;
+}
+
+SortedDateColumn::SortedDateColumn(
+   std::string column_name,
+   std::vector<std::chrono::year_month_day> values
+)
+    : column_name(std::move(column_name)),
+      values(std::move(values)) {}
```
JonasKellerer, line 35: Same here, const ref?

src/silo/storage/dictionary.cpp
```diff
@@ -354,4 +354,11 @@ uint32_t Dictionary::getRegionCount() const {
    return region_count;
 }
 
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
```
JonasKellerer, line 357: Not important, but I think there is also //NOLINTBEGIN and //NOLINTEND

JonasKellerer (approved):
I like it. Only minor changes, mostly concerning pass py value vs pass by const ref.

src/silo/storage/dictionary.cpp
fengelniederhammer: This will resolve itself after a rebase on your branch anyway.

Issue #101, opened 2023-05-04
Generalize the metadata store


fengelniederhammer:
use #99 and #100 to generalize the metadata store such that in can store arbitrary columns with string, date and pango_lineage values. Also adapt the queries such that they work on the new metadata store. Encapsulate access to the columns, such that the queries don't always access public fields.

PR #105, opened 2023-05-10
Generalize the Dictionary to hold arbitrary columns

### LINKED ISSUE #99: Generalize the Dictionary to hold arbitrary columns

JonasKellerer:
Resolves #99 

Needs #100 and is followed by #101 to fully take effect.


src/silo/storage/dictionary.cpp
```diff
@@ -12,6 +12,45 @@
...
+
+   for (const auto& column_name : pango_lineage_column_names) {
+      pango_lineage_lookup[column_name] = std::unordered_map<PangoLineage, ValueId>();
+      reverse_pango_lineage_lookup[column_name] = std::vector<PangoLineage>();
+   }
+}
+
+Dictionary::Dictionary()
+    : Dictionary(
+         std::vector<std::string>({"country", "region", "division", "column"}),
```
fengelniederhammer, line 50: This looks like a leftover?
```suggestion
         std::vector<std::string>({"country", "region", "division"}),
```

src/silo/storage/dictionary.cpp
```diff
@@ -56,302 +95,110 @@ void Dictionary::updateDictionary(
...
 uint32_t Dictionary::getPangoLineageIdInLookup(const std::string& pango_lineage) const {
-   if (pango_lineage_dictionary.contains(pango_lineage)) {
-      return pango_lineage_dictionary.at(pango_lineage);
+   if (pango_lineage_lookup.at("pango_lineage").contains(pango_lineage)) {
+      return pango_lineage_lookup.at("pango_lineage").at(pango_lineage);
    }
    return UINT32_MAX;
 }
 
 std::string error_string = "NOID";
```
fengelniederhammer, line 126: According to my IDE, this is unused. Can we delete it?

src/silo/storage/dictionary.cpp
```diff
@@ -56,302 +95,110 @@ void Dictionary::updateDictionary(
...
-   return UINT32_MAX;
+   return std::nullopt;
 }
 
-[[maybe_unused]] const std::string& Dictionary::getColumn(uint32_t column_id_in_lookup) const {
-   if (column_id_in_lookup == UINT32_MAX) {
-      return error_string;
+std::optional<std::string> Dictionary::lookupPangoLineageValue(
+   const std::string& column_name,
+   uint32_t value_id
```
fengelniederhammer, line 196:
```suggestion
   ValueId value_id
```

src/silo/storage/dictionary.cpp
```diff
@@ -56,302 +95,110 @@ void Dictionary::updateDictionary(
...
+   return UINT32_MAX;
+}
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+std::string Dictionary::getColumn(uint32_t /*column_id_in_lookup*/) const {
+   return "";
 }
 
-[[maybe_unused]] const std::string& Dictionary::getGeneralLookup(uint64_t general_id_in_lookup
+std::optional<ValueId> Dictionary::lookupValueId(
+   const std::string& column_name,
```
fengelniederhammer, line 175: Let's also use the alias here
```suggestion
   const ColumnName& column_name,
```

include/silo/storage/dictionary.h
```diff
@@ -3,35 +3,38 @@
...
-   uint32_t region_count = 0;
-   uint32_t additional_columns_count = 0;
-   uint64_t general_count = 0;
+   std::unordered_map<ColumnName, std::unordered_map<std::string, ValueId>> string_lookup;
+   std::unordered_map<ColumnName, std::vector<std::string>> reverse_string_lookup;
+   void updateStringLookup(const ColumnName& column_name, const std::string& value);
+
+   std::unordered_map<ColumnName, std::unordered_map<PangoLineage, ValueId>> pango_lineage_lookup;
+   std::unordered_map<ColumnName, std::vector<PangoLineage>> reverse_pango_lineage_lookup;
+   void updatePangoLineageLookup(const ColumnName& column_name, const PangoLineage& pango_lineage);
```
fengelniederhammer, line 29: We could probably also encapsulate this and deduplicate it? Does that make sense? And maybe find a better name than "TypedLookup".
```suggestion
   TypedLookup<std::string> string_lookup;
   TypedLookup<PangoLineage> pango_lineage_lookup;
```

src/silo/storage/dictionary.test.cpp
```diff
@@ -0,0 +1,82 @@
...
+   ASSERT_FALSE(dictionary.lookupValueId("country", "NotACountry").has_value());
+}
+
+TEST_F(DictionaryTestFixture, shouldLookupValueId) {
+   ASSERT_EQ(dictionary.lookupValueId("country", "Switzerland"), 0);
+   ASSERT_EQ(dictionary.lookupValueId("division", "Bern"), 1);
+
+   ASSERT_EQ(
+      dictionary.lookupValueId("division", "Bern"), dictionary.lookupValueId("division", "Bern")
+   );
```
fengelniederhammer, line 48: Why do we need that assertion?^^

fengelniederhammer (approved):
Just a couple of minor things, overall it looks good :)

src/silo/storage/dictionary.test.cpp
JonasKellerer: I thought that reading a value twice should still result in the same value. But is not neccessary per se :)

include/silo/storage/dictionary.h
JonasKellerer: Good idea.

PR #107, opened 2023-05-11
Validate metadata columns against config

### LINKED ISSUE #103: Validate metadata columns against config
Make sure that we only read columns that are configured in the config. How do we handle columns that are present in the metadata but not in the config? Do we ignore them?

JonasKellerer:
Resolves #103

### DIFF
--- a/src/silo/database.cpp
+++ b/src/silo/database.cpp
@@ -520,6 +520,14 @@ silo::DetailedDatabaseInfo silo::Database::detailedDatabaseInfo() const {
 }
 
 void silo::Database::preprocessing(const PreprocessingConfig& config) {
+   SPDLOG_INFO("preprocessing - validate database config");
+   config::ConfigRepository().getValidatedConfig(config.database_config_file);
+
+   SPDLOG_INFO("preprocessing - validate metadata file against config");
+   silo::preprocessing::MetadataValidator().validateMedataFile(
+      config.metadata_file, config.database_config_file
+   );
+
    SPDLOG_INFO("preprocessing - building pango lineage counts");
    pango_descriptor = std::make_unique<preprocessing::PangoLineageCounts>(
       preprocessing::buildPangoLineageCounts(alias_key, config.metadata_file)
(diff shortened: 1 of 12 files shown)


src/silo/database.cpp
```diff
@@ -550,6 +552,14 @@ void silo::Database::preprocessing(const PreprocessingConfig& config) {
       config.sequence_file.extension()
    );
 
+   SPDLOG_INFO("preprocessing - validate database config");
+   config::ConfigRepository().getValidatedConfig(config.database_config_file);
+
+   SPDLOG_INFO("preprocessing - validate metadata file agains config");
```
fengelniederhammer, line 558:
```suggestion
   SPDLOG_INFO("preprocessing - validate metadata file against config");
```

src/silo/database.cpp
```diff
@@ -550,6 +552,14 @@ void silo::Database::preprocessing(const PreprocessingConfig& config) {
       config.sequence_file.extension()
    );
 
+   SPDLOG_INFO("preprocessing - validate database config");
```
fengelniederhammer, line 555: I wonder whether we should actually do that earlier. There is no need to do the whole partitioning when we notice later on that the metadata file was actually invalid.

src/silo/preprocessing/metadata_validator.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+void MetadataValidator::validateMedataFile(
+   const std::filesystem::path& metadata_file,
+   const std::filesystem::path& config_file
+) const {
+   const auto csv_reader = MetadataReader::getReader(metadata_file);
+   const auto config = config_repository_.getValidatedConfig(config_file);
+   const auto metadata_columns = csv_reader.get_col_names();
+
+   for (const auto& config_column : config.schema.metadata) {
```
fengelniederhammer, line 17: See Chaoran's comment in the ticket: I think we should also check the other way round and maybe write a warning to the log file that there are additional columns?

fengelniederhammer (approved):

PR #111, opened 2023-05-30
feat: generalize string columns in metadata store

### LINKED ISSUE #84: Specify column in `DateBetween` filter
Currently, the column of the `DateBetween` filter expression is implicitly hardcoded to "date".

The should be changed such that `DateBetween` accepts a `column` field to specify the column.

### LINKED ISSUE #101: Generalize the metadata store

fengelniederhammer:
resolves #101
resolves #84 

This is not yet finished, but I'd appreciate early feedback: 

For serialization, I think it's easier to let every indexed column have it's own little dictionary instead of a reference to the global dictionary. What do you think of the idea?


src/silo/query_engine/filter_expression/negation.cpp
```diff
@@ -31,7 +31,7 @@ std::unique_ptr<operators::Operator> Negation::compile(
       return child_operator;
    }
    if (child_operator->type() == operators::SELECTION) {
-      dynamic_cast<operators::Selection*>(child_operator.get())->negate();
+      dynamic_cast<operators::Selection<uint64_t>*>(child_operator.get())->negate();
```
fengelniederhammer, line 34: This is a little hacky, since we need to cast the operator to some selection where the generic type does not matter...

Taepper: I agree and in the long run we probably need something better but did you test this? I would expect the `dynamic_cast` to return a nullptr if the selection is not of type `uint64_t`

fengelniederhammer: It does work.

```
{
  "action": {
    "type": "Aggregated"
  },
  "filterExpression": {
    "type": "Not",
    "child": {
      "type": "StringEquals",
      "column": "division",
      "value": "Solothurn"
    }
  }
}
```
returns the expected result.

src/silo/query_engine/filter_expression/negation.cpp
```diff
@@ -31,7 +31,7 @@ std::unique_ptr<operators::Operator> Negation::compile(
       return child_operator;
    }
    if (child_operator->type() == operators::SELECTION) {
-      dynamic_cast<operators::Selection*>(child_operator.get())->negate();
+      dynamic_cast<operators::Selection<std::string>*>(child_operator.get())->negate();
```
Taepper, line 34: I would think that this also breaks if the selection is not of type `std::string`

Taepper (approved):
This looks good. I will separately deal with the Negation case of Selections, and can rebase+commit them once you have no WIP changes

PR #115, opened 2023-06-05
fix: overloading the negate method to avoid casting operators

Taepper:
As a followup to #111. Every operator gets a copy and negate method that is overloaded to compilation logic


src/silo/query_engine/query_engine.cpp
```diff
@@ -33,7 +33,16 @@ QueryEngine::QueryEngine(const silo::Database& database)
...
 // NOLINTNEXTLINE(readability-function-cognitive-complexity)
 response::QueryResult QueryEngine::executeQuery(const std::string& query) const {
-   nlohmann::json json = nlohmann::json::parse(query);
+   nlohmann::json json;
+   try {
+      json = nlohmann::json::parse(query);
+   } catch (QueryParseException qe) {
+      throw qe;
+   } catch (const nlohmann::json::parse_error& ex) {
+      throw QueryParseException("The query was not a valid JSON: parse_error");
```
fengelniederhammer, line 42: Why don't we include `ex.what()` here?

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
```diff
@@ -0,0 +1,98 @@
...
+   )
+   CHECK_SILO_QUERY(
+      json.contains("symbol"), "The field 'symbol' is required in a NucleotideEquals expression"
+   )
+   CHECK_SILO_QUERY(
+      json["symbol"].is_string(),
+      "The field 'symbol' in a NucleotideEquals expression needs to be a string"
+   )
+   const uint32_t position = json["position"].get<uint32_t>() - 1;
+   const std::string& nucleotide_symbol = json["symbol"];
```
fengelniederhammer, line 94: Don't we want to verify that this is actually a valid nucleotide symbol?

src/silo/query_engine/query_engine.cpp
```diff
@@ -7,21 +7,14 @@
 
 // query_parse_exception.h must be before the RAPIDJSON_ASSERT because it is used there
 #include "silo/query_engine/query_parse_exception.h"
```
fengelniederhammer, line 9: Please also delete that comment now and move the include in line with the other SILO includes.

src/silo/query_engine/query_engine.cpp
```diff
@@ -40,20 +33,29 @@ QueryEngine::QueryEngine(const silo::Database& database)
...
 // NOLINTNEXTLINE(readability-function-cognitive-complexity)
 response::QueryResult QueryEngine::executeQuery(const std::string& query) const {
-   rapidjson::Document json_document;
-   json_document.Parse(query.c_str());
-   if (!json_document.HasMember("filterExpression") || !json_document["filterExpression"].IsObject() ||
-       !json_document.HasMember("action") || !json_document["action"].IsObject()) {
+   nlohmann::json json;
+   try {
+      json = nlohmann::json::parse(query);
+   } catch (QueryParseException qe) {
```
fengelniederhammer, line 39: "Clang-Tidy: Catch handler catches by value; should catch by reference instead"

src/silo/query_engine/query_engine.cpp
```diff
@@ -40,20 +33,29 @@ QueryEngine::QueryEngine(const silo::Database& database)
...
    std::vector<std::string> compiled_queries(database.partitions.size());
 
    response::QueryResult query_result;
-   std::unique_ptr<filters::Expression> filter;
+   std::unique_ptr<silo::query_engine::filter_expressions::Expression> filter;
    {
       BlockTimer const timer(query_result.parse_time);
-      filter = query_engine::parseExpression(database, json_document["filterExpression"], 0);
+      filter = json["filterExpression"]
+                  .get<std::unique_ptr<silo::query_engine::filter_expressions::Expression>>();
```
fengelniederhammer, line 58: :+1: :)

include/silo/query_engine/filter_expressions/expression.h
```diff
@@ -15,17 +17,27 @@ struct DatabasePartition;
...
-      const DatabasePartition& database_partition
+      const DatabasePartition& database_partition,
+      AmbiguityMode mode
    ) const = 0;
 };
 
+// NOLINTNEXTLINE(invalid-case-style)
+void from_json(const nlohmann::json& json, std::unique_ptr<Expression>& filter);
+
+std::unique_ptr<Expression> parse(const nlohmann::json& json);
```
fengelniederhammer, line 37: Looks like this is actually unused, isn't it?

fengelniederhammer (approved):
Looks good. Just a couple of minor comments.

I think it might have made sense to split the two changes into separate PRs. And maybe rebase on main before merging ;)

src/silo/query_engine/operators/complement.cpp
```diff
@@ -46,8 +44,16 @@ Type Complement::type() const {
...
+   result->flip(0, row_count);
    return result;
 }
 
+std::unique_ptr<Operator> Complement::copy() const {
+   return std::make_unique<Complement>(child->copy(), row_count);
+}
+
+std::unique_ptr<Operator> Complement::negate() const {
+   return child->copy();
```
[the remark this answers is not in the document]
Taepper: Because it is a const function

src/silo/query_engine/operators/index_scan.cpp
```diff
@@ -1,15 +1,17 @@
...
-}
+IndexScan::IndexScan(const roaring::Roaring* bitmap, uint32_t row_count)
+    : bitmap(bitmap),
+      row_count(row_count) {}
 
 IndexScan::~IndexScan() noexcept = default;
 
 std::string IndexScan::toString() const {
-   // TODO think about passing strings for debug printing?
+   // TODO(someone) think about passing strings for debug printing?
```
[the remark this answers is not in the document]
Taepper: It is not yet clear what strings we want to pass

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
Taepper: But where would we get that information from?

src/silo/query_engine/operators/complement.cpp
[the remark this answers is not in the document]
Taepper: and we cannot have two `unique_ptr`s to the same object

src/silo/query_engine/operators/index_scan.cpp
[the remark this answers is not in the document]
Taepper: After we are certain that the column format does no longer change we can decide on what debug info to pass

Issue #69, opened 2023-04-02
Extract preprocessing from database


Taepper:
The structs 
`std::unique_ptr<PangoLineageCounts> pango_descriptor;`
and
`std::unique_ptr<Partitions> partition_descriptor;`
are only necessary for the preprocessing. These need to be removed from the `database` class

PR #121, opened 2023-06-12
feat: more config fields are being read, slight adjustment to columns

Taepper:
Many places needed to be touched in order for the next PRs which add more query actions for the additional endpoints. Also cleaned minor things up in a few places


include/silo/storage/column/column.h
```diff
@@ -0,0 +1,16 @@
...
+#define SILO_COLUMN_H
+
+#include <cstddef>
+#include <cstdint>
+
+namespace silo::storage::column {
+
+class Column {
+  public:
+   virtual std::string getAsString(std::size_t idx) const = 0;
```
fengelniederhammer, line 11: C++ has too many int types... Should we use `size_t` more consistently? I think we should agree on something^^

src/silo/common/date.cpp
```diff
@@ -0,0 +1,21 @@
...
+   const uint32_t date_value = (year << 16) + (month << 12) + day;
+   return Date{date_value};
+}
+
+std::string silo::common::dateToString(silo::common::Date date) {
+   const uint32_t year = date >> 16;
+   const uint32_t month = (date >> 12) & 0xF;
+   const uint32_t day = date & 0xFFF;
+   return std::to_string(year) + "-" + std::to_string(month) + "-" + std::to_string(day);
+}
```
fengelniederhammer, line 21: Please add a unit test (especially for error handling of `stringToDate` since this has to process user input.)

src/silo/config/database_config.cpp
```diff
@@ -1,11 +1,16 @@
...
 namespace silo::config {
 
-DatabaseMetadataType toDatabaseMetadataType(std::string_view type) {
-   if (type == "string") {
+DatabaseMetadataType toDatabaseMetadataType(std::string_view type, bool generate_index) {
+   if (type == "string" && generate_index) {
+      return DatabaseMetadataType::INDEXED_STRING;
+   }
+   if (type == "string" && !generate_index) {
       return DatabaseMetadataType::STRING;
```
fengelniederhammer, line 14: So, do we want to hide the difference between `INDEXED_STRING` and `STRING` from the user? We could also simply make the user configure this directly instead of the additional field `generateIndex`.

src/silo/storage/metadata_store.cpp
```diff
@@ -43,28 +41,39 @@ unsigned MetadataStore::fill(
...
-            item.name == "date" ? storage::column::DateColumn(true)
-                                : storage::column::DateColumn(false)
+            item.name == database_config.schema.date_to_sort_by ? storage::column::DateColumn(true)
+                                                                : storage::column::DateColumn(false)
          );
       }
    }
 }
 
+const storage::column::Column& MetadataStore::getColumn(const config::DatabaseMetadata& metadata
```
fengelniederhammer, line 62: Why is this unused? Is it preparation for something else?

src/silo/storage/metadata_store.cpp
```diff
@@ -43,28 +41,39 @@ unsigned MetadataStore::fill(
...
+   if (metadata.type == config::DatabaseMetadataType::INDEXED_STRING) {
+      return indexed_string_columns.at(metadata.name);
+   }
+   if (metadata.type == config::DatabaseMetadataType::PANGOLINEAGE) {
+      return pango_lineage_columns.at(metadata.name);
+   }
+   if (metadata.type == config::DatabaseMetadataType::DATE) {
+      return date_columns.at(metadata.name);
+   }
+   throw "fill failed";
```
fengelniederhammer, line 76: Let's not throw a string ;)

include/silo/storage/column/pango_lineage_column.h
```diff
@@ -15,11 +15,11 @@ class PangoLineageColumn : public Column {
   public:
    template <class Archive>
    [[maybe_unused]] void serialize(Archive& archive, const unsigned int /* version */) {
-      archive & values;
-      archive & value_id_lookup;
-      archive & id_value_lookup;
-      archive & indexed_values;
-      archive & indexed_sublineage_values;
+      archive& values;
```
fengelniederhammer, line 18: You could simply squash this commit into the previous one -> less noise in the Git history.

conanfile.py
```diff
@@ -9,7 +9,6 @@ class SiloRecipe(ConanFile):
         "boost/1.81.0",
         "poco/1.12.4",
         "onetbb/2021.7.0",
-        "rapidjson/cci.20220822",
```
fengelniederhammer, line 12: Good catch :)

include/silo/storage/column/pango_lineage_column.h
```diff
@@ -7,22 +7,27 @@
...
       archive& value_id_lookup;
+      archive& id_value_lookup;
       archive& indexed_values;
+      archive& indexed_sublineage_values;
    }
 
   private:
+   std::vector<uint32_t> values;
    std::unordered_map<common::PangoLineage, uint32_t> value_id_lookup;
+   std::vector<common::PangoLineage> id_value_lookup;
```
fengelniederhammer, line 28: How much performance benefit is there from using a `vector` over an `unordered_map`? We are essentially misusing the vector for a map. Using a map would make the code more obvious.

fengelniederhammer, line 28:
```suggestion
   std::vector<common::PangoLineage> value_lookup;
```

or maybe

```suggestion
   std::vector<common::PangoLineage> value_id_to_value_lookup;
```
?

include/silo/storage/column/pango_lineage_column.h
```diff
@@ -7,22 +7,27 @@
...
    [[maybe_unused]] void serialize(Archive& archive, const unsigned int /* version */) {
+      archive& values;
       archive& value_id_lookup;
+      archive& id_value_lookup;
       archive& indexed_values;
+      archive& indexed_sublineage_values;
    }
 
   private:
+   std::vector<uint32_t> values;
```
fengelniederhammer, line 26:
```suggestion
   std::vector<uint32_t> value_ids;
```
Wouldn't that be more correct?

include/silo/storage/column/string_column.h
```diff
@@ -24,31 +26,37 @@ class RawStringColumn {
...
       archive& value_id_lookup;
+      archive& id_value_lookup;
       archive& indexed_values;
-      archive& sequence_count;
    }
 
   private:
+   std::vector<uint32_t> values;
    std::unordered_map<std::string, uint32_t> value_id_lookup;
+   std::vector<std::string> id_value_lookup;
```
fengelniederhammer, line 49: (See comments in `pango_lineage_column.h`)

fengelniederhammer (changes requested):
The unit test is missing, otherwise it looks good.

src/silo/storage/metadata_store.cpp
Taepper: Yes, it is used by the code for the aggregation endpoint

src/silo/storage/metadata_store.cpp
Taepper: Ah yes, I added a TODO comment in a following commit, because I was unsure what exception to throw here, as it is more of a helper function. Thinking about it, the way to go is probably returning an `std::optional` instead of callers catching an intermediate exception and then needing to rethrow a more specific exception.

fengelniederhammer: An optional sounds like a good idea :+1: 
Optionals and such are usually a good idea in expected error cases that should be recoverable. Exceptions are better suited for unexpected, unrecoverable cases.

src/silo/config/database_config.cpp
Taepper: I guess this just makes it a lot simpler, afterall we prohibit non-indexing of pango lineages anyways. But maybe not hiding this and instead adding another helper function that deduces the used column type from the field `type` and `generate_index` is better afterall?

include/silo/storage/column/pango_lineage_column.h
Taepper: I wanted to keep it similar to the above `value_id_lookup` but can see that it may be confusing

Taepper: I would think this is fine. Especially as it is the simpler data structure.

fengelniederhammer: Ah, my intention behind the name there was "a lookup for value ids", not a "value to id lookup" ;)

src/silo/storage/metadata_store.cpp
Taepper: Well I guess the exception is not recoverable in the path anyways, as the query will be aborted anyways. So I went with a custom exception instead

fengelniederhammer (changes requested):

src/silo/storage/column/pango_lineage_column.cpp
```diff
@@ -5,59 +5,54 @@ namespace silo::storage::column {
...
       return {};
    }
 
-   return indexed_sublineage_values[value_id_lookup.at(value)];
+   return indexed_sublineage_values[value_to_id_lookup.at(value)];
 }
 
+std::string PangoLineageColumn::getAsString(std::size_t idx) const {
+   return id_to_value_lookup[value_ids[idx]].value;
+};
```
fengelniederhammer, line 56: Let's also add a small unit test for this function. The version for `IndexedStringColumn` doesn't work, but I'll fix that in my next PR.

Taepper: What do you mean that it does not work? What should the test encompass?

fengelniederhammer: The test should be:
* insert a value into the column
* get its string value by its index

Those indexed columns are relatively complicated. In the `IndexedStringColumn`, the `id_to_value_lookup` is not updated.

src/silo/storage/column/pango_lineage_column.cpp
```diff
@@ -5,59 +5,54 @@ namespace silo::storage::column {
...
+   if (!value_to_id_lookup.contains(value)) {
       return {};
    }
 
-   return indexed_sublineage_values[value_id_lookup.at(value)];
+   return indexed_sublineage_values[value_to_id_lookup.at(value)];
 }
 
+std::string PangoLineageColumn::getAsString(std::size_t idx) const {
+   return id_to_value_lookup[value_ids[idx]].value;
```
fengelniederhammer, line 55:
```suggestion
   return id_to_value_lookup.at(value_ids.at(idx)).value;
```

src/silo/storage/column/pango_lineage_column.cpp
Taepper: Ah yes, that fix is already in #122 
Will commit that test shortly

Taepper:
Also resolves #69

fengelniederhammer:
I'm not sure whether it's a good idea to have custom date parsing. If we want to validate dates, we need to know that e.g. 2023-06-31 is not a valid date, etc. I think I would prefer using some library that takes care of this.

src/silo/storage/column/string_column.cpp
```diff
@@ -12,25 +12,37 @@ const std::vector<std::string>& RawStringColumn::getValues() const {
...
 
 void IndexedStringColumn::insert(const std::string& value) {
-   if (!value_id_lookup.contains(value)) {
-      value_id_lookup[value] = value_id_lookup.size();
+   uint32_t value_id;
+   if (!value_to_id_lookup.contains(value)) {
+      value_id = value_to_id_lookup.size();
+      value_to_id_lookup[value] = value_id;
+      id_to_value_lookup.push_back(value);
       indexed_values.emplace_back();
```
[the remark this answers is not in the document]
JonasKellerer: https://yasenh.github.io/post/cpp-diary-1-emplace_back/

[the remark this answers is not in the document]
fengelniederhammer: C++ and it's implicit copies and moves...

This sounds like we should actually use `emplace_back` here?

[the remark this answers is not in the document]
Taepper: `push_back` is certainly just as valid. This is more a philosophical question. A copy is made nonetheless, so it does not matter. I think `push_back` is the clearer choice here. See also abseil tip 112: https://abseil.io/tips/112
Still a detail

fengelniederhammer (approved):
I'm still not convinced that having custom date parsing is a good idea, but I don't want to block on this for too long. If we notice that this was a bad idea, it should not be too hard to revert it.

include/silo/database.h
```diff
@@ -61,7 +60,10 @@ class Database {
 
    [[maybe_unused]] void naiveIndexAllNucleotideSymbolsN();
 
-   [[maybe_unused]] void saveDatabaseState(const std::string& save_directory);
+   [[maybe_unused]] void saveDatabaseState(
+      const std::string& save_directory,
```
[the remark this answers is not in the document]
Taepper: This function is currently unused, this just keeps it compiling. Isnt this a task for when preprocessing is restructured?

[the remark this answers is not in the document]
JonasKellerer: Ah. Sry. I thought you touched in on your pr. Yes, it should be on the new preprocessing task.

src/silo/preprocessing/partition.cpp
```diff
@@ -140,7 +140,7 @@ Partitions buildPartitions(PangoLineageCounts pango_lineage_counts, Architecture
 
          // Merge pango_lineages, such that chunks are not get very small
          descriptor.partitions[0].chunks.emplace_back();
-         for (auto& pango : pango_lineage_counts.pango_lineage_counts) {
+         for (const auto& pango : pango_lineage_counts.pango_lineage_counts) {
```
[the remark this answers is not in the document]
Taepper: A variable ending in count count would make me expect a number type, so maybe it is not the best name either? As above I would say we should reconsider this function generally in the partition refactor.

Issue #110, opened 2023-05-27
Order by and limit actions 


chaoran-chen:
Supporting order by and limit would be very useful because we are often interested in e.g. the most common variants, the countries with the most sequences, the mutations with the largest proportions, etc.

For order by, in addition to sorting by a (combination of) field(s), we would also like to sort randomly. This would be useful for randomly selecting a subset of the data (the current LAPIS supports it for some endpoints).

PR #122, opened 2023-06-12
Full aggregation support

### LINKED ISSUE #90: Implement `groupByFields` for `Aggregated` action
Implement the following:
* The `Aggregated` action accepts a field `groupByFields` which is an array of strings.
* When not specified, the `Aggregated` action will return a total count (current state)
* When specified, the `Aggregated` action will return multiple count items with the count stratified by the values of the specified fields (similar to a Pivot table).

Taepper:
Add full support for arbitrary aggregation queries resolves #90 #110


src/silo/query_engine/query_engine.cpp
```diff
@@ -74,86 +75,20 @@ response::QueryResult QueryEngine::executeQuery(const std::string& query) const
    for (unsigned i = 0; i < database.partitions.size(); ++i) {
       SPDLOG_DEBUG("Simplified query for partition {}: {}", i, compiled_queries[i]);
    }
-   LOG_PERFORMANCE("Execution (filter): {} microseconds", std::to_string(query_result.filter_time));
+   LOG_PERFORMANCE("Execution (filter): {} microseconds", std::to_string(filter_time));
+
+   auto action = json["action"].get<std::unique_ptr<silo::query_engine::actions::Action>>();
```
fengelniederhammer, line 80: If this fails, we already did the potentially expensive filtering of the DB -> Let's move this up?

I'd even propose to create another struct `SiloQuery { filter_expression, action }` and simply call `json.get<SiloQuery>()`.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+#include "silo/database.h"
+#include "silo/query_engine/operator_result.h"
+#include "silo/query_engine/query_parse_exception.h"
+#include "silo/query_engine/query_result.h"
+#include "silo/storage/metadata_store.h"
+
+namespace silo::query_engine::actions {
+
+struct Tuple {
```
fengelniederhammer, line 23:
```suggestion
struct RowValues {
```
or 
```suggestion
struct RowValuesOfInterest {
```
?
The second one is a bit lengthy, but makes the purpose clear.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+      for (const std::string& value : tuple.string_columns) {
+         auto hash = std::hash<std::string>()(value);
+         // NOLINTBEGIN(readability-magic-numbers)
+         hash = ((hash >> 16) ^ hash) * 0x45d9f3b;
+         hash = ((hash >> 16) ^ hash) * 0x45d9f3b;
+         hash = (hash >> 16) ^ hash;
+         seed ^= hash + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+         // NOLINTEND(readability-magic-numbers)
+      }
+      return seed;
```
fengelniederhammer, line 202: I don't know much about hash functions, but
* why is this so complicated?
* why does this not involve `int_columns`?

From a naive point of view I would have expected that this function looks somewhat like

```
return hash(string_columns) + hash(int_columns);
```

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+      }
+      result.push_back({fields});
+   }
+   return result;
+}
+
+}  // namespace silo::query_engine::actions
+
+template <>
+struct std::hash<silo::query_engine::actions::Tuple> {
```
fengelniederhammer, line 190: Let's maybe move this up so that it sits closer to `Tuple`?

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+      return QueryResult{std::vector<AggregationResult>{{tuple_fields}}};
+   }
+   if (group_by_fields.size() == 1) {
+      const std::string group_by_field = group_by_fields[0];
+      const auto& metadata = database.database_config.getMetadata(group_by_field);
+      if (metadata.name == database.database_config.schema.primary_key) {
+         throw QueryParseException("Cannot group by primary key field: '" + group_by_field + "'");
+      }
+      // TODO(taepper) optimize when equal to partition_by field
+      // TODO(taepper) optimize single field groupby
```
fengelniederhammer, line 236: Please add an issue number instead of a name. Experience shows that TODOs without ticket numbers are very likely to stay in the code forever.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+      for (auto& [key, value] : map) {
+         final_map[key] += value;
+      }
+   }
+   std::vector<Tuple> tuples;
+   for (auto& [tuple, count] : final_map) {
+      tuples.push_back(tuple);
+      tuples.back().addField(count);
+   }
+   group_by_metadata.push_back({"count", config::ValueType::NUMBER, false});
```
fengelniederhammer, line 279: This looks hacky^^ Is it worth creating a separate struct, so that we don't have to fall back on `DatabaseMetadata` here for a column that doesn't exist?

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+   if (randomize_order) {
+      // obtain a time-based seed:
+      const uint32_t seed = std::chrono::system_clock::now().time_since_epoch().count();
+      std::default_random_engine rng(seed);
+      std::shuffle(tuples.begin(), tuples.end(), rng);
+   }
+   if (!order_by_definition.empty()) {
+      applyOrderBy(tuples, order_by_definition);
+   }
```
fengelniederhammer, line 288: The code read like it's possible to randomize AND order by. Is that correct?

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+   std::unordered_map<Tuple, uint32_t> final_map;
+   for (auto& map : maps) {
+      for (auto& [key, value] : map) {
+         final_map[key] += value;
+      }
+   }
+   std::vector<Tuple> tuples;
+   for (auto& [tuple, count] : final_map) {
+      tuples.push_back(tuple);
+      tuples.back().addField(count);
```
fengelniederhammer, line 277: This also feels very hacky. How about using a struct `RowValuesCount { count , RowValues }` instead? It doesn't need all the fancy functionality that `Tuple` (aka `RowValues`)  has, but could be a plain struct, as far as I can see.

fengelniederhammer: The we could also get rid of `group_by_metadata.push_back({"count", config::ValueType::NUMBER, false});`, coudn't we?

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,304 @@
...
+   }
+   group_by_metadata.push_back({"count", config::ValueType::NUMBER, false});
+   if (randomize_order) {
+      // obtain a time-based seed:
+      const uint32_t seed = std::chrono::system_clock::now().time_since_epoch().count();
```
fengelniederhammer, line 282:
```suggestion
      const uint32_t time_based_seed = std::chrono::system_clock::now().time_since_epoch().count();
```
:)

src/silo/query_engine/actions/aggregated.cpp
Taepper: I want tuple. This is standard

src/silo/query_engine/actions/aggregated.cpp
Taepper: The metadata is the descriptor of what values are present in the tuple. This is by far the simplest solution

src/silo/query_engine/actions/aggregated.cpp
Taepper: Yes, this comes for free and offers great functionality if the user wants to shuffle within key ranges

src/silo/query_engine/actions/aggregated.cpp
fengelniederhammer: But why do we need to implement vector hashing ourselves? Isn't this already implemented in the standard lib?

src/silo/query_engine/actions/aggregated.cpp
Taepper: This is more difficult that it may appear. This is working and very simple. I would keep it like this for now, unless we have a very good reason not to or until we have other requirements that will lead to us needing something like this

src/silo/query_engine/query_engine.cpp
Taepper: In a later PR, when we have the actions implemented as well?

src/silo/query_engine/actions/aggregated.cpp
fengelniederhammer: The metadata is the descriptor of which values are present in metadata columns. Tuples are not metadata columns (they only originate from them).

IMO: Not confusing developers is a lot more important than the "simple solution". A new struct generates just a little more code and should not have performance impact.

Taepper: But what would be in that struct? Another variant of either a column value or synthetic types that are created during the aggregation?

Taepper: Yes this would work and was the initial solution, but what if the user orders by count? Which is a common use-case imo

Taepper: That is why I needed to seamlessly include the `count` variable into the tuple, otherwise *all* later logic would need to hard code the count column and it was not feasible to create a working solution

src/silo/query_engine/actions/aggregated.cpp
Taepper: Not really, I am sure there is a boost lib for that, but this is more performant. I refactored it and push the changes again. Maybe this makes the whole hashing topic more self-contained

src/silo/query_engine/actions/aggregated.cpp
fengelniederhammer: But let's go for "Either hardcode it there, that we always retrieve `count`". This should also work with ordering by count, doesn't it?

Taepper: I do not think that is easily possible.

But what is the problem with the current solution? This is nothing out of the ordinary

fengelniederhammer: The current solution is very hacky. Obvious code makes it easier for future developers (including us in a couple of weeks).

src/silo/query_engine/query_engine.cpp
fengelniederhammer: Or maybe simply in a separate commit? I don't see a reason why we need to wait for the other actions?

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,300 @@
...
+};
+
+std::string strToLower(std::string s) {
+   std::transform(s.begin(), s.end(), s.begin(), [](unsigned char character) {
+      return std::tolower(character);
+   });
+   return s;
+}
+
+TupleAccessor parseOrderByField(
```
JonasKellerer, line 96: Could we test this somehow? I feel, that a test woul make the idea clearer.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,300 @@
...
+#include "silo/query_engine/query_result.h"
+#include "silo/storage/metadata_store.h"
+
+using silo::common::Hasher;
+
+namespace silo::query_engine::actions {
+
+struct Tuple {
+   std::vector<std::string> string_columns;
+   std::vector<uint32_t> int_columns;
```
JonasKellerer, line 28: Are these the values in the int_columns/string_columns? Should we name it so it repersents that?

JonasKellerer (commented):
For me, most of it is naming. And encapsulating ideas, so that the level of detail is the same in each function.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,302 @@
+#include "silo/query_engine/actions/aggregated.h"
+
+#include <algorithm>
```
fengelniederhammer, line 3: This seems unused now

include/silo/storage/column/column.h
```diff
@@ -7,10 +7,7 @@
 
 namespace silo::storage::column {
 
-class Column {
-  public:
-   virtual std::string getAsString(std::size_t idx) const = 0;
-};
+class Column {};
```
fengelniederhammer, line 10: Why do we need that class at all?

include/silo/storage/column/indexed_string_column.h
```diff
@@ -0,0 +1,72 @@
...
+
+class IndexedStringColumnPartition : public Column {
+   friend class boost::serialization::access;
+
+  private:
+   template <class Archive>
+   [[maybe_unused]] void serialize(Archive& archive, const unsigned int /* version */) {
+      // clang-format off
+      archive & value_ids;
+      archive & indexed_values;
```
fengelniederhammer, line 30:
```suggestion
      archive& value_ids;
      archive& indexed_values;
```
for consistent formatting

include/silo/storage/column/indexed_string_column.h
```diff
@@ -0,0 +1,72 @@
...
+
+   [[nodiscard]] roaring::Roaring filter(const std::string& value) const;
+
+   void insert(const std::string& value);
+
+   const std::vector<silo::Idx>& getValues() const;
+};
+
+class IndexedStringColumn : public Column {
+   friend class boost::serialization::access;
```
fengelniederhammer, line 49: Why do we actually need that friend class here? We didn't really test it yet, but at least it compiled without it where I implemented the serialize thingy.

include/silo/storage/database_partition.h
```diff
@@ -10,6 +10,10 @@
 #include "silo/storage/metadata_store.h"
 #include "silo/storage/sequence_store.h"
 
+namespace silo {
+struct Database;
+}
```
fengelniederhammer, line 15: Why do we need this here? Why do we need `friend class Database`? I'd actually expect that the partition does not depend on the db. Instead the db depends on the partition.

include/silo/storage/metadata_store.h
```diff
@@ -13,46 +13,45 @@
...
-
 class PangoLineageAliasLookup;
 
+namespace storage::column {
+struct DateColumnPartition;
+struct IndexedStringColumnPartition;
+struct IntColumnPartition;
+struct PangoLineageColumnPartition;
+struct StringColumnPartition;
+}  // namespace storage::column
```
fengelniederhammer, line 26: Why is a forward declaration possible here? I'd expect that the compiler has to know the size of those classes to determine the size of the fields.

src/silo/common/bidirectional_map.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+template <typename V>
+BidirectionalMap<V>::BidirectionalMap()
+    : id_to_value(),
+      value_to_id() {}
+
+template <typename V>
+V BidirectionalMap<V>::getValue(Idx idx) const {
+   return id_to_value.at(idx);
+}
```
fengelniederhammer, line 18: Why does `getId` return an optional and this returns the value itself? I'd expect both to return an optional.

src/silo/config/database_config.cpp
```diff
@@ -44,4 +46,16 @@ ColumnType DatabaseMetadata::getColumnType() const {
...
 }
 
+DatabaseMetadata DatabaseConfig::getMetadata(const std::string& name) const {
+   auto element = std::find_if(
+      std::begin(schema.metadata),
+      std::end(schema.metadata),
+      [&name](const auto& metadata) { return metadata.name == name; }
+   );
+   if (element == std::end(schema.metadata)) {
+      throw std::runtime_error("Unknown metadata type: " + std::string(name));
```
fengelniederhammer, line 56: I find type confusing, because seeing this as a user I would expect that I got the type STRING, PANGO_LINEAGE, INT, etc. wrong.
```suggestion
      throw std::runtime_error("Did not find metadata with name: " + std::string(name));
```

src/silo/database.cpp
```diff
@@ -496,16 +493,79 @@ void silo::Database::preprocessing(
...
+                          : storage::column::DateColumn(false);
+         for (size_t i = 0; i < partitions.size(); ++i) {
+            partitions[i].insertColumn(item.name, column.createPartition());
+         }
+         date_columns.emplace(item.name, column);
+      } else if (column_type == config::ColumnType::INT) {
+         storage::column::IntColumn column;
+         for (size_t i = 0; i < partitions.size(); ++i) {
+            partitions[i].insertColumn(item.name, column.createPartition());
+         }
```
fengelniederhammer, line 538: Clang Tidy suggests to use range-based loops here: `for (auto & partition : partitions)`

include/silo/database.h
```diff
@@ -70,9 +83,23 @@ class Database {
 
    [[nodiscard]] const PangoLineageAliasLookup& getAliasKey() const;
 
+   const silo::storage::column::DateColumn& getDateColumn(std::string name) const;
```
fengelniederhammer, line 86:
```suggestion
   const silo::storage::column::DateColumn& getDateColumn(const std::string& name) const;
```
According to clang tidy, this saves some copies.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+   bool operator==(const Tuple& other) const { return this->data == other.data; }
+};
+
+}  // namespace silo::query_engine::actions
+
+template <>
+struct std::hash<silo::query_engine::actions::Tuple> {
+   std::size_t operator()(const silo::query_engine::actions::Tuple& tuple) const {
+      const std::string_view strView(tuple.data.data(), tuple.data.size());
```
fengelniederhammer, line 127: Clang-Tidy: Invalid case style for variable 'strView'

endToEndTests/test/queries/GroupByLineageOrderByCountLimit.json
```diff
@@ -0,0 +1,32 @@
+{
+  "testCaseName": "Group all sequences by lineage and order by count descending, with limit 4",
+  "query": {
+    "action": {
+      "type": "Aggregated",
+      "groupByFields": ["pango_lineage"],
+      "orderByFields": ["count desc"],
```
fengelniederhammer, line 7: We could also think about passing the information like this:
```suggestion
      "orderByFields": [{ "fields": "count", "descending": true }]
```
With descending being false by default.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,310 @@
...
+   bool randomize_order = false;
+   for (const std::string& order_by_field : order_by_fields) {
+      if (order_by_field == "random()") {
+         randomize_order = true;
+         break;
+      }
+      order_by_definition.push_back(parseOrderByField(order_by_field));
+   }
+   for (const OrderByField& field : order_by_definition) {
+      if (field.name != "count" && !std::any_of(group_by_metadata.begin(), group_by_metadata.end(), [&field](const config::DatabaseMetadata& group_by_field) {
```
fengelniederhammer, line 256: Maybe extract a constant for `"count"` as this appears several times in this file? That would make the connection a lot clearer.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,310 @@
...
+) {
+   std::vector<AggregationResult> result;
+   result.reserve(tuple_counts.size());
+   for (auto& [tuple, count] : tuple_counts) {
+      std::map<std::string, std::variant<std::string, int32_t, double>> fields =
+         tuple.getFields(database);
+      fields["count"] = static_cast<int32_t>(count);
+      result.push_back({fields});
+   }
+   return result;
```
fengelniederhammer, line 202: This looks a lot nicer :+1: (The whole file, but this function in particular)

src/silo/query_engine/query.cpp
```diff
@@ -0,0 +1,30 @@
...
+      nlohmann::json json = nlohmann::json::parse(query_string);
+      if (!json.contains("filterExpression") || !json["filterExpression"].is_object() ||
+          !json.contains("action") || !json["action"].is_object()) {
+         throw QueryParseException("Query json must contain filterExpression and action.");
+      }
+      filter = json["filterExpression"]
+                  .get<std::unique_ptr<silo::query_engine::filter_expressions::Expression>>();
+      action = json["action"].get<std::unique_ptr<silo::query_engine::actions::Action>>();
+   } catch (const QueryParseException& qe) {
+      throw qe;
```
fengelniederhammer, line 22: Why do we catch this exception if we only rethrow it?

src/silo/query_engine/query_engine.cpp
```diff
@@ -11,149 +11,59 @@
...
-
-   response::QueryResult query_result;
-   std::unique_ptr<silo::query_engine::filter_expressions::Expression> filter;
-   {
-      BlockTimer const timer(query_result.parse_time);
-      filter = json["filterExpression"]
-                  .get<std::unique_ptr<silo::query_engine::filter_expressions::Expression>>();
-      SPDLOG_DEBUG("Parsed query: {}", filter->toString(database));
-   }
+QueryResult QueryEngine::executeQuery(const std::string& query_string) const {
```
fengelniederhammer, line 30: Now this is a nice and readable function :)

src/silo/storage/column/indexed_string_column.test.cpp
```diff
@@ -0,0 +1,42 @@
...
+   under_test.insert("value 2");
+   under_test.insert("value 2");
+   under_test.insert("value 3");
+   under_test.insert("value 1");
+
+   // EXPECT_EQ(under_test.getAsString(0U), "value 1");
+   // EXPECT_EQ(under_test.getAsString(1U), "value 2");
+   // EXPECT_EQ(under_test.getAsString(2U), "value 2");
+   // EXPECT_EQ(under_test.getAsString(3U), "value 3");
+   // EXPECT_EQ(under_test.getAsString(4U), "value 1");
```
fengelniederhammer, line 41: Why is this commented out?

include/silo/database.h
```diff
@@ -36,6 +44,12 @@ class Database {
    std::unique_ptr<ReferenceGenome> reference_genome;
    std::vector<DatabasePartition> partitions;
 
+   std::unordered_map<std::string, storage::column::StringColumn> string_columns;
+   std::unordered_map<std::string, storage::column::IndexedStringColumn> indexed_string_columns;
+   std::unordered_map<std::string, storage::column::IntColumn> int_columns;
+   std::unordered_map<std::string, storage::column::DateColumn> date_columns;
+   std::unordered_map<std::string, storage::column::PangoLineageColumn> pango_lineage_columns;
```
fengelniederhammer, line 51: Sorry, another naming discussion: :)

Why are those "columns" called column? 
* The int column does nothing at all. -> Why do we need this at all?
* The date column stores `is_sorted` which is redundant. -> Why do we need this at all?
* The (indexed) string and pango lineage columns serve as lookup -> Bundle them in a class `ValueLookup` or `Dictionary` and rename them to something like `StringValueLookup`?

fengelniederhammer: Btw I expect the same for float columns. Probably we don't need the global float column, do we?

fengelniederhammer (commented):
Some details. Overall it looks good.

include/silo/database.h
Taepper: I see your points. The fact that they serve as factories for column partitions encapsulated the internal logic of columns and partitions away from the user which was quite nice in my opinion.

endToEndTests/test/queries/Exact.json
```diff
@@ -13,7 +13,9 @@
...
     }
   },
-  "expectedQueryResult": {
-    "count": 93
-  }
+  "expectedQueryResult": [
+    {
+      "count": 93
+    }
+  ]
```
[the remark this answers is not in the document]
Taepper: But the structure overall is as we want to have it? We need the vector for all groups of the result and do not want to encapsulate it into an object?

endToEndTests/test/queries/invalidAction.json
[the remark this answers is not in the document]
Taepper: Hmm, must have been by mistake

endToEndTests/test/queries/Exact.json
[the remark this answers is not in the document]
fengelniederhammer: Yes. This is more a note than a task for this PR.

include/silo/storage/column/indexed_string_column.h
Taepper: I thought it is more consistent style to have the serialization as a private member instead of static, where we then need the friend class

src/silo/common/bidirectional_map.cpp
Taepper: The way I thought about that:
The `Idx` is internal and if an `Idx` is obtained during the execution of the user query in any way, it must be valid and stay valid indefinitely

include/silo/storage/column/indexed_string_column.h
fengelniederhammer: Ok, valid :+1:

include/silo/database.h
fengelniederhammer: Ah, I didn't see that. That's a fair point to keep them then.

endToEndTests/test/queries/Exact.json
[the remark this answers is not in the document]
Taepper: Still just wanted open the possibility of direct feedback, in case you would have preferred a different structure

include/silo/common/types.h
```diff
@@ -0,0 +1,14 @@
...
+#define SILO_TYPES_H
+
+#include <stddef.h>
+#include <cstdint>
+
+namespace silo {
+// Referencing to silo internal pointers
+typedef uint32_t Idx;
+// Representing the ID of a row in the database
+typedef size_t TID;
```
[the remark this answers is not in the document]
Taepper: T actually stands for Tuple. tid is also used in Postgres:
https://www.postgresql.org/docs/current/datatype-oid.html

I will for now revert the introduction of a typedef, as I did not begin the discussed division/refactoring of the used (unsigned) integers types throughout SILO, to not blow up the size of the PR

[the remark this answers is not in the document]
Taepper: Both TupleID and TID is fine by me. Maybe let us see how readable it will be once the refactoring takes place and then decide

include/silo/storage/database_partition.h
Taepper: Because of one private member access which I now made public. This is still not perfectly polished now (but it regards preprocessing), but I guess it is better than this backwards dependence that you noticed

endToEndTests/test/queries/GroupByLineageOrderByCountLimit.json
Taepper: But sometimes the user may want to order ascending by some columns and descending by others (e.g. ascending year, descending count within year). I would not neccessarily want to bloat requests like that. It is more of a feature for advanced users where we get away with the current "easter egg"-like functionality

include/silo/storage/column/float_column.h
```diff
@@ -1,31 +1,57 @@
...
 #include <string>
 #include <vector>
 
-#include "silo/storage/column/column.h"
+namespace boost::serialization {
+struct access;
+}
+
+namespace silo {
+struct Database;
```
fengelniederhammer, line 13: Why do we need this here? This is again a cyclic dependency: database -> columns -> float column -> database.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+size_t getTupleSize(std::vector<config::DatabaseMetadata>& group_by_metadata) {
+   size_t size = 0;
+   for (const auto& metadata : group_by_metadata) {
+      if (metadata.getColumnType() == config::ColumnType::STRING) {
+         size += sizeof(common::String<common::STRING_SIZE>);
+      } else if (metadata.getColumnType() == config::ColumnType::FLOAT) {
+         size += sizeof(double);
+      } else if (metadata.getColumnType() == config::ColumnType::FLOAT) {
+         size += sizeof(int32_t);
```
fengelniederhammer, line 33:
```suggestion
      } else if (metadata.getColumnType() == config::ColumnType::INT) {
         size += sizeof(int32_t);
```
This looks like a copy paste error

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -0,0 +1,319 @@
...
+      result.resize(limit.value());
+   } else {
+      std::sort(result.begin(), result.end(), cmp);
+   }
+}
+const std::string COUNT_FIELD = "count";
+
+std::vector<AggregationResult> generateResult(
+   std::unordered_map<Tuple, uint32_t>& tuple_counts,
+   const Database& database
```
fengelniederhammer, line 198: This is now unused.

fengelniederhammer (approved):
Three issues to fix, then we can merge it :)

fengelniederhammer (changes requested):
```
  {
    "action": {
      "type": "Aggregated",
      "groupByFields": ["division"],
      "orderByFields": ["division", "count"]
    },
    "filterExpression": {
      "type": "True"
    }
  }
```
let's SILO crash with an address sanitizer error.

Taepper:
Oh, there was still an error that was a remnant from a debugging session. (`aggregated line 280: map = maps.local();` outside of `parallel_for`). This should have also fixed the ASAN error

fengelniederhammer (approved):

PR #124, opened 2023-06-13
Inject SILO's data instead of storing it in the Docker image

### LINKED ISSUE #108: Inject SILO's data instead of storing it in the Docker image
Currently we store the data that SILO uses at runtime in the Docker images. Change this to make it configurable: Docker should mount a directory into the running container that contains the data.

JonasKellerer:
Resolves #108


JonasKellerer:
I'll add a test :)

fengelniederhammer (changes requested):
As discussed: Let's restructure the `api.cpp` to first parse all arguments, pick a mode, and then run silo.

testBaseData/test_preprocessing_config.yaml
fengelniederhammer, line 1: When writing the documentation, I noticed that we should probably split this config:
A user should only need to configure 
```
inputDirectory: "./testBaseData/"
outputDirectory: "./output/"
metadataFilename: "small_metadata_set.tsv"
sequenceFilename: "small_sequence_set.fasta"
pangoLineageDefinitionFilename: "pango_alias.txt"
referenceGenomeFilename: "..."
```
Notice the `referenceGenomeFilename` which does not exist yet here ;)

`partitionFolder`, `sortedPartitionFolder` and `serializationFolder` is IMO internal stuff that a user should not need to configure and we can simply put that into the `outputDirectory`. Either we hardcode that, or we provide it in a config file that's stored in the Docker image (meaning a developer has to provide it, but not a maintainer).

README.md
```diff
@@ -54,11 +54,16 @@ docker build . --tag=silo
 Run docker container
 
 ```shell
-docker run -p 8081:8081 silo
+docker run -p 8081:8081 -v your/data/directory:the/directory/in/docker/image silo /siloApi --preprocessingConfig=the/directory/in/docker/image/test_preprocessing_config.yaml --databaseConfig=the/directory/in/docker/image/test_database_config.yaml -a
```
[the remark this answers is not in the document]
JonasKellerer: As discussed, we specify in which mode silo is run at start of the container.

README.md
```diff
@@ -54,11 +54,16 @@ docker build . --tag=silo
 Run docker container
 
 ```shell
-docker run -p 8081:8081 silo
+docker run -p 8081:8081 -v your/data/directory:the/directory/in/docker/image silo --api --preprocessingConfig=the/directory/in/docker/image/test_preprocessing_config.yaml --databaseConfig=the/directory/in/docker/image/test_database_config.yaml
 ```
 
+The mounted directory `your/data/directory` is the place on your machine, where the data for SILO is located. The data
+will be mounted to the directory `the/directory/in/docker/image` in the docker image. Both the preprocessingConfig and
```
Taepper, line 61: Do we really want the user to provide the location in the docker image? Can we not just use `data` here? At least it is not quite clear to me, when someone would actually change the path where the docker stores internal state

JonasKellerer: I had that in there, because we specify in the preprocessing config, where the data lies. But as you say it would be better, to keep that away from the user.

src/silo/storage/reference_genome.cpp
```diff
@@ -0,0 +1,39 @@
...
+   if (!std::filesystem::exists(reference_genome_path)) {
+      throw std::filesystem::filesystem_error(
+         "Global reference genome file " + reference_genome_path.relative_path().string() +
+            " does not exist",
+         std::error_code()
+      );
+   }
+
+   std::ifstream reference_file(reference_genome_path.string());
+   std::vector<std::string> global_reference;
```
Taepper, line 22: Maybe rename to `reference_genomes` as it is no longer a globally shared variable

Taepper (approved):

README.md
Taepper: I see that the test docker compose file sets it as `/app/testBaseData`, but maybe we stay even more generic, but still under `/app` if that is wanted?

Taepper: But I have no clear preference over `/app/data` or any other descriptive naming

JonasKellerer:
I changed the pathing now, so that the user does not have to specify, where the data is put in the docker container (so allways /app/data):
`docker run -p 8081:8081  -v your/data/directory:/app/data  silo --api --preprocessingConfig=./yourRelativePathTo/preprocessing_config.yaml --databaseConfig=./yourRelativePathTo/database_config.yaml`

In the config the user the specifies, where the data is located relative to the mounted folder, to make it as easy as possible. 

For the developer the tests are the same as before. So running build/silo_test will make the tests work. In clion, one has to specify that the working directory is `..`. Also the tests in the ci work now.

However, these changes come with a small drawback: When trying to run the production docker container locally with 
`docker run -v $(pwd)/testBaseData:/app/data -p 8081:8081 -t silo --api --preprocessingConfig=./test_preprocessing_config.yaml --databaseConfig=./test_database_config.yaml` it will throw an error: `filesystem error: ./testBaseData/ does not exist: No error information`. This is the case, because we mount testBaseData but in the provided `test_preprocessing_config.yaml`  it is specified that the data is in the folder `./testBaseData` which is of course not present, because it is mounted. So when trying to run the container for your self you have to have your own `preprocessing_config.yaml`, or use the new yaml `preprocessing_config_for_ci.yaml`. I think is okay, because when running silo one has to have own configs anyways.
What do you think @Taepper @chaoran-chen @fengelniederhammer   ?

chaoran-chen:
I think that's fine for now!

Regarding the data path: Should we make it maybe even shorter, just `/data`?

PR #132, opened 2023-06-15
Float columns

### LINKED ISSUE #83: Implement number data types
We need columns that store integers and floats. Both should be queryable via "equals" and via "between" (similar to a date).

fengelniederhammer:
resolves #83

### DIFF
--- a/include/silo/storage/column/float_column.h
+++ b/include/silo/storage/column/float_column.h
@@ -1,0 +1,33 @@
+#ifndef SILO_FLOAT_COLUMN_H
+#define SILO_FLOAT_COLUMN_H
+
+#include <string>
+#include <vector>
+
+#include "silo/storage/column/column.h"
+
+namespace silo::storage::column {
+
+class FloatColumn : public Column {
+  public:
+   template <class Archive>
+   [[maybe_unused]] void serialize(Archive& archive, const unsigned int /* version */) {
+      archive& values;
+   }
+
+  private:
+   std::vector<double> values;
+
+  public:
+   FloatColumn();
+
+   [[nodiscard]] const std::vector<double>& getValues() const;
+
+   void insert(double value);
+
+   [[nodiscard]] std::string getAsString(std::size_t idx) const override;
+};
+
+}  // namespace silo::storage::column
+
+#endif  // SILO_FLOAT_COLUMN_H
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,11 @@ name: LAPIS-SILO
 
 on:
   push:
-  pull_request:
+
+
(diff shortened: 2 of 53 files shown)


fengelniederhammer:
Do we also want a `FloatEquals`filter? What do you think @Taepper @chaoran-chen ?

include/silo/storage/column/float_column.h
```diff
@@ -0,0 +1,33 @@
...
+  public:
+   template <class Archive>
+   [[maybe_unused]] void serialize(Archive& archive, const unsigned int /* version */) {
+      archive& values;
+   }
+
+  private:
+   std::vector<double> values;
+
+  public:
```
JonasKellerer, line 21: Only a style thing: Do we want to have multiple sections of public and private in the class?

include/silo/storage/column/int_column.h
```diff
@@ -0,0 +1,33 @@
...
+  public:
+   template <class Archive>
+   [[maybe_unused]] void serialize(Archive& archive, const unsigned int /* version */) {
+      archive& values;
+   }
+
+  private:
+   std::vector<uint64_t> values;
+
+  public:
```
JonasKellerer, line 21: Same as above

JonasKellerer (approved):
Looks good. Only a minor formating question left.

include/silo/storage/column/float_column.h
fengelniederhammer: In general probably not, but in this case I consider the Archive thing boiler plate, that's simply on top of the class definition. 
Then we have the rest with a private and a public section.

We could change it, but that's how we have it everywhere AFAIK.

Taepper: I also think the grouping by private public is a good idea in general, but other groupings can outweigh that guideline such as Fabian's mentioned consideration

chaoran-chen:
@fengelniederhammer, re `FloatEquals`: sure, if it's not too much trouble, it's probably useful to have that as well.

fengelniederhammer:
I implemented FloatEquals, but it somehow doesn't work yet. I don't know whether there is a mistake somewhere or whether float comparison just doesn't work.

JonasKellerer:
I found the error with float equals:
In the metadata file we write "0.90". This is converted in `metadata_strore.cpp` line 39 through `std::stof(value)`. This produces a representation of "0.89999997615814209". However, when the query is read in `float_equals.cpp` in line 55 as "0.90000000000000002". This difference comes, because nlohmann_josn uses the a 15digit precision https://json.nlohmann.me/home/faq/#serialization-issues and stof does not. I changed it to `std::stod(value)` which should have the same precision as nlohmann. This fixes the issue for now.

PR #142, opened 2023-06-30
feat: read pangolinage alias from json instead of txt

### LINKED ISSUE #140: Remove custom format of pango aliases
The current format for pango aliases differs slightly from the aliases of cov-lineages.

https://github.com/cov-lineages/pango-designation/blob/master/pango_designation/alias_key.json

As we now have json libraries included already, we should parse the standard format

JonasKellerer:
Resolves #140


Taepper (approved):
Looks good, just add the `static_cast` from my PR to fix the actions and then delete the `pango_alias.txt` file

Issue #131, opened 2023-06-15
Download fasta files


fengelniederhammer:
Make it possible to download sequences:
* unaligned nucleotide sequences
* aligned nucleotide sequences
* aligned amino acid sequences

Implement an action for each of the sequence categories. It should support at least 100k sequences in one download.

#112 is a prerequisite where we experiment with how to download sequences. In this issue, we want to fully implement the functionality.

Issue #85, opened 2023-04-05
Include Amino Acid Sequences


fengelniederhammer:
Currently the database only stores nucleotide sequences. The database should also store amino acid sequences.

Details:
- Probably create subtasks and link them here to split this into smaller bits.

PR #144, opened 2023-07-03
feat: Generalizing the config

Taepper:
This is the proposed generalized config and accompanying test data.
Any additional remarks or change requests?

I would envision, that we do not mention the explicit filenames in any config but instead expect the files through conventional file names (`{'gene','nuc'}'_'<name>`)


chaoran-chen:
The notation/convention sounds good to me!

In principle, I'd say that it's better to avoid redundancy whenever possible but I don't know whether there are advantages to have the names in `database_config.yaml` as well.

Taepper:
IMO, the two points speaking for the inclusion of the files is:

- consistency with metadata columns -> all filterable data is mentioned in the config
- the flexibility to configure specific (sequence) columns. E.g. potentially specifying aliases, cache behaviour, etc.

Taepper:
Resolves #149
Works on #7 
and already prepares / is a prerequisite for
#85

Dockerfile
```diff
@@ -5,7 +5,8 @@ RUN apk update && apk add --no-cache py3-pip \
     cmake=3.24.4-r0 \
     linux-headers=5.19.5-r0 \
     boost-build=1.79.0-r0 \
-    libtbb=2021.7.0-r0
+    libtbb=2021.7.0-r0 \
+    zstd-dev=1.5.5-r0
```
fengelniederhammer, line 9: Is there a reason why we don't install that via Conan? It seems to be available: https://conan.io/center/zstd

Taepper: At first, I added it only to the `conanfile.py`, but CMake it wouldn't find the package, this is still somewhat experimenting

Taepper: Is there a reason why we do not do the same for e.g. https://conan.io/center/tbb ?

fengelniederhammer: Yes. TBB complains a lot when you try to link it statically, since it depends too much on the individual CPU. So we decided to give in on that end.

include/silo/common/zstdfasta_reader.h
```diff
@@ -0,0 +1,37 @@
...
+#include <zstd.h>
+
+#include "silo/common/input_stream_wrapper.h"
+
+namespace silo {
+class ZstdFastaReader {
+  private:
+   silo::InputStreamWrapper in_file;
+   ZSTD_DDict* zstd_dictionary;
+   ZSTD_DCtx* zstd_context;
```
fengelniederhammer, line 16: Is it possible to use smart pointers here?

src/silo/common/zstdfasta_reader.test.cpp
fengelniederhammer, line 1: Why is this empty? :)

src/silo/database.cpp
```diff
@@ -55,55 +54,37 @@ const PangoLineageAliasLookup& Database::getAliasKey() const {
...
-               throw PreprocessingException(
-                  "Sequences in meta data and sequence data for chunk " +
-                  buildChunkName(partition_index, chunk_index) +
-                  " are not equal. The sequence store has " +
-                  std::to_string(sequence_store_sequence_count) + " rows, the metadata store has " +
-                  std::to_string(metadata_store_sequence_count) + " rows."
-               );
+            for (auto& [nuc_name, reference_sequence] : reference_genomes.nucleotide_sequences) {
+               std::filesystem::path sequence_filename = input_folder;
+               sequence_filename += "nuc_" + nuc_name + std::filesystem::path::preferred_separator;
```
fengelniederhammer, line 79: Do we want to force users to use this abbreviation here? Those files will be generated by scripts mostly anyway, so it would also not hurt much so call the files `nucleotideSequence_main...`.

src/silo/database.cpp
```diff
@@ -298,44 +257,50 @@ void addStatisticToBitmapContainerSize(
...
-                  } else if (genome_symbol == NUCLEOTIDE_SYMBOL::GAP) {
-                     bitmap_container_size_per_genome_section.size_per_genome_symbol_and_section
-                        .at(genomeSymbolRepresentation(NUCLEOTIDE_SYMBOL::GAP))
-                        .at(position_index / section_length) += statistic.n_bitset_containers;
-                  } else {
-                     bitmap_container_size_per_genome_section.size_per_genome_symbol_and_section
-                        .at("NOT_N_NOT_GAP")
-                        .at(position_index / section_length) += statistic.n_bitset_containers;
+         for (const auto& database_partition : partitions) {
+            for (auto& [_, seq_store] : database_partition.nuc_sequences) {
```
fengelniederhammer, line 274: My IDE suggests: 
```suggestion
            for (const auto& [_, seq_store] : database_partition.nuc_sequences) {
```

src/silo/prepare_dataset.cpp
```diff
@@ -315,35 +303,34 @@ void sortSequenceFile(
...
       file_pos_to_sorted_pos[key_date_pair.file_pos] = number_of_sorted_files++;
    }
 
-   SPDLOG_TRACE("Calculated positions for every sequence {}", chunk_str);
+   // SPDLOG_TRACE("Calculated positions for every sequence {}", chunk_str);
 
    sequence_in.reset();
 
-   SPDLOG_TRACE("Reset file seek, now read second time, sorted {}", chunk_str);
+   // SPDLOG_TRACE("Reset file seek, now read second time, sorted {}", chunk_str);
```
fengelniederhammer, line 333: Comment those back in?

src/silo/prepare_dataset.cpp
```diff
@@ -295,15 +287,11 @@ std::unordered_map<std::string, silo::common::Date> sortMetadataFile(
 }
 
 void sortSequenceFile(
-   silo::FastaReader& sequence_in,
-   std::ostream& sequence_out,
+   silo::ZstdFastaReader& sequence_in,
+   silo::ZstdFastaWriter& sequence_out,
    const PartitionChunk& chunk,
```
fengelniederhammer, line 292: This seems unused. Is it really only used for logging? Then we could maybe log something else.

src/silo/query_engine/actions/nuc_mutations.cpp
```diff
@@ -1,37 +1,41 @@
...
+#include "silo/storage/sequence_store.h"
 
 using silo::query_engine::OperatorResult;
 
 namespace {
 
-std::pair<std::vector<size_t>, std::vector<size_t>> splitBitmaps(
-   const silo::Database& database,
+std::pair<std::vector<size_t>, std::vector<size_t>> preFilterBitmaps(
+   const silo::SequenceStore& sequence_stores,
```
fengelniederhammer, line 26: This should be singular
```suggestion
   const silo::SequenceStore& sequence_store,
```

src/silo/query_engine/filter_expressions/has_mutation.cpp
```diff
@@ -27,7 +27,9 @@ std::unique_ptr<operators::Operator> HasMutation::compile(
    const silo::DatabasePartition& database_partition,
    AmbiguityMode mode
 ) const {
-   const char ref_symbol = database.reference_genome->genome_segments[0].at(position);
+   const char ref_symbol = database.reference_genomes.nucleotide_sequences
+                              .at(database.database_config.default_nucleotide_sequence)
+                              .at(position);
```
fengelniederhammer, line 32: This probably also needs to be generalized at some point?

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
```diff
@@ -28,18 +27,20 @@ std::unique_ptr<silo::query_engine::operators::Operator> NucleotideSymbolEquals:
    const silo::DatabasePartition& database_partition,
    Expression::AmbiguityMode mode
 ) const {
-   if (position >= database.reference_genome->genome_segments[0].size()) {
+   const auto& seq_store_partition =
+      database_partition.nuc_sequences.at(database.database_config.default_nucleotide_sequence);
```
fengelniederhammer, line 31: Same here: This probably also needs to be generalized at some point?

src/silo/storage/reference_genomes.test.cpp
fengelniederhammer, line 1: Let's also test that amino acid sequences are read correctly, since this is already implemented?

src/silo/storage/sequence_store.test.cpp
fengelniederhammer, line 1: Now this test doesn't test anything anymore. Why did you delete the tests? Can we restore them?

testBaseData/test_database_config.yaml
```diff
@@ -1,5 +1,20 @@
...
+    - name: N
+    - name: S
+    - name: ORF1A
+    - name: ORF1B
+    - name: ORF3A
+    - name: ORF6
+    - name: ORF7A
+    - name: ORF7B
+    - name: ORF8
+    - name: ORF9B
```
fengelniederhammer, line 17: Why do we need to configure that here? Isn't this information already clear from reading the `reference-genomes.json`?

fengelniederhammer (changes requested):
Overall it looks good.

src/silo/storage/sequence_store.test.cpp
Taepper: The tests only covered the function to get Ambiguous Bitmaps from a symbol, which is unused as it is handled when translating a filter expression to an operator tree

testBaseData/test_database_config.yaml
Taepper: Right, this was from the first proposal and was supposed to be deleted here

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
Taepper: Yes, many expressions need to support multiple segment targets. This will be done when we also support amino acids

include/silo/common/zstdfasta_reader.h
Taepper: Unfortunately not, as it is a C library, but good catch to add the relevant constructors, assignment operators and destructors to both classes

src/silo/database.cpp
Taepper: I proposed it as such and Chaoran liked the notation https://github.com/GenSpectrum/LAPIS-SILO/pull/144#issue-1785831275

src/silo/query_engine/filter_expressions/has_mutation.cpp
Taepper: Same as below

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
Taepper: ... as I want to decouple changes to storage and queries where it is not absolutely necessary. Therefore the PR for AAcids will also first only cover storage and then a second PR will cover queries (including the namings here)

include/silo/common/zstdfasta_reader.h
fengelniederhammer: Maybe a dumb question, but can't we wrap those raw pointers somehow? From the comment on `ZSTD_createDCtx` I suspect that those are supposed to be used as shared pointers in a sense?

src/silo/database.cpp
fengelniederhammer: If Chaoran doesn't have a strong opinion on it, I would still vote for the longer name. Our aim is to make this usable for external maintainers, so we should avoid obfuscating things.

src/silo/storage/sequence_store.test.cpp
fengelniederhammer: Then let's delete this file?

include/silo/common/zstdfasta_reader.h
Taepper: Yes, we could wrap them in a type that calls `ZSTD_freeDCtx` method in the destructor and then wrap then we also have the possibility to put that new type in a shared pointer, where multiple accesses to the dictionaries are possible. 
I guess this will create 4 files, but it removes the ugly constructors etc. from the FastaReader and Writer and can be reused in the streaming output for bulk downloading: #131 

Will do!

Taepper: For now the wrapper class itself is in a `unique_ptr` in the ZstdFastaReader / Writer, as we do not need to share it.
But we can as you correctly read from the comment also use it in `shared_ptr` when we need to

src/silo/database.cpp
chaoran-chen: I think that `nuc` (just like e.g. `aa` or `qc`) is a very common abbreviation for the users/maintainers that we target, so I think that it is fine to use it (but it's not a super strong opinion)

Taepper: Additionally the file ending `fasta` is already the indication that it is a sequence file and we only prepend `gene_` and `nuc_` to avoid collisions between the two.

Taepper: I will add the discussion to #156
Maybe afterall we do make it configurable? An alternative would also be, to let the nucleotideSequence prefix default to the empty string, as the use-case will most often only contain one such sequence. Then the default filename could also be `aligned` to further comply with standard outputs of other software systems

fengelniederhammer (approved):

PR #146, opened 2023-07-04
fix: test cases verifying that the positions index for mutation distribution are now 1-indexed

Taepper:


Taepper:
resolves #61

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
```diff
@@ -28,6 +28,12 @@ std::unique_ptr<silo::query_engine::operators::Operator> NucleotideSymbolEquals:
    const silo::DatabasePartition& database_partition,
    Expression::AmbiguityMode mode
 ) const {
+   if (position >= database.reference_genome->genome_segments[0].size()) {
+      throw QueryParseException(
+         "NucleotideEquals position is out of bounds '" + std::to_string(position + 1) + "' > '" +
+         std::to_string(database.reference_genome->genome_segments[0].size()) + "'"
+      );
```
fengelniederhammer, line 35: Maybe add a test for that case, too?

fengelniederhammer (approved):

src/silo/query_engine/filter_expressions/nucleotide_symbol_equals.cpp
Taepper: I have some prepared that were a little cumbersome to debug, as they only threw in github actions. Wanted to do this later with #147 as we have other issues that are currently more pressing

Issue #147, opened 2023-07-04
More invalid API queries in tests


Taepper:
Currently the verification of user input is not exhaustive and cause crashes of SILO. After we meet the functionality demands of COVID-LAPIS we should extensively increase the testing of faulty inputs

fengelniederhammer:
```
[2023-07-04 17:46:53.486] [console_logger] [info] [query_handler.cpp:29] received query: {"action":{"type":"Details","fields":["division","age","pango_lineage"]},"filterExpression":{"children":[{"column":"division","value":"Basel-Stadt","type":"StringEquals"}],"type":"And"}}
[2023-07-04 17:46:53.486] [console_logger] [debug] [query_engine.cpp:33] Parsed query: (division = 'Basel-Stadt')
[2023-07-04 17:46:53.486] [console_logger] [error] [error_request_handler.cpp:19] Caught exception: _Map_base::at

```

Taepper:
Also include tests with invalid user data. (Duplicate/missing primary keys, wrong datatypes, primary keys of sequence files not matching metadata file -> missing in metadata, missing in sequence data)

Taepper:
This has improved a lot already, if we get some explicit errors, we will add more tests

PR #151, opened 2023-07-04
Support for null values in all columns

### LINKED ISSUE #148: Deal with Null values
Currently some column types cannot deal with null values. Null Values are given as `?` or empty strings in the tsv file for a field. See also https://docs.nextstrain.org/projects/ncov/en/latest/guides/data-prep/local-data.html#missing-metadata

String columns will parse them as is. User can filter them by typing in a question mark or empty string themselves.

Int columns need to be adapted `INT32_MIN` will be reserved for null values

Float columns are adapted and `NAN` will be used for null values

PangoLineages and Dates have a special datatype and parsing already where null will be added

Taepper:
All columns can now support null values. Resolves #148


Taepper:
With the exception of primary key columns of course, here we should add tests with invalid test datasets in #147

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -93,26 +96,49 @@ struct Tuple {
             data_pointer += sizeof(decltype(value));
          } else if (metadata.getColumnType() == config::ColumnType::INT) {
             const int32_t value = *reinterpret_cast<const int32_t*>(data_pointer);
-            fields[metadata.name] = value;
+            if (value == INT32_MIN) {
+               fields[metadata.name] = std::nullopt;
```
fengelniederhammer, line 100: I'm not sure whether this is a good idea. What if a value at some point really is `INT32_MIN`?

src/silo/query_engine/query_result.cpp
```diff
@@ -16,7 +16,11 @@ void to_json(nlohmann::json& json, const QueryResult& query_result) {
 // NOLINTNEXTLINE(readability-identifier-naming)
 void to_json(nlohmann::json& json, const QueryResultEntry& result_entry) {
    for (const auto& [field, value] : result_entry.fields) {
-      json[field] = value;
+      if (value.has_value()) {
+         json[field] = value.value();
+      } else {
+         json[field] = nlohmann::json();
+      }
```
fengelniederhammer, line 23: Would that work?
```suggestion
      json[field] = value.value_or(nlohmann::json());
```

include/silo/common/nulltype.h
fengelniederhammer, line 1: Do we need that?

src/silo/query_engine/actions/aggregated.cpp
fengelniederhammer: As discussed: Let's create an issue so that we can review this again at a later point:
* Either we really make int values `optional`s
* or we forbid users from inputting `INT32_MIN` in metadata and queries.

fengelniederhammer (approved):

include/silo/common/nulltype.h
Taepper: Ah, not anymore! Thansk!

src/silo/query_engine/query_result.cpp
Taepper: I experimented with this three lines quite a lot, but there was no easy way to set a field to null. Your proposed solution also makes the field an empty string (also in the float / int case) instead of null

fengelniederhammer: How is it different to the current solution then? Why does this yield "null"?

Taepper: The type of the statement is different I suppose. Could not quite figure it out myself. An explicit cast to another type or an embedded `to_json` somewhere could maybe do it, but I did not get it to work

Issue #157, opened 2023-07-05
info Endpoint – Statistics about storage efficiency / roaring container usage


Taepper:
The /info endpoint currently returns the size of the first sequence store. We might want to clearly lay out, which information the endpoint should provide

Taepper:
Info endpoint should provide:
- number of sequences
- number of partitions

PR #163, opened 2023-07-07
Reenable linter and code style standardization

Taepper:
Reenable the linter


include/silo/common/aa_symbol_map.h
```diff
@@ -0,0 +1,33 @@
...
+      // clang-format off
+      archive& data;
+      // clang-format on
+   }
+
+   std::array<T, AA_SYMBOL_COUNT> data;
+
+  public:
+   inline T& operator[](AA_SYMBOL symbol) { return data.at(static_cast<uint8_t>(symbol)); }
+   inline const T& at(AA_SYMBOL symbol) const { return data.at(static_cast<uint8_t>(symbol)); }
```
JonasKellerer, line 28: Why does this need to be inline? Couldn't we put this into a cpp file?

src/silo/database.cpp
```diff
@@ -172,9 +193,9 @@ BitmapContainerSize::BitmapContainerSize(size_t genome_length, size_t section_le
       total_bitmap_size_computed(0) {
    size_per_genome_symbol_and_section["NOT_N_NOT_GAP"] =
       std::vector<size_t>((genome_length / section_length) + 1, 0);
-   size_per_genome_symbol_and_section[genomeSymbolRepresentation(NUCLEOTIDE_SYMBOL::GAP)] =
+   size_per_genome_symbol_and_section["-"] =
```
JonasKellerer, line 196: I think we use `"-"` and `"N"` also below (line 326 ff). Could we make this a constant, so we do not change this here by chance and not below?

JonasKellerer (approved):
Wow. Many lines changed. Thanks :)

include/silo/common/aa_symbol_map.h
Taepper: Is there an upside to that? I don't think it will save an import, but of course we can put it there, although it raises the file count

src/silo/database.cpp
Taepper: This was the only usage of the helper function genomeSymbolRepresentation which put an additional header include into `nucleotide_symbols`, but we can call the function to return the char and then call an explicit string instructor. Although, this is only user-facing analysis output which is bound to change with #157 anyways. I am fine with both anyways

Issue #164, opened 2023-07-10
TUM-IDP: Add storage and search of insertions


Taepper:
As part of the TUM IDP @danielgrittner and @GeorgKreuzmayr will add their implemented algorithms and data-structures as a PR.

PR #166, opened 2023-07-10
Insertion Columns and Search

Taepper:
Preparation of insertion columns #164!

Most relevant parts of the edits should happen in the TODO marked areas of the insertion column


danielgrittner:
Hey Alex (@Taepper) !
Looking forward to your feedback :)

include/silo/storage/column/insertion_index.h
```diff
@@ -0,0 +1,74 @@
...
+      archive& insertion_positions;
+      // clang-format on
+   }
+
+   struct Insertion {
+      std::string value;
+      sequence_ids_t sequence_ids;
+   };
+
+   using one_mer_index_t = std::array<sequence_ids_t, NUC_SYMBOL_COUNT>;
```
Taepper, line 43: I was a little confusing that the typedef `sequence_ids_t` was used here. `insertion_ids_t` would be a bit more descriptive I would think

include/silo/common/nucleotide_symbols.h
```diff
@@ -125,6 +129,24 @@ inline std::optional<NUCLEOTIDE_SYMBOL> charToNucleotideSymbol(char character) {
          return std::nullopt;
    }
 }
+
+inline size_t toNucleotideSymbolId(char c) {
```
Taepper, line 133: Functions like these were once in the codebase but I would like to keep them out and use the enum classes NucleotideSymbol and AASymbol.
For the indexing reasons I think you preferred these functions, you can have a look at NucleotideSymbolMap, which also uses a constexpr sized array for storage and lookup of value types for a nucleotide symbol

include/silo/query_engine/operators/bitmap_producer.h
```diff
@@ -0,0 +1,41 @@
...
+#include "silo/query_engine/operator_result.h"
+#include "silo/query_engine/operators/operator.h"
+
+namespace roaring {
+class Roaring;
+}  // namespace roaring
+
+namespace silo::query_engine::operators {
+
+class BitmapProducer : public Operator {
```
Taepper, line 18: I thought, we can do without this function.
The plan in my head would have been to map the prefiltering and the regex match as seperate operators, which can then be reordered by the optimizer to further reduce intermediate result sizes.

But trying to refactor I noticed this was not easily possible, as the notion of prefiltering for **InsertionPositions**, not sequence ids can not be modelled by an operator tree.
Therefore I guess keeping this operator as is, is the way to go

Taepper (commented):
Looks good!
A few things: The queries also get a position at which the insertion are to be searched for.
Secondly, I would refactor the insertion_index a bit. Using `roaring::Roaring` as the `sequence_id_t` and the `NUCLEOTIDE_SYMBOLS` instead of `string`s and `size_t`s

A few suggestions in https://github.com/GenSpectrum/LAPIS-SILO/tree/insertions_proposal

Searching for empty strings is not possible as per https://github.com/GenSpectrum/LAPIS/blob/8de8c7c966ad08cdf9d5ecf5f3baaab060d142e2/server/src/main/antlr/ch/ethz/lapis/api/parser/VariantQuery.g4#L4 and should be responded to with an error

danielgrittner:
@Taepper Thanks for the review and the proposal! I incorporated your feedback, fixed some issues, and resolved the merge conflicts. Should be good to go now imo :)

src/silo/storage/column/insertion_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+void InsertionIndex::addLazily(const std::string& insertions_string, uint32_t sequence_id) {
+   if (insertions_string.empty()) {
+      return;
+   }
+   for (auto& position_and_insertion : splitBy(insertions_string, DELIMITER_INSERTIONS)) {
+      auto [position, insertion] = parseInsertion(position_and_insertion);
+
+      auto it1 =
+         collected_insertions.emplace(position, std::unordered_map<std::string, sequence_ids_t>{});
+      auto it2 = it1.first->second.emplace(insertion, sequence_ids_t{});
```
fengelniederhammer, line 184: Please improve the naming. Maybe something along the lines of:
```suggestion
      auto insertions_at_position =
         collected_insertions.emplace(position, std::unordered_map<std::string, sequence_ids_t>{}).first;
      auto sequence_ids_at_position = insertions_at_position->second.emplace(insertion, sequence_ids_t{});
```

include/silo/storage/column/insertion_index.h
```diff
@@ -0,0 +1,97 @@
...
+      [[maybe_unused]] void serialize(Archive& archive, const uint32_t /* version */) {
+         // clang-format off
+         archive & value;
+         archive & sequence_ids;
+         // clang-format on
+      }
+
+      std::string value;
+      sequence_ids_t sequence_ids;
+   };
```
fengelniederhammer, line 54: Why are those inner classes / structs? This makes the outer class quite lengthy and I don't really see a benefit of having it nested in another class.

src/silo/storage/column/insertion_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+   }
+   return {result.begin(), result.end()};
+}
+
+}  // namespace
+
+void InsertionIndex::InsertionPosition::buildThreeMerIndex() {
+   using bitset_one_mers_t = NucleotideSymbolMap<bool>;
+   using bitset_two_mers_t = NucleotideSymbolMap<bitset_one_mers_t>;
+   using bitset_three_mers_t = NucleotideSymbolMap<bitset_two_mers_t>;
```
fengelniederhammer, line 68: Isn't there a simpler way to express this type? In other languages, I'd write something like `Map<(NucleotideSymbol, NucleotideSymbol, NucleotideSymbol), bool>`. I'm pretty sure C++ supports something similar.

This construct here is tricky to understand. Future readers of the code would value something more intuitive.

fengelniederhammer: The same basically applies to the `three_mer_index_t`.

src/silo/storage/column/insertion_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+      return seed;
+   }
+};
+
+std::vector<InsertionIndex::three_mer_t> extractThreeMers(const std::string& search_pattern) {
+   std::unordered_set<InsertionIndex::three_mer_t, ThreeMerHash> result;
+   for (const auto& continuous_string : splitBy(search_pattern, REGEX_ANY)) {
+      auto continuous_symbols = stringToNucleotideSymbolVector(continuous_string);
+      if (continuous_symbols == std::nullopt) {
+         throw std::runtime_error("Wrong symbol in pattern: " + continuous_string);
```
fengelniederhammer, line 50: It would actually be nice to know which symbol was wrong.

src/silo/storage/column/insertion_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+      }
+   }
+}
+
+std::unique_ptr<InsertionIndex::sequence_ids_t> InsertionIndex::InsertionPosition::
+   searchWithThreeMerIndex(
+      const std::vector<three_mer_t>& search_three_mers,
+      const std::regex& search_pattern
+   ) const {
+   assert(!search_three_mers.empty());
```
fengelniederhammer, line 106: Please don't use `assert`. We've had those already. This aborts the program if it fails. SILO is a continuously running program (currently a REST API), you don't want to shut it down. You want to throw exceptions that you can handle, so that you can send subsequent requests.

endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json
```diff
@@ -0,0 +1,18 @@
...
+    "filterExpression": {
+      "type": "InsertionContains",
+      "column": "insertions",
+      "position": 25701,
+      "value": "CC+++"
+    }
+  },
+  "expectedError": {
+    "error": "Bad request",
+    "message": "The field 'value' in the InsertionContains expression does not contain a valid regex pattern: \"CC+++\""
```
fengelniederhammer, line 16: I would be nice to tell the user what a valid pattern would be.

src/silo/storage/column/insertion_index.cpp
Taepper: If compiled in release mode asserts are actually no-ops (i.e. function purely as sanity checks in debugging)

testBaseData/small_metadata_set.tsv
```diff
@@ -1,101 +1,101 @@
...
-EPI_ISL_3016465	B.1.1.7	2021-07-05	Europe	Switzerland	Valais	2021-01-16	""	0.95
-EPI_ISL_3247294	""	2021-07-15	Europe	Switzerland	Basel-Stadt	2020-12-25	56	0.94
-EPI_ISL_3578231	P.1	2021-05-12	Europe	Switzerland	Zürich	2021-01-12	57	0.93
-EPI_ISL_3465732	AY.43	2021-08-05	Europe	Switzerland	Vaud	2021-02-23	58	0.92
-EPI_ISL_2367431	B.1	2020-03-16	Europe	Switzerland	Vaud	2021-01-22	59	0.91
-EPI_ISL_3465556	AY.43	2021-08-04	Europe	Switzerland	Solothurn	2021-01-15	50	0.90
-EPI_ISL_2359636	B.1.1.189	2021-02-03	Europe	Switzerland	Vaud	2020-11-04	57	0.89
\ No newline at end of file
+gisaid_epi_isl	pango_lineage	date	region	country	division	unsorted_date	age	qc_value	insertions
+EPI_ISL_1408408	B.1.1.7	2021-03-18	Europe	Switzerland	Basel-Land		4	0.98	""
```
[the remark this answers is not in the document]
Taepper: Huh, I guess, I created this. I used the CLion provided csv editor, which inserted this for the empty cells I suppose. But I am gonna fix it manually

include/silo/common/nucleotide_symbols.h
```diff
@@ -125,6 +129,22 @@ inline std::optional<NUCLEOTIDE_SYMBOL> charToNucleotideSymbol(char character) {
          return std::nullopt;
    }
 }
+
+inline std::optional<std::vector<NUCLEOTIDE_SYMBOL>> stringToNucleotideSymbolVector(
```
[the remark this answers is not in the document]
Taepper: The "inline" keyword generally does not influence whether a function is inlined or not. It is just telling the compiler that this function is compiled for each compilation unit and should not get a symbol for the linker (which would conflict across compilation units)

Sometimes I try not to increase the file count for very small functions, like the others in this file. I guess this is border-line. We can transfer them to a `.cpp` but I have no strong opinion here

src/silo/common/numeric_conversion.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+uint32_t tryConvertStringToU32(const std::string& input) {
+   std::stringstream string_number_input(input);
+   uint32_t target;
+   if (!(string_number_input >> target)) {
+      const std::string message = "Failed to convert \"" + input + "\" to uint32";
+      throw NumericConversionException(message);
+   }
+   return target;
+}
```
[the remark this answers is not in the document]
danielgrittner: The standard library provides two options: either `std::stringstream` or `std::stoul`. But `std::stoul` did not throw for out-of-range numbers, therefore, I chose `std::stringstream`.

Nonetheless, I just found `boost::lexical_cast` which works fine. So I will remove the code again :)

include/silo/common/string_utils.h
```diff
@@ -0,0 +1,14 @@
+#ifndef SILO_STRING_UTILS_H
+#define SILO_STRING_UTILS_H
+
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace silo {
+
+std::vector<std::string> splitBy(const std::string& value, const std::string_view delimiter);
```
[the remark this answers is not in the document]
danielgrittner: The standard library does not provide a split function. Regarding boost, there is `boost::split` which does not allow to split strings by a delimiter consisting of multiple characters.

include/silo/storage/column/insertion_index.h
```diff
@@ -0,0 +1,97 @@
...
+struct access;
+}  // namespace boost::serialization
+
+namespace silo::storage::column::insertion {
+
+class InsertionIndex {
+   friend class boost::serialization::access;
+
+  public:
+   using three_mer_t = std::array<NUCLEOTIDE_SYMBOL, 3>;
```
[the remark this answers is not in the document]
danielgrittner: K-mer is a typical term in bioinformatics: https://en.wikipedia.org/wiki/K-mer

testBaseData/small_metadata_set.tsv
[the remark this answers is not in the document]
danielgrittner: I removed it again :)

include/silo/common/nucleotide_symbols.h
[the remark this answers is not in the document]
danielgrittner: I decided to split `nucleotide_symbols.h` into `.h` and `.cpp`

src/silo/query_engine/filter_expressions/insertion_contains.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+bool validateRegexPattern(const std::string& value) {
+   // Build the following regex pattern: ^([nuc-symbols]|\.\*)*$
+   std::stringstream regex_pattern_string;
+   regex_pattern_string << "^([";
+   for (const auto nuc : NUC_SYMBOLS) {
+      regex_pattern_string << nucleotideSymbolToChar(nuc);
+   }
+   regex_pattern_string << "]|\\.\\*)*$";
+   const std::regex regex_pattern(regex_pattern_string.str());
```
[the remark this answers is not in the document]
danielgrittner: Moved it to a static variable

include/silo/storage/column/insertion_index.h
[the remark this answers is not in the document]
danielgrittner: Adapted it

src/silo/storage/column/insertion_index.cpp
danielgrittner: Adapted it

include/silo/storage/column/insertion_index.h
```diff
@@ -0,0 +1,97 @@
...
+   using sequence_ids_t = roaring::Roaring;
```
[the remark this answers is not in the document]
danielgrittner: Removed `sequence_ids_t` again

include/silo/storage/column/insertion_index.h
danielgrittner: I moved them out now

src/silo/storage/column/insertion_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+}  // namespace
+
+void InsertionIndex::InsertionPosition::buildThreeMerIndex() {
+   using bitset_one_mers_t = NucleotideSymbolMap<bool>;
+   using bitset_two_mers_t = NucleotideSymbolMap<bitset_one_mers_t>;
+   using bitset_three_mers_t = NucleotideSymbolMap<bitset_two_mers_t>;
+
+   for (size_t insertion_id = 0; insertion_id < insertions.size(); ++insertion_id) {
+      const auto& insertion = insertions[insertion_id];
+      const auto& insertion_value = insertion.value;
```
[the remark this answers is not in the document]
danielgrittner: Adapted it

src/silo/storage/column/insertion_index.cpp
danielgrittner: C++ doesn't provide this out-of-the-box as far as I know but you can implement some helper functionality to achieve that (see `template_utils.h`). I also checked `boost::hana` but I could not find anything useful there to achieve this.

src/silo/storage/column/insertion_index.cpp
danielgrittner: Adapted it

src/silo/storage/column/insertion_index.cpp
danielgrittner: I left the assert for now because it should not affect a running instance compiled in release mode. But if you prefer to not have `assert`s In the codebase, then I will remove it.

src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -182,6 +187,16 @@ struct Tuple {
                fields[metadata.name] = string_value;
             }
             data_pointer += sizeof(decltype(value));
+         } else if (metadata.getColumnType() == config::ColumnType::INSERTION) {
+            const silo::Idx value = *reinterpret_cast<const silo::Idx*>(data_pointer);
+            std::string string_value =
```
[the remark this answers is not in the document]
danielgrittner: Will leave it non-const because `string_value` is moved a couple of lines below this line.

src/silo/query_engine/actions/details.cpp
```diff
@@ -133,6 +133,15 @@ QueryResult Details::execute(
                } else {
                   row_fields[metadata.name] = string_value;
                }
+            } else if (metadata.getColumnType() == config::ColumnType::INSERTION) {
+               const auto& column = columns.insertion_columns.at(metadata.name);
+               const silo::Idx value = column.getValues()[sequence_id];
+               std::string string_value = column.lookupValue(value);
```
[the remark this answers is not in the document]
danielgrittner: same as above

src/silo/query_engine/filter_expressions/insertion_contains.cpp
```diff
@@ -0,0 +1,110 @@
...
+         auto search_result = insertion_column.search(position, value);
+         return OperatorResult(search_result.release());
+      },
+      database_partition.sequenceCount
+   );
+}
+
+namespace {
+
+bool validateRegexPattern(const std::string& value) {
```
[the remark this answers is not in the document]
danielgrittner: This is tested with `endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json` and `endToEndTests/test/invalidQueries/insertionContains_invalidPattern2.json`

include/silo/storage/column/insertion_index.h
```diff
@@ -0,0 +1,97 @@
...
+   using insertion_ids_t = std::vector<uint32_t>;
+
+  private:
```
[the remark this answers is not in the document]
danielgrittner: Refactored it

endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json
danielgrittner: I adapted it. As a side note, since this message is user-facing (I guess?), the regex for zero or any nucleotide sequence differs here (user '?' vs internal '.*')

include/silo/storage/column/insertion_index.h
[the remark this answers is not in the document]
JonasKellerer: :see_no_evil: Didn't know that, because I only have a physics background. :) Sure. then it is fine :)

src/silo/storage/column/insertion_column.test.cpp
```diff
@@ -0,0 +1,60 @@
...
+   under_test.insert("25701:ACCA");
+   under_test.insert("25701:CCG");
+   under_test.insert("25701:CCG");
+   under_test.insert("25701:TTACAT,25701:ACCA");
+   under_test.insert("25701:ACCA");
+   under_test.insert("25701:TTACAT,25701:ACCA,25701:AGCTGTTCAG");
+
+   under_test.buildInsertionIndex();
+
+   const auto result1 = under_test.search(25701, ".*CC.*");
```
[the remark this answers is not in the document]
JonasKellerer: They decided to go with different syntax? :facepalm:

src/silo/query_engine/filter_expressions/insertion_contains.cpp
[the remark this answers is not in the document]
JonasKellerer: Ah, sry. I ment unit test this, so we know fast if it breaks. But this is fine for me.

src/silo/storage/column/insertion_index.cpp
fengelniederhammer: I don't really see the point of having it. You could as well throw an exception.

The check is done on calling side, too. So we could as well omit it either here or on the calling side, whichever makes more sense.

src/silo/storage/column/insertion_index.cpp
fengelniederhammer: Hm, unfortunate. Well then let's leave it as it is...

endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json
fengelniederhammer: Why does it differ? Where does the `?` come in? I didn't see it in the code?

src/silo/common/string_utils.test.cpp
```diff
@@ -0,0 +1,37 @@
...
+   const auto result = splitBy(input, delimiter);
+   EXPECT_EQ(result, std::vector<std::string>({"ABC", ""}));
+}
+
+TEST(splitBy, correctWithEmptyString) {
+   const std::string input;
+   const std::string_view delimiter(",");
+
+   const auto result = splitBy(input, delimiter);
+   EXPECT_EQ(result, std::vector<std::string>{""});
```
fengelniederhammer, line 36: Isn't this wrong? I would expect to see `["", ""]`.

This is what Javascript does:
```
        expect(','.split(',')).toEqual(['', '']);
        expect('a,'.split(',')).toEqual(['a', '']);
        expect(',b'.split(',')).toEqual(['', 'b']);
        expect('a,b'.split(',')).toEqual(['a', 'b']);
        expect('a,b,'.split(',')).toEqual(['a', 'b', '']);
```

src/silo/query_engine/filter_expressions/insertion_contains.cpp
```diff
@@ -0,0 +1,115 @@
...
+   const std::string& value = json["value"].get<std::string>();
+   CHECK_SILO_QUERY(
+      !value.empty(),
+      "The field 'value' in an InsertionContains expression must not be an empty string"
+   )
+   CHECK_SILO_QUERY(
+      validateInsertionSearchValue(value),
+      "The field 'value' in the InsertionContains expression does not contain a valid regex "
+      "pattern: \"" +
+         value + "\". It must only consist of nucleotide symbols and the regex symbol '?'."
```
fengelniederhammer, line 110: The regex says this:
```suggestion
         value + "\". It must only consist of nucleotide symbols and the regex symbol '.*'."
```

src/silo/common/string_utils.test.cpp
danielgrittner: `input` contains an empty string `""`, so splitting by `,` (`delimiter`) should return `[""]` (In JS: `expect("".split(",").toEqual(""));`

src/silo/query_engine/filter_expressions/insertion_contains.cpp
danielgrittner: https://github.com/GenSpectrum/LAPIS/blob/8de8c7c966ad08cdf9d5ecf5f3baaab060d142e2/server/src/main/antlr/ch/ethz/lapis/api/parser/VariantQuery.g4#L44

As far as I know the user and the query language syntax uses `?` while it then gets parsed to `.*` which is passed on to SILO. But I guess Alex (@Taepper) can better comment on this.

src/silo/common/string_utils.test.cpp
fengelniederhammer: Ah sorry, I misread. I still find C++ syntax confusing sometimes.

src/silo/query_engine/filter_expressions/insertion_contains.cpp
fengelniederhammer: SILO should be self-contained and operable without LAPIS. So SILO should return error messages that are valid for SILO.

Taepper: Yes, I would say that SILO should only know about regex. This makes it more flexible to extend or generalise to different columns

endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json
Taepper: With this PR I would say that SILO already gets regex and rejects everything else. So we do not do that transformation here but I would expect it already in LAPIS.
We can change this later, but for now it is out of scope

src/silo/query_engine/filter_expressions/insertion_contains.cpp
danielgrittner: Alright. Since SILO should only know about regex and be self-contained, I changed the error message from `?` to `.*`

src/silo/storage/column/insertion_index.cpp
danielgrittner: Alright. I removed it 👍

endToEndTests/test/invalidQueries/insertionContains_invalidPattern2.json
```diff
@@ -0,0 +1,18 @@
+{
+  "testCaseName": "Insertion Contains with invalid pattern",
```
fengelniederhammer, line 2: If we want to keep this second test, please give it a unique name.

src/silo/common/nucleotide_symbols.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+std::optional<char> findIllegalNucleotideChar(const std::string& nucleotides) {
+   for (const auto nuc_char : nucleotides) {
+      auto symbol = charToNucleotideSymbol(nuc_char);
+      if (symbol == std::nullopt) {
+         return nuc_char;
+      }
+   }
+   return std::nullopt;
+}
```
fengelniederhammer, line 106: A (short) unit test for those two functions would be nice.

fengelniederhammer (approved):

Taepper:
Great!

Issue #165, opened 2023-07-10
TUM-IDP: Add pattern search


Taepper:
As part of the TUM IDP @danielgrittner  and @GeorgKreuzmayr will add their implemented algorithms and data-structures for pattern search as a PR.

PR #167, opened 2023-07-10
Pattern Search

Taepper:
Preparation of pattern search https://github.com/GenSpectrum/LAPIS-SILO/issues/165!

Most relevant parts of the edits should happen in the TODO marked areas. Of course, the additional data structures that need to be added in the sequence store which are not yet added, will make up a large part of this PR


src/silo/storage/sequence_store.cpp
```diff
@@ -135,10 +138,61 @@ void silo::SequenceStorePartition::fillNBitmaps(const std::vector<std::string>&
...
+            std::vector<uint32_t> counts(sequence_count, 0);
+            const auto end =
+               std::min(static_cast<size_t>(idx + slice_length), reference_genome.size());
+            for (uint16_t genome_pos = idx; genome_pos < end; ++genome_pos) {
+               const auto* negated_reference_bitmap =
+                  this->getBitmap(genome_pos, this->reference_genome[genome_pos]);
+               const auto* n_symbol_bitmap = this->getBitmap(genome_pos, NUCLEOTIDE_SYMBOL::N);
+               auto pure_mutation = std::make_unique<roaring::Roaring>(roaring_bitmap_andnot(
+                  &negated_reference_bitmap->roaring, &n_symbol_bitmap->roaring
+               ));
```
GeorgKreuzmayr, line 166: @Taepper I did not find where the reference genome bitmap gets flipped. Does SILO still do it? Anyways, this part is not yet correct. Should we do a union over all bitmaps except the reference bitmap to get the ones that are actually mutated to a symbol other than N

PR #173, opened 2023-07-17
refactor: some more logging during preprocessing

fengelniederhammer:
This should make it a bit more transparent.


JonasKellerer (approved):

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apk update && apk add --no-cache py3-pip \
     boost-build=1.79.0-r0 \
     libtbb=2021.7.0-r0
 
-RUN pip install conan==2.0.2
+RUN pip install conan==2.0.8
```
fengelniederhammer, line 10: Let's keep this in sync with the version in the linter Dockerfile.

src/silo/database.cpp
```diff
@@ -448,6 +448,8 @@ void Database::preprocessing(
    const config::DatabaseConfig& database_config_
 ) {
    database_config = database_config_;
+   SPDLOG_DEBUG("preprocessing - using database config: {}", database_config);
+   SPDLOG_DEBUG("preprocessing - using preprocessing config: {}", preprocessing_config);
```
fengelniederhammer, line 452: Now we are logging this twice - once should suffice.

src/silo/database.cpp
```diff
@@ -448,6 +448,8 @@ void Database::preprocessing(
    const config::DatabaseConfig& database_config_
 ) {
    database_config = database_config_;
+   SPDLOG_DEBUG("preprocessing - using database config: {}", database_config);
```
fengelniederhammer, line 451: I think we should
* either write a custom log message not using the fmt lib so that this line doesn't pretend to log the full object
* or actually log the full object (adapting the fmt implementation to include all fields)

src/silo/config/database_config.cpp
```diff
@@ -67,4 +67,29 @@ std::optional<DatabaseMetadata> DatabaseConfig::getMetadata(const std::string& n
...
\ No newline at end of file
+}  // namespace silo::config
+
+[[maybe_unused]] auto fmt::formatter<silo::config::DatabaseConfig>::format(
+   const silo::config::DatabaseConfig& database_config,
+   fmt::format_context& ctx
+) -> decltype(ctx.out()) {
+   return format_to(
+      ctx.out(),
+      "DatabaseConfig[default_nucleotide_sequence: {}, schema: {}",
```
Taepper, line 78: closing bracket is missing

Taepper (approved):

Taepper (approved):

PR #177, opened 2023-07-20
Produce serialized state

### LINKED ISSUE #9: Make preprocessing pipeline produce serialized state
Currently, the preprocessing pipeline has an intermediate step to produce processed data (a partitioned version of the input data) that can be read by SILO. We can simplify (from a users perspective) the pipeline to produce a serialized state.

The pipeline would then take the sequence input data, transform it, and produce a serialized state. The partitioned data can remain as a technically necessary intermediate step, but a user triggering the pipeline should not need to see this data.

Taepper:
Resolves #9
Now #8 is possible so the subgoal #7 can be tackled!


src/silo/database.cpp
```diff
@@ -459,49 +533,53 @@ void Database::preprocessing(
 
    SPDLOG_INFO("preprocessing - building database");
 
-   build(preprocessing_config, partition_descriptor);
+   database.build(preprocessing_config, partition_descriptor, reference_genomes);
+
+   database.saveDatabaseState("output/serialized_state/", partition_descriptor, reference_genomes);
```
fengelniederhammer, line 538: I think I would like it better, if that function call was not in this method.
* It can be tested individually: doing preprocessing without writing it to the disc.
* Whoever call this method, can freely decide what to do with the resulting database, e.g. one can still call `saveDatabaseState`.

fengelniederhammer (changes requested):
Please add a unit test that shows that the database can be loaded from a saved state successfully. Otherwise this looks good to me.

include/silo/common/bidirectional_map.h
```diff
@@ -22,12 +22,11 @@ template <typename V>
 class BidirectionalMap {
    friend class boost::serialization::access;
 
-  private:
```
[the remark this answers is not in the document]
Taepper: Ah, this is just cosmetic. class members are private by default

src/silo/database.cpp
Taepper: Ah, here the problem is, for saving the database, one needs some datastructures which we want to discard once we finished the preprocessing to minimize coupling between preprocessing part and storage/querying part of the database

Taepper: I now removed all preprocessing structs from the database loading, which makes it more flexible anyways! (and reduced some coupling during the building of the database as well)

fengelniederhammer (approved):

PR #178, opened 2023-07-24
feat: reenable pushdown of And expressions through selections

Taepper:
The initial version (aka Master Thesis state) of SILO had push down of And expressions / Intersection operators through Selections that need to iterate over all values.
This helped to separate the part for which SILO was really optimized for: working with indexes that have already been precomputed and performing the requested arithmetic on them

The queries for which we do not have indexes are best evaluated at the end, when we already reduced the number of rows we need to check using our indexes.

With this PR selections can have child operators again and will only be evaluated on those. If no child operator is present, they still iterate over the entire Table.

### DIFF
--- a/src/silo/query_engine/filter_expressions/and.cpp
+++ b/src/silo/query_engine/filter_expressions/and.cpp
@@ -16,6 +16,7 @@
 #include "silo/query_engine/operators/full.h"
 #include "silo/query_engine/operators/intersection.h"
 #include "silo/query_engine/operators/operator.h"
+#include "silo/query_engine/operators/selection.h"
 #include "silo/query_engine/operators/union.h"
 #include "silo/query_engine/query_parse_exception.h"
 #include "silo/storage/database_partition.h"
@@ -27,6 +28,7 @@ struct Database;
 namespace silo::query_engine::filter_expressions {
 
 using OperatorVector = std::vector<std::unique_ptr<operators::Operator>>;
+using operators::Operator;
 
 And::And(std::vector<std::unique_ptr<Expression>>&& children)
     : children(std::move(children)) {}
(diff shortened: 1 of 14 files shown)


src/silo/query_engine/filter_expressions/and.cpp
```diff
@@ -71,61 +78,84 @@ std::tuple<OperatorVector, OperatorVector> And::compileChildren(
...
-      return std::make_unique<operators::Full>(database_partition.sequenceCount);
-   }
-   if (non_negated_child_operators.size() == 1 && negated_child_operators.empty()) {
-      return std::move(non_negated_child_operators[0]);
-   }
-   if (negated_child_operators.size() == 1 && non_negated_child_operators.empty()) {
-      return std::make_unique<operators::Complement>(
+      if (predicates.empty()) {
+         return std::make_unique<operators::Full>(database_partition.sequenceCount);
+      }
```
fengelniederhammer, line 128:
```suggestion
      if (predicates.empty()) {
         return std::make_unique<operators::Full>(database_partition.sequenceCount);
      }
      return std::make_unique<operators::Selection>(
         std::move(predicates), database_partition.sequenceCount
      );
```
Then we can remove the `if (index_arithmetic_operator == std::nullopt)` below, since it's only possible to reach that `if` via this branch.

fengelniederhammer: Then we can make the `index_arithmetic_operator` a non-optional, since it's always assigned.

fengelniederhammer (approved):
Please simplify the `if`s in `and.cpp`, as they are weirdly intertwined and thus hard to understand. Otherwise this looks good to me.

PR #179, opened 2023-07-24
Recombinant lineages

### LINKED ISSUE #143: Consider recombinant lineages
Currently only filters for non-recombinant lineages are allowed. We should also allow recombinant lineages.

Taepper:
Resolves #143

### DIFF
--- a/endToEndTests/test/info.test.js
+++ b/endToEndTests/test/info.test.js
@@ -7,7 +7,7 @@ describe('The /info endpoint', () => {
       .get('/info')
       .expect(200)
       .expect('Content-Type', 'application/json')
-      .expect({ nBitmapsSize: 3898, sequenceCount: 100, totalSize: 66467326 })
+      .expect({ nBitmapsSize: 3898, sequenceCount: 100, totalSize: 68915226 })
       .end(done);
   });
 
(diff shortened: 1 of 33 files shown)


fengelniederhammer (commented):
I'm not yet sure what the logic should be exactly.

E.g. XB is defined as 
```
  "XB": [
    "B.1.634",
    "B.1.631"
  ]
```
Should XB also be recognized as sublineage of B.1.634?

If I interpret this PR correctly, so far this is only about "if the alias is array-valued, then we cannot resolve this any further, since it's a recombinant". Are we missing something on that end?

Taepper:
> I'm not yet sure what the logic should be exactly.
> 
> E.g. XB is defined as
> 
> ```
>   "XB": [
>     "B.1.634",
>     "B.1.631"
>   ]
> ```
> 
> Should XB also be recognized as sublineage of B.1.634?
> 
> If I interpret this PR correctly, so far this is only about "if the alias is array-valued, then we cannot resolve this any further, since it's a recombinant". Are we missing something on that end?

I asked chaoran about this. All recombinants are to be treated as a new root

fengelniederhammer (approved):

PR #180, opened 2023-07-25
fix: compiling And: append selection_child->predicates to predicates

fengelniederhammer:
LAPIS tests are currently failing, since
```
{
  "action": {
    "type": "Aggregated"
  },
  "filterExpression": {
    "type": "And",
    "children": [
      {
        "type": "IntBetween",
        "column": "age",
        "from": 51,
        "to": 53
      }
    ]
  }
}
```
currently always returns `Full`. This should fix the issue.

Also I found it easier to follow if we have the possibility to trace compilation in the logs. It's quite cluttered, since it happens several times in parallel, but it's better than nothing. There is still room to improve.


Taepper (approved):
👍🏼

PR #182, opened 2023-07-26
More robust SILO data reading

Taepper:
ignore superfluous sequences and fill missing sequences with Ns

Various bug fixes that occurred when testing with medium sized data sets


JonasKellerer (changes requested):
Good idea to let silo take care of missing/wrong input data. Thanks.

Only two minor things

src/silo/common/date.cpp
```diff
@@ -21,12 +23,14 @@ silo::common::Date silo::common::stringToDate(const std::string& value) {
       return 0;
    }
    auto split_position = value.find('-', 0);
-   if (split_position == value.size()) {
-      throw DateFormatException("Expect dates to be delimited by '-': " + value);
+   if (split_position == std::string::npos) {
+      SPDLOG_INFO("Expect dates to be delimited by '-': " + value + "\nIgnoring date");
```
fengelniederhammer, line 27: Shouldn't those be logged as warnings?

src/silo/preprocessing/partition.cpp
```diff
@@ -229,14 +232,16 @@ Chunk::Chunk(std::vector<silo::common::UnaliasedPangoLineage>&& lineages, uint32
 void Chunk::addChunk(Chunk&& other) {
    prefix = commonPangoPrefix(prefix, other.getPrefix());
    count_of_sequences += other.count_of_sequences;
-
+   // Add all pango lineages but keep invariant of pango lineages being sorted
```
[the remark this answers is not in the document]
Taepper: I would keep it, as it clarifies the code IMO

src/silo/common/date.cpp
Taepper: I also thought about it, as for now it is not the expected case, I guess you are correct. We might change it later (when the format we officially support changes)

src/silo/preprocessing/partition.cpp
```diff
@@ -229,14 +232,16 @@ Chunk::Chunk(std::vector<silo::common::UnaliasedPangoLineage>&& lineages, uint32
...
    auto copy_of_my_lineages = std::move(pango_lineages);
+   pango_lineages.clear();
+   pango_lineages.resize(copy_of_my_lineages.size() + other.pango_lineages.size());
    std::merge(
       copy_of_my_lineages.begin(),
       copy_of_my_lineages.end(),
       other.pango_lineages.begin(),
       other.pango_lineages.end(),
-      std::back_inserter(pango_lineages)
+      pango_lineages.begin()
```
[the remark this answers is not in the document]
Taepper: hmm, as we need to copy and then clear the pango_lineages anyways, I expected this to be the more natural way

fengelniederhammer (approved):

src/silo/common/zstdfasta_writer.cpp
```diff
@@ -39,3 +50,14 @@ void silo::ZstdFastaWriter::writeRaw(const std::string& key, const std::string&
              << std::to_string(compressed_genome.size()) << '\n'
              << compressed_genome << '\n';
 }
+
+void silo::ZstdFastaWriter::writeDefault(const std::string& key) {
```
fengelniederhammer, line 54: What is this default sequence about? What should it be used for?

Taepper: When talking about what should happen with mutational queries on sequences, that have no aligned sequence, it was said that we should just handle them as if they were a full sequence of `N`s. Therefore I altered the ZstdSequenceWriter to be able to write a default value for a key that is provided in the constructor

src/silo/prepare_dataset.cpp
```diff
@@ -227,104 +227,77 @@ void silo::partitionData(
...
+          : row(row),
+            primary_key(primary_key),
+            date(date) {}
    };
-   std::vector<RowWithDate> rows;
+   std::vector<RowWithKeyDate> rows;
+   rows.reserve(chunk.size);
 
    for (auto& row : metadata_reader.reader) {
-      std::this_thread::sleep_for(std::chrono::nanoseconds(1));
```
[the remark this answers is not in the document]
Taepper: Ah, I noticed that it was not necessary when I ran it, as the iteration is slower because we read the entire row instead of just one column. But better add it back in I guess just to be safe

JonasKellerer (approved):

PR #183, opened 2023-07-29
More memory-efficient build process

Taepper:
Now by default the symbol of the reference genome is flipped. And we insert the (flipped) data into already flipped bitmaps


Taepper:
Huh, here we have the tiny problem, that roaring on my local mac and roaring on linux differ in the internal memory usage (and some container types)

endToEndTests/test/query.test.js
```diff
@@ -12,12 +12,15 @@ describe('The /query endpoint', () => {
...
-      return expect(response.body.queryResult).to.deep.equal(testCase.expectedQueryResult);
+      let response = await server.post('/query').send(testCase.query);
+      try {
+        expect(response.status).to.equal(200);
+        expect(response.header['content-type']).to.equal('application/json');
+      } catch (error) {
+        console.error('Error in response header! Error body:', response.body);
+        throw error;
+      }
+      expect(response.body.queryResult).to.deep.equal(testCase.expectedQueryResult);
```
Taepper, line 23: Also, I hope you are okay with these changes, as this helps in debugging failing tests, where before the error message was not displayed, if a successful result was expected.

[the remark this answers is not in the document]
Taepper: Ah, yes that looks better with a clearer control flow!

src/silo/storage/aa_store.cpp
```diff
@@ -13,22 +13,49 @@
...
+      bitmap.shrinkToFit();
+      const uint32_t count = flipped_bitmap_before == symbol ? sequence_count - bitmap.cardinality()
+                                                             : bitmap.cardinality();
       if (count > max_count) {
          max_symbol = symbol;
          max_count = count;
       }
    }
-   if (max_symbol.has_value()) {
+   if (max_symbol != flipped_bitmap_before) {
```
JonasKellerer, line 40: Can we find a better name for `flipped_bitmap_before` because I find this comparison strange? Maybe something like `symbol_before_flipping_bitmaps`, or is this the wrong meaning?

src/silo/storage/sequence_store.cpp
```diff
@@ -15,36 +15,52 @@
...
+      bitmap.shrinkToFit();
+      const uint32_t count = flipped_bitmap_before == symbol ? sequence_count - bitmap.cardinality()
+                                                             : bitmap.cardinality();
       if (count > max_count) {
          max_symbol = symbol;
          max_count = count;
       }
    }
-   if (max_symbol.has_value()) {
+   if (max_symbol != flipped_bitmap_before) {
```
JonasKellerer, line 42: Same comment as before :)

JonasKellerer (approved):

Issue #185, opened 2023-07-31
Separate preprocessing and starting the API


fengelniederhammer:
Implement a separate Docker image that starts SILO with `--preprocessing`.
* It loads data (metadata.tsv, fasta files) from a given path
* It preprocesses the database
* It saves the serialized state to a configured location.

When starting SILO with `--api`, then it loads a serialized database state from a configured location.

#8 is a follow up to this.

PR #188, opened 2023-08-03
Return the data version

### LINKED ISSUE #130: Return the data version
Return the data version in every response: a header called `Lapis-Data-Version` that contains the UNIX timestamp of when the data set was generated (the serialized state was written).

JonasKellerer:
Resolves #130

The data version is returned in the header as `Data-Version` and not `Lapis-Data-Version` as requested in the ticked, since it is the data version from silo. 

This should be reworked, with the verbose preprocessing.

### DIFF
--- a/endToEndTests/test/info.test.js
+++ b/endToEndTests/test/info.test.js
@@ -1,4 +1,4 @@
-const { server } = require('./common');
+const { server, headerToHaveDataVersion } = require('./common');
 const { expect } = require('chai');
 
 describe('The /info endpoint', () => {
@@ -7,6 +7,7 @@ describe('The /info endpoint', () => {
       .get('/info')
       .expect(200)
       .expect('Content-Type', 'application/json')
+      .expect(headerToHaveDataVersion)
       .expect({ nBitmapsSize: 3898, sequenceCount: 100, totalSize: 60057070 })
       .end(done);
   });
@@ -79,6 +80,7 @@ describe('The /info endpoint', () => {
           'Y': 5980620,
         });
       })
+      .expect(headerToHaveDataVersion)
       .end(done);
   });
 });
--- a/include/silo/database.h
+++ b/include/silo/database.h
@@ -61,7 +61,7 @@ class Database {
    [[nodiscard]] const PangoLineageAliasLookup& getAliasKey() const;
 
    void setDataVersion(const DataVersion& data_version);
-   DataVersion getDataVersion() const;
+   virtual DataVersion getDataVersion() const;
 
   private:
    PangoLineageAliasLookup alias_key;
--- a/include/silo/database.h
+++ b/include/silo/database.h
@@ -1,6 +1,7 @@
 #ifndef SILO_DATABASE_H
 #define SILO_DATABASE_H
 
+#include <silo/common/data_version.h>
 #include <cstddef>
 #include <cstdint>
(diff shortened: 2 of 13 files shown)


endToEndTests/test/info.test.js
```diff
@@ -1,12 +1,21 @@
 const { server } = require('./common');
 const { expect } = require('chai');
 
+function headerToHaveDataVersion(response) {
```
fengelniederhammer, line 4: Let's maybe put this into `common.js`?

src/silo/database.cpp
```diff
@@ -504,6 +524,10 @@ Database Database::loadDatabaseState(const std::string& save_directory) {
          }
       }
    );
+
+   SPDLOG_TRACE("Loading data_version from {}", save_directory + "data_version.silo");
```
fengelniederhammer, line 528: Isn't that worth an info?

src/silo_api/query_handler.cpp
```diff
@@ -29,10 +33,11 @@ void QueryHandler::post(
    SPDLOG_INFO("received query: {}", query);
 
    response.setContentType("application/json");
-
    try {
       const auto query_result = query_engine.executeQuery(query);
 
+      response.set("data-version", data_version);
```
fengelniederhammer, line 39: Is it possible to test this in the `request_handler_factory.test.cpp`?

src/silo_api/info_handler.cpp
```diff
@@ -88,6 +88,8 @@ void InfoHandler::get(
 ) {
    const auto request_parameter = getQueryParameter(request);
 
+   response.set("data-version", database.getDataVersion().toString());
```
fengelniederhammer, line 91: I think headers are more often than not in upper case.
```suggestion
   response.set("Data-Version", database.getDataVersion().toString());
```

fengelniederhammer (approved):

src/silo_api/info_handler.cpp
JonasKellerer: As discussed, the data version is parsed in lower case in the e2e tests. Thus, we keep it lower case in the beginning. Additionally, it does not seem to have a real convention on the header case style.

src/silo_api/query_handler.cpp
JonasKellerer: Done :)

include/silo/database.h
```diff
@@ -1,6 +1,7 @@
 #ifndef SILO_DATABASE_H
 #define SILO_DATABASE_H
 
+#include <silo/common/data_version.h>
```
Taepper, line 4: Reorder to usual place

Taepper:
Are we sure, that we do not want to set the data version during preprocessing? And then the build only reads the data version directly from the file.

This would enable us to use the data version during change detection. And set a new data version if changes have been detected. This would also make it more flexible to already get the data version from the preprocessed input, in case we decide that SILO should not be the authority to mint new versions

JonasKellerer:
So far I have set the point where data version is mined at the end of the build process, which is also the end of the preprocessing step; just to have a first point and that it can be used in LAPIS [#272](https://github.com/GenSpectrum/LAPIS/issues/272). When looking at the preprocessing in #185 again we could also look at this again and find out a better spot?

PR #190, opened 2023-08-04
185 separate preprocessing and starting the api

### LINKED ISSUE #185: Separate preprocessing and starting the API

JonasKellerer:
Resolves #185


src/silo_api/api.cpp
```diff
@@ -20,6 +20,23 @@
...
+   }
+   SPDLOG_DEBUG("preprocessingConfig not found in config file. Using default value: config.yaml");
+   return "preprocessingConfig.yaml";
+}
+
+std::string databaseConfigPath(const Poco::Util::AbstractConfiguration& config) {
+   if (config.hasProperty("databaseConfig")) {
+      return config.getString("databaseConfig");
+   }
+   SPDLOG_DEBUG("databaseConfig not found in config file. Using default value: databaseConfig.yaml"
```
Taepper, line 35: I have seen that we generally used snake_case instead of camelCase for these files. Maybe we should change the default? (I know that I introduced it in the wrong way in a previous PR)

src/silo/database.cpp
```diff
@@ -514,6 +512,9 @@ Database Database::preprocessing(
    Database database;
    database.database_config = database_config_;
 
+   SPDLOG_INFO("preprocessing - mining data version");
+   database.setDataVersion(DataVersion(DataVersion::mineDataVersion()));
```
fengelniederhammer, line 516: I think I actually find it a bit weird, that `DataVersion::mineDataVersion()` doesn't already return a `DataVersion`

src/silo_api/api.cpp
```diff
@@ -124,9 +130,20 @@ class SiloServer : public Poco::Util::ServerApplication {
...
+      SPDLOG_INFO("Starting SILO preprocessing");
+      const std::string preprocessing_config_path = preprocessingConfigPath(config());
+      const std::string database_config_path = databaseConfigPath(config());
+
+      auto preprocessing_config =
+         silo::preprocessing::PreprocessingConfigReader().readConfig(preprocessing_config_path);
+      auto database_config =
+         silo::config::ConfigRepository().getValidatedConfig(database_config_path);
+
+      auto database_preprocessing =
```
fengelniederhammer, line 143:
```suggestion
      auto preprocessed_database =
```
I still find that name confusing. It's a `Database`, so my brain expects that the variable ends with `database`.

src/silo_api/api.cpp
```diff
@@ -77,7 +95,7 @@ class SiloServer : public Poco::Util::ServerApplication {
          handleApi();
       }
 
-      if (config().hasProperty("processData")) {
+      if (config().hasProperty("preprocessing")) {
          handleProcessData();
```
fengelniederhammer, line 99: How about returning directly after this? Or maybe even let `handleApi` and `handleProcessData` return something, so we could write `return handleProcessData()` here?

fengelniederhammer (changes requested):
Some thoughts that might be worth discussing:
* Do we want some setup for local development? I.e. a script that generates serialized output to a localtion that's on the gitignore, such that one can simply start SILO locally without the need of preprocessing? Or bundle it into a script that does preprocessing and then starts SILO?
* Why does SILO expect a preprocessing config when starting with `--api`? Actually it only needs a path where it expects it's data. I'd propose to simply configure that path and have a reasonable default, if it's not specified. The default should be such that one should start the Docker container only with `--api` and correctly mounted data. No other config should be necessary in that case. For local development, it should be configurable.
* IMO we should adapt the Docker image such that SILO is started in `/app`.

Otherwise this looks good :+1:

src/silo/database.cpp
```diff
@@ -514,6 +512,9 @@ Database Database::preprocessing(
    Database database;
    database.database_config = database_config_;
 
+   SPDLOG_INFO("preprocessing - mining data version");
+   database.setDataVersion(DataVersion::mineDataVersion());
+
```
Taepper, line 517: Maybe print the data version here

src/silo_api/api.cpp
```diff
@@ -9,16 +11,79 @@
...
 #include "silo_api/request_handler_factory.h"
+#include "silo_api/runtime_config.h"
+
+silo::preprocessing::PreprocessingConfig preprocessingConfig(
+   const Poco::Util::AbstractConfiguration& config
+) {
+   silo::preprocessing::OptionalPreprocessingConfig default_preprocessing_config;
+   if (std::filesystem::exists("./default_preprocessing_config.yaml")) {
+      default_preprocessing_config = silo::preprocessing::PreprocessingConfigReader().readConfig(
+         "./default_preprocessing_config.yaml"
```
Taepper, line 32: Why do we both have a default preprocessing config, and then also default values in the code in case they are missing in the user provided or default preprocessing config? I do not see a case where the user would override default config values in the "default_preprocessing_config.yaml"

fengelniederhammer: We could also agree to not have default values in the code. But I would definitely give a user the opportunity to overwrite values of the default config.

Like this we have three stages:
* Default values that one can simply use (e.g. the `gene_' prefix)
* Environment specific default values (locally, you don't want to set the data path to `/data`, but in the Docker image this is convenient).
* Values that a  user likely will specify (e.g. the metadata file name).

Taepper (approved):

fengelniederhammer (approved):

PR #192, opened 2023-08-07
test: clearly bundle an example dataset

fengelniederhammer:
I'd like to reference this in the documentation, so I wanted to separate it from the other test configs that we in that directory.


docker-compose-for-tests.yml
```diff
@@ -3,11 +3,10 @@ services:
...
     volumes:
-      - ./testBaseData:/data
+      - ./testBaseData/exampleDataset:/data
     ports:
       - "8080:8081"
     command:
       - "--api"
-      - "--preprocessingConfig=./preprocessing_config_for_ci.yaml"
+      - "--preprocessingConfig=./preprocessing_config.yaml"
       - "--databaseConfig=./test_database_config.yaml"
```
Taepper, line 12: I just wanted to mention that we now have defaults for the config paths:
./preprocessing_config.yaml
and ./database_config.yaml
Which we might want to use here, but we can also showcase the configurability

Taepper (approved):

docker-compose-for-tests.yml
fengelniederhammer: I'd showcase the configurability here, that's why I wanted to make it explicit. We might redesign the Docker images a little where we make use of the defaults :+1:

PR #195, opened 2023-08-09
feat: fasta aligned action

Taepper:
This PR introduces an action to retrieve fasta sequences. (Either for one or multiple sequences)


src/silo/query_engine/actions/fasta_aligned.cpp
fengelniederhammer, line 1: Can we test this action in a unit test (now that we can have sequences of arbitrary length)?

src/silo/query_engine/actions/fasta_aligned.cpp
```diff
@@ -1,34 +1,171 @@
...
    );
+
+   tbb::
+      parallel_for(
+         tbb::blocked_range<size_t>(0, sequence_store.positions.size()),
+         [&](const auto local) {
+            for (auto position_id = local.begin(); position_id != local.end(); position_id++) {
+               const NucPosition& position = sequence_store.positions.at(position_id);
+               for (const NUCLEOTIDE_SYMBOL symbol : NUC_SYMBOLS) {
+                  if (symbol != position.symbol_whose_bitmap_is_flipped &&
```
fengelniederhammer, line 53: Why is it possible to not consider positions with flipped bitmaps?

src/silo/query_engine/actions/aa_mutations.cpp
```diff
@@ -171,24 +164,74 @@ QueryResult AAMutations::execute(
...
 
 // NOLINTNEXTLINE(readability-identifier-naming)
 void from_json(const nlohmann::json& json, std::unique_ptr<AAMutations>& action) {
    CHECK_SILO_QUERY(
-      json.contains("sequenceName") && json["sequenceName"].is_string(),
-      "AminoAcidMutations action must have the string field sequenceName"
+      !json.contains("sequenceName") ||
+         (json["sequenceName"].is_string() || json["sequenceName"].is_array()),
+      "AminoAcidMutations action can have the field sequenceName of type string or an array of "
+      "strings, but no other type"
```
fengelniederhammer, line 217:
```suggestion
      "The field sequenceName of AminoAcidMutations action must have type string or an array, if present."
```

src/silo/query_engine/actions/aa_mutations.cpp
```diff
@@ -171,24 +164,74 @@ QueryResult AAMutations::execute(
...
    )
-   const std::string aa_sequence_name = json["sequenceName"].get<std::string>();
+   std::vector<std::string> sequence_names;
+   if (json.contains("sequenceName") && json["sequenceName"].is_array()) {
+      for (const auto& child : json["sequenceName"]) {
+         CHECK_SILO_QUERY(
+            child.is_string(),
+            "AminoAcidMutations action can have the field sequenceName of type string or an array "
+            "of "
+            "strings, but no other type; while parsing array encountered the element " +
```
fengelniederhammer, line 226:
```suggestion
            "AminoAcidMutations action sequenceNames must be of type string, but found " +
```

fengelniederhammer (approved):
I'm unsure about this `symbol != position.symbol_whose_bitmap_is_flipped` condition, otherwise this looks good.

Taepper:
Not final at all, but I just wanted to have something working, while writing it out to a buffer will still take some time I suppose. Maybe introduce a answer size limit for this action until we implement the compressed and/or streaming download

src/silo/query_engine/actions/fasta_aligned.cpp
Taepper: I will maybe introduce those next week if it's okay. As we need a database for the query engine to run and don't have any such tests yet, we could maybe introduce a good framework for such tests, which I would do in a separate PR

Taepper: Note that the fastaAligned_multiple.json endToEndTest already tests this function with sequences of differing lengths

fengelniederhammer: I had the different edge and error cases in mind, e.g. the sequence name does not exist. And also testing that the resulting sequence is actually correct, since this is vital. The reconstruction of the sequence is relatively complex.

fengelniederhammer:
Ok, for some working solution, this is definitely fine, although maybe not the solution we should build upon in LAPIS.

PR #197, opened 2023-08-10
implement silo data updates at runtime

Taepper:


include/silo_api/database_mutex.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace silo_api {
+
+class FixedDatabase {
+   std::shared_lock<std::shared_mutex> lock;
+
+  public:
+   FixedDatabase(silo::Database& database, std::shared_mutex& mutex);
+
+   silo::Database& database;
```
fengelniederhammer: Could we make this a const reference?

src/silo_api/database_watcher.cpp
```diff
@@ -0,0 +1,95 @@
...
+      SPDLOG_INFO(
+         "Ignoring item added event: {}. Its name {} is not a valid data version.",
+         ev.item.path(),
+         path.filename().string()
+      );
+      return;
+   }
+   {
+      auto fixed_database = database_mutex.getDatabase();
+      if (!(fixed_database.database.getDataVersion() < new_data_version.value())) {
```
fengelniederhammer, line 87:
```suggestion
      if ((fixed_database.database.getDataVersion() >= new_data_version.value())) {
```

src/silo_api/database_watcher.cpp
```diff
@@ -0,0 +1,101 @@
...
+   return silo::Database::loadDatabaseState(max_element->first);
+}
+
+}  // namespace
+
+silo_api::DatabaseWatcher::DatabaseWatcher(const std::string& path, DatabaseMutex& database_mutex)
+    : watcher(path),
+      database_mutex(database_mutex) {
+   watcher.itemAdded += Poco::delegate(this, &DatabaseWatcher::onItemAdded);
+   auto initial_database = findInitialDatabase(path);
```
fengelniederhammer, line 60: I'm not sure whether I like this:
* Conceptually: Do we want to immediately start SILO with an empty database? We might need some way to tell consumers that it's currently loading data, but I think from an API perspective it would be nice if it would be available to answer calls as fast as possible.
* Codewise: This is quite unintuitive. This has the effect that calling a constructor blocks the program for a couple minutes until it loaded the data. Do we want to make this more explicit? Why does this have to happen in the constructor of a file watcher? We could move this to a background thread or simply call it in `api.cpp`:
```
  silo_api::DatabaseWatcher watcher(data_directory, database_mutex);
  auto initial_database = findInitialDatabase(path);
   if (initial_database.has_value()) {
      database_mutex.setDatabase(std::move(*initial_database));
   }
  ...
  server.start()
```

include/silo_api/database_watcher.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <Poco/Delegate.h>
+#include <Poco/DirectoryWatcher.h>
+#include <Poco/Path.h>
+
+#include "silo_api/database_mutex.h"
+
+namespace silo_api {
+
+class DatabaseWatcher {
```
fengelniederhammer, line 14: Strictly speaking, this doesn't watch the database, it watches a directory where serialized databases might be stored. Rename to `DatabaseDirectoryWatcher` or `NewSerializedDatabaseStateWatcher` or similar?

src/silo_api/database_watcher.cpp
Taepper: I like your comments about the refactoring.

But what is the conceptual alternative?
If the database is started with a data directory that already contains data, I would find it rather intuitive to have it start with the newest data version it finds. Otherwise we need to always tell it "Hey, load the data in XY"

Taepper: Maybe doing it even before starting the DatabaseWatcher is even better? As we then do not have a race condition (albeit an almost impossible one), that between starting the watcher and looking for the initial database, a directory is added

```
  auto initial_database = findInitialDatabase(path);
   if (initial_database.has_value()) {
      database_mutex.setDatabase(std::move(*initial_database));
   }
  silo_api::DatabaseWatcher watcher(data_directory, database_mutex);
  ...
  server.start()
```

fengelniederhammer: I'm not sure what's better from a consumers perspective:
* Starting immediately with an empty database (might be confusing, since it's empty). Only after a few minutes, the database would be loaded and SILO shows data.
* Delaying the start until the database is loaded (might be confusing, since it takes a while until SILO responds at all)
* Immediately starting the API, but returning something like "sorry, still loading", until the initial DB is loaded.

Maybe the third option is the best?

Taepper: Maybe the alternative of removing the Watcher for directory events and instead using a timed event that checks all directory entries is better? Then we do not need any separate logic

fengelniederhammer: Might be, yes. It would probably use the same logic that we have for "find initial database", doesn't it? Since this requires the logic anyway, we already have it.

Taepper: Habe es angepasst! Habe den Timer zum starten auf 0 gesetzt, schaut also sofort nach einem Database State. Als Callback wird immer das directory durchsucht. (Manche log levels sollte man noch auf TRACE runtersetzen vermute ich). Sehr praktisch ist auch, dass ein Timer, solange ein Callback läuft, diesen nicht wieder aufruft. (Also wenn das loadDatabase länger als eine Intervallzeit läuft, ist direkt abgefangen )

src/silo_api/database_directory_watcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+      return std::nullopt;
+   }
+   if (data_version_in_file != data_version) {
+      SPDLOG_INFO(
+         "Skipping {}. The data version in data_version.silo is not equal to the directory name",
+         path.string()
+      );
+      return std::nullopt;
+   }
+   return data_version.value();
```
fengelniederhammer, line 62: I think this yields the same but saves unnecessary conversions:
```suggestion
   return data_version;
```

src/silo/database.cpp
```diff
@@ -368,46 +369,97 @@ std::map<std::string, std::vector<AA_SYMBOL>> Database::getAASequences() const {
...
+         "Data version is empty. Please set a data version before saving the database."
+      );
+   }
+
+   const std::filesystem::path versioned_save_directory =
+      save_directory / getDataVersion().toString();
+
+   if (std::filesystem::exists(versioned_save_directory)) {
+      auto error = fmt::format(
+         "In the output directory {} there already exists a file/folder with the name equals to "
```
fengelniederhammer, line 414:
```suggestion
         "In the output directory {} there already exists a file/folder with the name equal to "
```

src/silo/database.cpp
```diff
@@ -422,68 +474,80 @@ void Database::saveDatabaseState(const std::string& save_directory) {
...
+   data_version_file >> data_version_string;
+   auto data_version = DataVersion::fromString(data_version_string);
+   if (data_version == std::nullopt) {
+      auto error = fmt::format(
+         "Data version file {} did not contain a valid data version: {}",
+         filename.string(),
+         data_version_string
+      );
+      SPDLOG_ERROR(error);
+      throw persistence::LoadDatabaseException(error);
```
fengelniederhammer, line 492: Why do we log and throw the same message? I think we could leave the log part to the place where we catch the exception.

src/silo/database.cpp
```diff
@@ -697,4 +759,10 @@ DataVersion Database::getDataVersion() const {
    return data_version_;
 }
 
+query_engine::QueryResult Database::executeQuery(const std::string& query) const {
+   const silo::query_engine::QueryEngine query_engine(*this);
+
+   return query_engine.executeQuery(query);
+}
```
fengelniederhammer, line 766: This kinda inverts the order of control. I find it more intuitive to have a query engine that holds a database (or the mutex in this case). That query engine lives somewhere as a singleton service that executes queries.

Why did you change it to this version? Maybe I can be convinced with some good arguments ;)

src/silo_api/api.cpp
```diff
@@ -1,11 +1,15 @@
...
 #include <optional>
+#include <shared_mutex>
 #include <string>
 
+#include <Poco/Delegate.h>
+#include <Poco/DirectoryWatcher.h>
 #include <Poco/Net/HTTPServer.h>
 #include <Poco/Net/HTTPServerParams.h>
 #include <Poco/Net/ServerSocket.h>
+#include <Poco/Path.h>
```
fengelniederhammer, line 12: Please check, I think we don't need some of those anymore.

include/silo_api/database_mutex.h
```diff
@@ -0,0 +1,35 @@
...
+#define SILO_DATABASE_MUTEX_H
+
+#include <mutex>
+#include <shared_mutex>
+
+#include "silo/database.h"
+
+namespace silo_api {
+
+class FixedDatabase {
```
fengelniederhammer, line 11: I find the name a bit confusing. How about this?
```suggestion
class DatabaseLock {
```

Or `LockedDatabase`?

src/silo_api/info_handler.cpp
```diff
@@ -9,6 +9,7 @@
 #include "silo/common/nucleotide_symbols.h"
 #include "silo/database.h"
```
fengelniederhammer, line 10: Can we remove this include now?

src/silo_api/info_handler.cpp
```diff
@@ -88,27 +89,18 @@ void InfoHandler::get(
 ) {
    const auto request_parameter = getQueryParameter(request);
 
-   response.set("data-version", database.getDataVersion().toString());
+   const auto fixed_database = database.getDatabase();
 
+   response.set("data-version", fixed_database.database.getDataVersion().toString());
+
+   nlohmann::json database_info;
```
fengelniederhammer, line 96:
```suggestion
   return_detailed_info = request_parameter.find("details") != request_parameter.end() && request_parameter.at("details") == "true";
   nlohmann::json database_info = return_detailed_info
      ? nlohmann::json(database.getDatabase().database.detailedDatabaseInfo())
      : nlohmann::json(database.getDatabase().database.getDatabaseInfo());
```

src/silo_api/request_handler_factory.test.cpp
```diff
@@ -6,14 +6,14 @@
...
 #include "silo/query_engine/query_result.h"
 #include "silo/storage/column/date_column.h"
 #include "silo/storage/column/indexed_string_column.h"
 #include "silo/storage/column/int_column.h"
 #include "silo/storage/column/pango_lineage_column.h"
 #include "silo/storage/column/string_column.h"
 #include "silo/storage/database_partition.h"
+#include "silo_api/database_mutex.h"
 #include "silo_api/manual_poco_mocks.test.h"
 #include "silo_api/request_handler_factory.h"
```
fengelniederhammer, line 18: My IDE claims that some of those includes are unused.

src/silo_api/request_handler_factory.test.cpp
```diff
@@ -22,28 +22,32 @@ class MockDatabase : public silo::Database {
...
   public:
-   explicit MockQueryEngine(const silo::Database& database)
-       : QueryEngine(database) {}
+   std::shared_mutex mutex;
+   MockDatabase mock_database;
 
-   MOCK_METHOD(silo::query_engine::QueryResult, executeQuery, (const std::string&), (const));
+   silo_api::FixedDatabase getDatabase() override {
+      std::shared_lock<std::shared_mutex> lock(mutex);
+      return silo_api::FixedDatabase(mock_database, std::move(lock));
```
fengelniederhammer, line 36: "Clang-Tidy: Avoid repeating the return type from the declaration; use a braced initializer list instead"

include/silo_api/database_mutex.h
```diff
@@ -0,0 +1,35 @@
...
+};
+
+class DatabaseMutex {
+   std::shared_mutex mutex;
+   silo::Database database;
+
+  public:
+   DatabaseMutex() = default;
+
+   explicit DatabaseMutex(silo::Database&& database);
```
fengelniederhammer, line 27: Is this constructor actually used somewhere?

fengelniederhammer (approved):
Some details, but overall this looks good :+1:

src/silo/database.cpp
fengelniederhammer: As discussed: Let's leave as is, since the mutex only lives in silo_api and silo itself does not need to know about the mutex.

src/silo_api/database_directory_watcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+      "Selected database with highest data-version {} in path {} for ingestion",
+      max_element->second.toString(),
+      max_element->first.string()
+   );
+   return *max_element;
+}
+
+}  // namespace
+
+void silo_api::DatabaseDirectoryWatcher::checkDirectoryForData(Poco::Timer& /*timer*/) {
```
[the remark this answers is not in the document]
Taepper: The timer is only a reference, we need to store it somewhere. But the restart statements are superfluous, I will remove them

PR #201, opened 2023-08-23
refactor: rename "position" in aa_mutations and nuc_mutations to "mutation"

JonasKellerer:
- to be compatible with lapisv1, which also returns mutation,count,proportion
- name represents the value better, since it has also symbols inside. Positions are only numbers.

This change is necessary for: https://github.com/GenSpectrum/LAPIS/issues/206


Taepper (changes requested):
Should have been with base aa-insertions. Anyways, now wait for #200

Taepper (approved):

Issue #139, opened 2023-06-29
Make namespaces consistently reflect directory structure


Taepper:
Currently many namespaces of headers are still not consistent with the directories they are stored in. This should be handled in a separate PR as it produces a lot of noise

PR #215, opened 2023-10-02
Redesigned preprocessing

Taepper:
Still very much WIP, but this is a redesign of the entire preprocessing to 
(i) get rid of the CSVReader dependency
(ii) more easily allow generic input files
(iii) prepare internal storage of sequences for bulk download and then down the road: full SQL support via duckdb


src/silo/preprocessing/sequence_info.cpp
```diff
@@ -0,0 +1,178 @@
...
+      if (std::find(nuc_sequence_names_to_validate.begin(), nuc_sequence_names_to_validate.end(), name)
+          == nuc_sequence_names_to_validate.end()) {
+         throw silo::preprocessing::PreprocessingException(fmt::format(
+            "The aligned nucleotide sequence {} which is contained in the reference sequences is "
+            "not contained in the input file {}.",
+            name,
+            input_filename
+         ));
+      }
+   }
```
fengelniederhammer, line 154: In the Pathoplexus backend we allow that a sequence entry only has some segments. We don't require that it has all segments. From what I understood it seems to be a normal case that the sequenced result only contains some segments.

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+      auto return_code = preprocessing_db.query(fmt::format(
+         "create\n"
+         "or replace table partition_keys as\n"
+         "select row_number() over () - 1 as id, partition_key, count\n"
+         "from (SELECT {} as partition_key, COUNT(*) as count "
+         "      FROM metadata_table "
+         "      GROUP BY partition_key "
+         "      ORDER BY partition_key);",
+         database_config.schema.partition_by.value()  // TODO make sure partition key is validated
```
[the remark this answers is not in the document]
Taepper: It is actually already done in the reading of the database config

src/silo/preprocessing/sequence_info.test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+TEST(SequenceInfo, validatesSuccessfulOnCorrectFile) {
+   ReferenceGenomes reference_genomes =
+      ReferenceGenomes::readFromFile("testBaseData/exampleDataset2/reference_genomes.json");
+   SequenceInfo sequence_info(reference_genomes);
+
+   duckdb::DuckDB duckdb;
+   duckdb::Connection connection(duckdb);
+   ASSERT_NO_THROW(
+      sequence_info.validate(connection, "testBaseData/exampleDataset2/sample.ndjson.zst")
```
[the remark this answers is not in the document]
Taepper: Yes, we certainly can. I also have additional testing ndjson files with only 1-3 uncompressed genomes. Otherwise, the file is too big for my taste. We already talked about the better testing framework, which we can now maybe tackle in December. I added an issue #226

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,205 @@
...
+   for (const std::string& metadata_field : extractStringListValue(*result, 0, 0)) {
+      metadata_fields_to_validate[metadata_field] = "metadata." + metadata_field;
+   }
+   detectInsertionLists(ndjson_file, metadata_fields_to_validate);
+
+   const std::unordered_map<std::string, std::string> validated_metadata_fields =
+      validateFieldsAgainstConfig(metadata_fields_to_validate, database_config);
+
+   MetadataInfo metadata_info;
+   metadata_info.database_config = database_config;
```
fengelniederhammer, line 179: To me it seems like this field is only written but never read.

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,205 @@
...
+      if (top_level_entry == "nucleotideInsertions" || top_level_entry == "aminoAcidInsertions") {
+         auto contained_insertions = connection.Query(
+            fmt::format("SELECT {}.* FROM '{}' LIMIT 0", top_level_entry, ndjson_file.string())
+         );
+         if (contained_insertions->ColumnCount() == 0) {
+            metadata_fields_to_validate[top_level_entry] = "''";
+         }
+         if (contained_insertions->ColumnCount() == 1) {
+            metadata_fields_to_validate[top_level_entry] = fmt::format(
+               "list_string_agg({}.{})", top_level_entry, contained_insertions->ColumnName(0)
```
fengelniederhammer, line 86: Just for my understanding: Is this the same as
```
               "list_string_agg({0}.{1})", top_level_entry, contained_insertions->ColumnName(0)
```
?

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,205 @@
...
+   const std::filesystem::path& ndjson_file,
+   std::unordered_map<std::string, std::string>& metadata_fields_to_validate
+) {
+   duckdb::DuckDB db(nullptr);
+   duckdb::Connection connection(db);
+   auto top_level_entries =
+      connection.Query(fmt::format("SELECT * FROM '{}' LIMIT 0", ndjson_file.string()));
+   for (size_t idx = 0; idx < top_level_entries->ColumnCount(); idx++) {
+      const std::string& top_level_entry = top_level_entries->ColumnName(idx);
+      if (top_level_entry == "nucleotideInsertions" || top_level_entry == "aminoAcidInsertions") {
```
fengelniederhammer, line 77: This seems pretty complicated. Isn't there an easier way to check whether that JSON object has the keys `nucleotideInsertions` and `aminoAcidInsertions`?

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,205 @@
...
+   }
+   for (const std::string& name : config_metadata_fields) {
+      if (!validated_metadata_fields.contains(name)) {
+         throw silo::preprocessing::PreprocessingException(fmt::format(
+            "The metadata field {} which is contained in the database config is "
+            "not contained in the input.",
+            name
+         ));
+      }
+   }
```
fengelniederhammer, line 57: Is this the desired behavior? I think in the backend we allow that metadata fields are not present at all if they are not `required = true`. Either we should allow it here or make the backend more strict (either requiring that the proprocessing pipeline adds it or letting the backend add it).

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,205 @@
...
+      child_values.begin(),
+      child_values.end(),
+      std::back_inserter(return_value),
+      [](const duckdb::Value& value) { return value.GetValue<std::string>(); }
+   );
+   return return_value;
+}
+
+std::unordered_map<std::string, std::string> validateFieldsAgainstConfig(
+   const std::unordered_map<std::string, std::string> found_metadata_fields,
```
fengelniederhammer, line 31: The IDE suggests
```suggestion
   const std::unordered_map<std::string, std::string>& found_metadata_fields,
```

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+   if (ndjson_input_filename.has_value()) {
+      SPDLOG_DEBUG("preprocessing - ndjson pipeline chosen");
+
+      std::string file_name = ndjson_input_filename.value();
+
+      SequenceInfo sequence_info(reference_genomes);
+      sequence_info.validate(preprocessing_db.getConnection(), file_name);
+
+      MetadataInfo metadata_info = MetadataInfo::validateFromNdjsonFile(file_name, database_config);
```
fengelniederhammer, line 342:
```suggestion
      const auto metadata_info = MetadataInfo::validateFromNdjsonFile(file_name, database_config);
```

src/silo/preprocessing/preprocessing_database.cpp
```diff
@@ -0,0 +1,204 @@
...
+      duckdb::ExpressionState& /*state*/,
+      duckdb::Vector& result
+   ) {
+      using namespace duckdb;
+      BinaryExecutor::Execute<string_t, string_t, string_t>(
+         args.data[0],
+         args.data[1],
+         result,
+         args.size(),
+         [&](const duckdb::string_t uncompressed, const duckdb::string_t genome_name) {
```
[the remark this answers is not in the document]
Taepper: Ah yes, that is better. We didn't yet in SILO ^^

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+namespace silo::preprocessing {
+
+MetadataInfo MetadataInfo::validateFromMetadataFile(
+   const std::filesystem::path& metadata_file,
+   const silo::config::DatabaseConfig& database_config
+) {
+   if (database_config.schema.metadata.empty()) {
+      throw PreprocessingException("Database config without fields not possible");
+   }
```
fengelniederhammer, line 113: We can also move this to a more central and earlier location, so that SILO throws an exception already when building the database config.

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+   SPDLOG_INFO("preprocessing - reading reference genome");
+   ReferenceGenomes reference_genomes =
+      ReferenceGenomes::readFromFile(preprocessing_config.getReferenceGenomeFilename());
+
+   std::string order_by_clause = database_config.schema.getStrictOrderByClause();
+   SPDLOG_INFO("preprocessing - order by clause is {}", order_by_clause);
+
+   buildTablesFromInput(reference_genomes);
+
+   const std::string metadata_filename = preprocessing_config.getMetadataInputFilename().string();
```
fengelniederhammer, line 44: Is this used somewhere?

conanfile.py
```diff
@@ -7,13 +7,14 @@ class SiloRecipe(ConanFile):
 
     requires = [
         "boost/1.82.0",
+        "duckdb/0.8.1",
         "poco/1.12.4",
+        "hwloc/2.9.3",
```
JonasKellerer, line 12: On my machine, I get an version conflict when running with ./build_with_conan.py
`ERROR: Version conflict: onetbb/2021.9.0->hwloc/2.9.2, ->hwloc/2.9.3.`
When I use 2.9.2 then it works. Do you have the same problem?

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+   const config::DatabaseConfig& database_config
+)
+    : preprocessing_config(preprocessing_config),
+      database_config(database_config),
+      preprocessing_db(preprocessing_config.getPreprocessingDatabaseLocation().value_or(":memory:")
+      ) {}
+
+Database Preprocessor::preprocess() {
+   SPDLOG_INFO("preprocessing - reading reference genome");
+   ReferenceGenomes reference_genomes =
```
fengelniederhammer, line 36: In general there are a lot of places where we could use `const auto`

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+            preprocessing_config.getGeneFilenameNoExtension(seq_name).replace_extension(
+               silo::preprocessing::FASTA_EXTENSION
+            )
+         );
+         silo::FastaReader fasta_reader(preprocessing_config.getGeneFilenameNoExtension(seq_name)
+                                           .replace_extension(silo::preprocessing::FASTA_EXTENSION)
+         );
+
+         auto return_code = preprocessing_db.query(fmt::format(
+
```
fengelniederhammer, line 169: delete empty line?

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+            R"-(
+create or replace view gene_{0} as
+select key, sequence,
+partitioned_metadata.partition_id as partition_id
+{2}
+from raw_gene_{0} raw, partitioned_metadata
+where raw.key = partitioned_metadata.{1};
+)-",
```
fengelniederhammer, line 177: I find it easier to read like this :+1:

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+      if (return_code->HasError()) {
+         throw PreprocessingException(
+            "Error in the execution of the duckdb statement for partition key table "
+            "generation: " +
+            return_code->GetError()
+         );
+      } else {
+         SPDLOG_TRACE("Executed statement for partition key table generation.");
+         SPDLOG_TRACE(return_code->ToString());
+      }
```
fengelniederhammer, line 218:
```suggestion
      }
      SPDLOG_TRACE("Executed statement for partition key table generation.");
      SPDLOG_TRACE(return_code->ToString());
      
```

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+      return_code = preprocessing_db.query(
+         R"-(
+create or replace table partitioning as
+with recursive
```
fengelniederhammer, line 223: What does `with recursive` do?

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+                           else l2.count end
+               from grouped_partition_keys l1,
+                    partition_keys l2,
+                    allowed_count
+where l1.to_id + 1 = l2.id)
+select row_number() over () - 1 as partition_id, from_id, to_id, count
+from (select from_id, max(to_id) as to_id, max(count) as count
+      from grouped_partition_keys
+      group by from_id)
+)-"
```
fengelniederhammer, line 244: Could you maybe explain what this statement does? (And I think some reformatting wouldn't hurt here)

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+   }
+
+   preprocessing::Partitions partition_descriptor = preprocessing_db.getPartitionDescriptor();
+
+   SPDLOG_INFO("preprocessing - building database");
+
+   return buildDatabase(partition_descriptor, reference_genomes, order_by_clause, alias_key);
+}
+
+void Preprocessor::buildPartitioningTable() {
```
fengelniederhammer, line 192: How about splitting this into two functions?
```
void Preprocessor::buildPartitioningTable() {
   if (database_config.schema.partition_by.has_value()) {
      return buildPartitioningTableByColumn(database_config.schema.partition_by.value())
   }
   
   buildEmptyPartitioning()
}
```
(naming can probably be improved)

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+      }
+
+      return_code = preprocessing_db.query(
+         R"-(
+create
+or replace table partition_key_to_partition as
+select partition_keys.partition_key as partition_key,
+  partitioning.partition_id as partition_id
+from partition_keys,
+     partitioning
```
fengelniederhammer, line 258: Is this syntax actually some sort of join?

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+            "create or replace view nuc_{0} as\n"
+            "select {1} as key, nuc_{0} as sequence,"
+            "{3}"
+            "{2} \n"
+            "from preprocessing_table, partition_key_to_partition "
+            "{4};",
+            seq_name,
+            database_config.schema.primary_key,
+            order_by_select,
+            partition_by_select,
```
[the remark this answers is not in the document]
fengelniederhammer: Same for the statement below

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+      preprocessing_db(preprocessing_config.getPreprocessingDatabaseLocation().value_or(":memory:")
+      ) {}
+
+Database Preprocessor::preprocess() {
+   SPDLOG_INFO("preprocessing - reading reference genome");
+   ReferenceGenomes reference_genomes =
+      ReferenceGenomes::readFromFile(preprocessing_config.getReferenceGenomeFilename());
+
+   std::string order_by_clause = database_config.schema.getStrictOrderByClause();
+   SPDLOG_INFO("preprocessing - order by clause is {}", order_by_clause);
```
fengelniederhammer, line 40: Let's push that down to where it's actually used?

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+create or replace view nuc_{0} as
+select key, sequence,
+partitioned_metadata.partition_id as partition_id
+{2}
+from raw_nuc_{0} raw, partitioned_metadata
+where raw.key = partitioned_metadata.{1};)-",
+            seq_name,
+            database_config.schema.primary_key,
+            order_by_select
+         ));
```
fengelniederhammer, line 148:
```suggestion
        string raw_table_name = "raw_nuc_" + seq_name;
         preprocessing_db.generateNucSequenceTable(
            raw_table_name,
            reference_sequence,
            preprocessing_config.getNucFilenameNoExtension(seq_name).replace_extension(
               silo::preprocessing::FASTA_EXTENSION
            )
         );

         auto return_code = preprocessing_db.query(fmt::format(
            R"-(
create or replace view nuc_{0} as
select key, sequence,
partitioned_metadata.partition_id as partition_id
{2}
from {raw_table_name} raw, partitioned_metadata
where raw.key = partitioned_metadata.{1};)-",
            seq_name,
            database_config.schema.primary_key,
            order_by_select
         ));
```
We could extract `"raw_nuc_" + seq_name` to make the connection more obvious

src/silo/zstdfasta/zstdfasta_table.cpp
```diff
@@ -0,0 +1,82 @@
...
+   }
+   return_value = connection.Query(fmt::format(
+      "CREATE TABLE {} ("
+      "    key STRING,"
+      "    sequence BLOB"
+      ");",
+      table_name
+   ));
+   if (return_value->HasError()) {
+      throw silo::preprocessing::PreprocessingException(return_value->ToString());
```
fengelniederhammer, line 25:
```suggestion
      throw silo::preprocessing::PreprocessingException(return_value->GetError());
   }
   return_value = connection.Query(fmt::format(
      "CREATE TABLE {} ("
      "    key STRING,"
      "    sequence BLOB"
      ");",
      table_name
   ));
   if (return_value->HasError()) {
      throw silo::preprocessing::PreprocessingException(return_value->GetError());
```
since we've been using that everywhere else, too. Or does `toString` also work?

I wonder wether it might make sense to extract a macro here? something like: `CHECK_QUERY(return_value, "error when executing this specific statement")` which results in something like
```
   if (return_value->HasError()) {
      throw silo::preprocessing::PreprocessingException("error when executing this specific statement" + ": " + return_value->GetError());
   }
```

src/silo/zstdfasta/zstdfasta_table.cpp
```diff
@@ -0,0 +1,82 @@
...
+   compressed_genome.resize(compressor.getSizeBound());
+   duckdb::Appender appender(connection, table_name);
+   while (true) {
+      key = file_reader.next(uncompressed_genome);
+      if (key == std::nullopt) {
+         break;
+      }
+      size_t compressed_size = compressor.compress(uncompressed_genome, compressed_genome);
+      auto compressed_data = reinterpret_cast<const unsigned char*>(compressed_genome.data());
+      duckdb::string_t key_value = key.value();
```
fengelniederhammer, line 75: "Clang-Tidy: 'auto compressed_data' can be declared as 'const auto *compressed_data' "
```suggestion
      const size_t compressed_size = compressor.compress(uncompressed_genome, compressed_genome);
      const auto* compressed_data = reinterpret_cast<const unsigned char*>(compressed_genome.data());
      const duckdb::string_t key_value = key.value();
```

src/silo/zstdfasta/zstdfasta_table.cpp
```diff
@@ -0,0 +1,82 @@
...
+      ");",
+      table_name
+   ));
+   if (return_value->HasError()) {
+      throw silo::preprocessing::PreprocessingException(return_value->ToString());
+   }
+}
+}  // namespace
+
+void silo::ZstdFastaTable::generate(
```
fengelniederhammer, line 30: Do we need this function? My IDE only find usages in tests.

src/silo/zstdfasta/zstdfasta_table.cpp
```diff
@@ -0,0 +1,82 @@
...
+      appender.Append(key_value);
+      appender.Append(duckdb::Value::BLOB(compressed_data, compressed_size));
+      appender.EndRow();
+   }
+   appender.Close();
+}
+
+void silo::ZstdFastaTable::generate(
+   duckdb::Connection& connection,
+   std::string table_name,
```
fengelniederhammer, line 57: "Clang-Tidy: The parameter 'table_name' is copied for each invocation but only used as a const reference; consider making it a const reference "

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+         if (return_code->HasError()) {
+            SPDLOG_ERROR(return_code->GetError());
+            throw PreprocessingException(return_code->GetError());
+         }
+      }
+   } else {
+      std::string order_by_select = ", partitioned_metadata." + database_config.schema.primary_key +
+                                    " as " + database_config.schema.primary_key;
```
fengelniederhammer, line 120: Do we need this? If I see it correctly, we could simply include `raw.key as {1}` in the statement below. `where raw.key = partitioned_metadata.{1}` should make sure that it's correct. Or is there a difference that I don't see?

In any other case, we could also include `partitioned_metadata.{1} as {1}` in the statement. I think it would still be more readable than constructing part of the select statement here.

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+         preprocessing_db.generateNucSequenceTable(
+            "raw_gene_" + seq_name,
+            reference_sequence,
+            preprocessing_config.getGeneFilenameNoExtension(seq_name).replace_extension(
+               silo::preprocessing::FASTA_EXTENSION
+            )
+         );
+         silo::FastaReader fasta_reader(preprocessing_config.getGeneFilenameNoExtension(seq_name)
+                                           .replace_extension(silo::preprocessing::FASTA_EXTENSION)
+         );
```
fengelniederhammer, line 166: This seems unused.

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+         auto return_code = preprocessing_db.query(fmt::format(
```
fengelniederhammer, line 168: I think this is missing a check for errors.

include/silo/common/date.h
```diff
@@ -8,6 +8,8 @@ namespace silo::common {
 
 typedef uint32_t Date;
 
+const Date NULL_DATE = 0;
```
fengelniederhammer, line 11: Let's also use this in `silo::common::stringToDate`? There are a couple of early returns where this might be appropriate.

include/silo/storage/column/date_column.h
```diff
@@ -34,6 +34,8 @@ class DateColumnPartition {
 
    void insert(const silo::common::Date& value);
 
+   void insertNull();
```
fengelniederhammer, line 37: This is unused. Is this expected? Or is `ColumnPartitionGroup::fill` wrong?

fengelniederhammer: Same for `InsertionColumnPartition<SymbolType>::insertNull()`, `IntColumnPartition::insertNull()` and `FloatColumnPartition::insertNull()`.

include/silo/database_info.h
```diff
@@ -14,6 +15,24 @@ struct DatabaseInfo {
...
+      -> decltype(ctx.out()) {
+      return format_to(
+         ctx.out(),
+         "sequence count: {}, total size: {}, N bitmaps size: {}",
+         database_info.sequence_count,
+         silo::formatNumber(database_info.total_size),
+         silo::formatNumber(database_info.n_bitmaps_size)
+      );
+   }
+};
```
fengelniederhammer, line 32: This is an implementation. Shouldn't this be in a .cpp file?

include/silo/preprocessing/metadata.h
```diff
@@ -5,29 +5,25 @@
 #include <string>
 #include <vector>
 
-#include <csv.hpp>
+#include <duckdb.hpp>
 
 namespace silo::preprocessing {
 
 class MetadataReader {
```
fengelniederhammer, line 12: This looks suspicous. Do we really still need this class?

include/silo/preprocessing/partition.h
```diff
@@ -21,71 +21,68 @@ class DatabaseConfig;
...
+   LineageGroup() = default;
 
   public:
-   Chunk(silo::common::UnaliasedPangoLineage lineage, uint32_t count);
-   Chunk(std::vector<silo::common::UnaliasedPangoLineage>&& lineages, uint32_t count);
+   LineageGroup(silo::common::UnaliasedPangoLineage lineage, uint32_t count);
+   LineageGroup(std::vector<silo::common::UnaliasedPangoLineage>&& lineages, uint32_t count);
 
-   void addChunk(Chunk&& other);
+   void addLineageGroup(LineageGroup&& other);
```
fengelniederhammer, line 37: This seems unused.

src/silo/preprocessing/partition.cpp
```diff
@@ -132,112 +76,35 @@ silo::preprocessing::Partitions::Partitions(std::vector<Partition> partitions_)
     : partitions(std::move(partitions_)) {
    for (uint32_t part_id = 0, limit = partitions.size(); part_id < limit; ++part_id) {
       const auto& part = partitions[part_id];
```
fengelniederhammer, line 78:
```suggestion
   for (const auto & partition : partitions) {
```

src/silo/preprocessing/partition.cpp
```diff
@@ -132,112 +76,35 @@ silo::preprocessing::Partitions::Partitions(std::vector<Partition> partitions_)
...
-         auto current = pango_lineage_counts.pango_lineage_counts.begin();
-         Chunk chunk{current->pango_lineage, current->count_of_sequences};
-         current++;
-         for (; current != pango_lineage_counts.pango_lineage_counts.end(); ++current) {
-            chunk.addChunk({current->pango_lineage, current->count_of_sequences});
-         }
-         partitions.emplace_back(std::vector<Chunk>{{chunk}});
+      for (uint32_t chunk_id = 0, limit2 = part.getPartitionChunks().size(); chunk_id < limit2;
+           ++chunk_id) {
+         all_partition_chunks.emplace_back(part.getPartitionChunks()[chunk_id]);
```
fengelniederhammer, line 81:
```suggestion
      for (auto chunk_id : part.getPartitionChunks()) {
         all_partition_chunks.emplace_back(chunk_id);
```

include/silo/preprocessing/partition.h
fengelniederhammer: Is `LineageGroup` used at all? I just deleted it and the code compiles without it.

include/silo/preprocessing/partition.h
```diff
@@ -21,71 +21,68 @@ class DatabaseConfig;
...
+   explicit Partition(std::vector<PartitionChunk>&& chunks);
+
+   explicit Partition(uint32_t partition_id, std::vector<LineageGroup>&& lineage_groups);
+
+   [[nodiscard]] const std::vector<PartitionChunk>& getPartitionChunks() const;
+
+   [[nodiscard]] uint32_t getSequenceCount() const;
+};
+
 enum Architecture { MAX_PARTITIONS, SINGLE_PARTITION, SINGLE_SINGLE };
```
fengelniederhammer, line 79: I suspect that this is also unused now.

include/silo/preprocessing/partition.h
```diff
@@ -21,71 +21,68 @@ class DatabaseConfig;
...
 
 class Partitions {
    std::vector<Partition> partitions;
 
    // Flat map of the counts and sizes of the partitions and containing chunks
-   std::vector<PartitionChunk> partition_chunks;
-
-   // Mapping all pango lineages to the chunk they are contained in
-   std::unordered_map<std::string, silo::preprocessing::PartitionChunk> pango_to_chunk;
+   std::vector<PartitionChunk> all_partition_chunks;
```
fengelniederhammer, line 85: This also seems essentially unused to me.

fengelniederhammer: Please check the whole file. There might be more unused stuff around, e.g. `createSingletonPartitions`

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -0,0 +1,515 @@
...
+#include "silo/preprocessing/sequence_info.h"
+#include "silo/storage/reference_genomes.h"
+#include "silo/zstdfasta/zstd_decompressor.h"
+#include "silo/zstdfasta/zstdfasta_table.h"
+
+namespace silo::preprocessing {
+
+Preprocessor::Preprocessor(
+   const preprocessing::PreprocessingConfig& preprocessing_config,
+   const config::DatabaseConfig& database_config
```
fengelniederhammer, line 27: "Clang-Tidy: Pass by value and use std::move "

src/silo/preprocessing/sequence_info.cpp
```diff
@@ -0,0 +1,178 @@
+#include "silo/preprocessing/sequence_info.h"
+
+#include <boost/algorithm/string/join.hpp>
+#include <duckdb.hpp>
+
+#include "silo/preprocessing/preprocessing_database.h"
```
fengelniederhammer, line 6: I think this import is superfluous

src/silo/query_engine/query_engine.cpp
```diff
@@ -42,16 +42,16 @@ QueryResult QueryEngine::executeQuery(const std::string& query_string) const {
    int64_t filter_time;
    {
       const BlockTimer timer(filter_time);
-      const tbb::blocked_range<size_t> range(0, database.partitions.size(), 1);
-      tbb::parallel_for(range.begin(), range.end(), [&](const size_t& partition_index) {
+      for (size_t partition_index = 0; partition_index != database.partitions.size();
+           partition_index++) {
```
fengelniederhammer, line 46: Is this intentionally still here?

src/silo/storage/build_exception.cpp
```diff
@@ -0,0 +1,10 @@
+#include "silo/storage/build_exception.h"
+
+#include <string>
+
+namespace silo {
+
+BuildException::BuildException(const std::string& error_message)
```
fengelniederhammer, line 7: Is this used anywhere?

src/silo/storage/column/pango_lineage_column.cpp
```diff
@@ -29,6 +29,18 @@ void PangoLineageColumnPartition::insert(const common::RawPangoLineage& value) {
    insertSublineageValues(resolved_lineage, row_number);
 }
 
+void PangoLineageColumnPartition::insertNull() {
+   const common::UnaliasedPangoLineage resolved_lineage = alias_key.unaliasPangoLineage({""});
```
fengelniederhammer, line 33: Is this really a good interpretation of `null`?

src/silo/storage/column/pango_lineage_column.cpp
```diff
@@ -29,6 +29,18 @@ void PangoLineageColumnPartition::insert(const common::RawPangoLineage& value) {
...
+   for (const auto& parent_lineage : resolved_lineage.getParentLineages()) {
+      (void)lookup.getOrCreateId(parent_lineage);
+   }
+   const Idx value_id = lookup.getOrCreateId(resolved_lineage);
+   const size_t row_number = value_ids.size();
+   value_ids.push_back(value_id);
+   indexed_values[value_id].add(row_number);
+   insertSublineageValues(resolved_lineage, row_number);
```
fengelniederhammer, line 41: In case we want to keep it as it is:
```suggestion
   insert({""});
```

src/silo/zstdfasta/zstdfasta_table_reader.test.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+TEST(ZstdFastaTableReader, correctlyReadsZstdFastaTableFromFastaFile) {
+   const std::filesystem::path input_directory{"testBaseData/fastaFiles/"};
+   const std::string sequence_filename{"test.fasta"};
+
+   std::filesystem::path file_path = input_directory / sequence_filename;
+
+   silo::FastaReader file_reader(file_path);
+
+   duckdb::DuckDB duckDB(nullptr);
```
fengelniederhammer, line 21:
```suggestion
   duckdb::DuckDB duck_db(nullptr);
```

src/silo/zstdfasta/zstdfasta_table_reader.test.cpp
fengelniederhammer, line 1: Those tests are very similar. Do we want to use some sort of parametrized tests?

src/silo/zstdfasta/zstdfasta_writer.cpp
fengelniederhammer, line 1: Where is this actually used? I only found a test.

testBaseData/exampleDataset2/database_config.yaml
fengelniederhammer, line 1: Can we find a more speaking name for the directory instead of exampleDataset2? "datasetForUnitTests"?

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,200 @@
...
+      std::back_inserter(config_metadata_fields),
+      [](auto metadata) { return metadata.name; }
+   );
+
+   std::unordered_map<std::string, std::string> validated_metadata_fields;
+   for (const auto& pair : found_metadata_fields) {
+      if (std::find(config_metadata_fields.begin(), config_metadata_fields.end(), pair.first)
+       != config_metadata_fields.end()) {
+         validated_metadata_fields.emplace(pair);
+      }
```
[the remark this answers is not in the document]
Taepper: I thought of it as a feature and not as a problem. This would be one of the only upside of providing a custom config to SILO. I would also imagine that an auto-detect mode for metadata fields and types would be desirable at some point

src/silo/preprocessing/sequence_info.cpp
```diff
@@ -0,0 +1,178 @@
...
+      sequence_selects.emplace_back(fmt::format(
+         "compressAA(alignedAminoAcidSequences.{0}, "
+         "'{0}') as gene_{0}",
+         name
+      ));
+   }
+   return sequence_selects;
+}
+
+std::string SequenceInfo::getNucInsertionSelect() {
```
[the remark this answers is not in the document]
Taepper: Ah yes, it was refactored into metadata_info at some point during this PR ^^

src/silo/preprocessing/preprocessing_database.cpp
```diff
@@ -0,0 +1,204 @@
...
+};
+
+std::unordered_map<std::string_view, tbb::enumerable_thread_specific<silo::ZstdCompressor>>
+   Compressors::nuc_compressors{};
+std::unordered_map<std::string_view, tbb::enumerable_thread_specific<std::string>>
+   Compressors::nuc_buffers{};
+std::unordered_map<std::string_view, tbb::enumerable_thread_specific<silo::ZstdCompressor>>
+   Compressors::aa_compressors{};
+std::unordered_map<std::string_view, tbb::enumerable_thread_specific<std::string>>
+   Compressors::aa_buffers{};
```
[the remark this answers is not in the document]
Taepper: Yeah, this is some magic that is required for the duckdb functions to work and this is required by C++. At least I did not find a nice way to circumvent this

src/silo/preprocessing/preprocessing_database.cpp
```diff
@@ -0,0 +1,204 @@
...
+      if (partition_size_bigint <= 0) {
+         throw PreprocessingException("Non-positive partition size encountered.");
+      }
+      if (partition_size_bigint > UINT32_MAX) {
+         throw PreprocessingException(
+            fmt::format("Overflow of limit UINT32_MAX ({}) for number of sequences.", UINT32_MAX)
+         );
+      }
+
+      uint32_t partition_size = static_cast<uint32_t>(partition_size_bigint);
```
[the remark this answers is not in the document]
Taepper: I do not see the advantage of hiding the type from the user in the cases where it is not clear. I also think that our clang-tidy (which is currently not working) advises against this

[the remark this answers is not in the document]
Taepper: I added the const and the auto after cast

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+void detectInsertionLists(
+   const std::filesystem::path& ndjson_file,
+   std::unordered_map<std::string, std::string>& metadata_fields_to_validate
+) {
+   duckdb::DuckDB db(nullptr);
+   duckdb::Connection connection(db);
+   auto top_level_entries =
+      connection.Query(fmt::format("SELECT * FROM '{}' LIMIT 0", ndjson_file.string()));
```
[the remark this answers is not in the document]
Taepper: It reads exactly 0 lines. duckdb will still figure out the type of the columns by peeking into the values. There it detects that they are structs with certain keys

src/silo/preprocessing/metadata_info.cpp
Taepper: Yes

src/silo/preprocessing/metadata_info.cpp
Taepper: I do not want to marry the implementation here to the exact column names that are looked into. This is pretty flexible to be extended to look for different required fields to be in different structs in the input

src/silo/preprocessing/metadata_info.cpp
Taepper: Here I would again argue for explicit `null`

src/silo/preprocessing/metadata_info.cpp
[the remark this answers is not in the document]
fengelniederhammer: Hm, interesting. So this relies on all entries having the same schema.

conanfile.py
Taepper: I will check again

src/silo/preprocessing/metadata_info.cpp
fengelniederhammer: Ok, then we need to adapt the backend. @chaoran-chen what do you think? Should the backend do it or should the preprocessing pipeline do it?

src/silo/preprocessing/preprocessor.cpp
Taepper: It is part of the SQL I will write comments by function names as you suggested

src/silo/preprocessing/preprocessor.cpp
Taepper: Yes, somewhat. The partitioning will contain the ranges into which the input should be partitioned. This SQL maps the original keys to the range_id they are assigned to

src/silo/zstdfasta/zstdfasta_table.cpp
Taepper: Currently not, but it will be needed to allow more flexible inputs. And the more obvious testing is a great pro as well

fengelniederhammer: At least in the current implementation those are two distinct function. Can we at least extract some common parts?

conanfile.py
Taepper: Is it only on your machine or on the server?

src/silo/zstdfasta/zstdfasta_table.cpp
Taepper: Part of the reason why I created the duckdb wrapper as preprocessing_database was to have a central logging location, but to get the abstractions there correct was somehow a little difficult, thus I made some compromises. I will try to refactor that again

src/silo/zstdfasta/zstdfasta_table_reader.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+std::optional<std::string> silo::ZstdFastaTableReader::nextKey() {
+   if (!current_chunk) {
+      return std::nullopt;
+   }
+
+   return current_chunk->GetValue(0, current_row).GetValue<std::string>();
+}
+
+std::optional<std::string> silo::ZstdFastaTableReader::nextSkipGenome() {
```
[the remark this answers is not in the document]
Taepper: This is for consistency with the other Fasta readers

include/silo/zstdfasta/zstdfasta_table_reader.h
```diff
@@ -0,0 +1,53 @@
...
+      std::string_view compression_dict,
+      std::string_view where_clause,
+      std::string_view order_by_clause
+   );
+
+   std::optional<std::string> nextSkipGenome();
+
+   std::optional<std::string> next(std::string& genome);
+
+   std::optional<std::string> nextCompressed(std::string& compressed_genome);
```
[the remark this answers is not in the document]
Taepper: Huh, I had in a previous intermediate version used the function from outside. When streaming from the in-memory table to a file in zstdfasta format. So I think there is a usecase to have the interface facing outside

src/silo/zstdfasta/zstdfasta_table_reader.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <spdlog/spdlog.h>
+#include <duckdb.hpp>
+#include <duckdb/common/types/blob.hpp>
+
+#include "silo/common/fasta_format_exception.h"
+#include "silo/preprocessing/preprocessing_exception.h"
+#include "silo/zstdfasta/zstd_decompressor.h"
+
+silo::ZstdFastaTableReader::ZstdFastaTableReader(
+   duckdb::Connection& connection,
```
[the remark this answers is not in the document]
Taepper: Ah, that is actually not possible. You cannot copy the connection.

src/silo/zstdfasta/zstdfasta_table_reader.cpp
```diff
@@ -0,0 +1,112 @@
...
+   std::string_view compression_dict,
+   std::string_view where_clause,
+   std::string_view order_by_clause
+)
+    : connection(connection),
+      table_name(table_name),
+      where_clause(where_clause),
+      order_by_clause(order_by_clause),
+      decompressor(std::make_unique<ZstdDecompressor>(compression_dict)),
+      DEBUG_dictionary(compression_dict) {
```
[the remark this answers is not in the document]
Taepper: Which of these?

src/silo/zstdfasta/zstdfasta_table_reader.cpp
```diff
@@ -0,0 +1,112 @@
...
+   std::string_view table_name,
+   std::string_view compression_dict,
+   std::string_view where_clause,
+   std::string_view order_by_clause
+)
+    : connection(connection),
+      table_name(table_name),
+      where_clause(where_clause),
+      order_by_clause(order_by_clause),
+      decompressor(std::make_unique<ZstdDecompressor>(compression_dict)),
```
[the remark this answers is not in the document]
Taepper: Oh yeah, this was a little cumbersome to deal with. In short: Because the ZstdDecompressor uses zstd library components that are neither trivially copyable / moveable nor are easy to write move constructors for (I tried, but many months ago, when introducing Zstd(De)compressors). I want the table reader to be moveable however. A unique ptr makes this possible, as it is moveable (a moved ptr stays a valid pointer)

include/silo/zstdfasta/zstdfasta_table.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <string>
+
+namespace duckdb {
+struct Connection;
+}
+
+namespace silo {
```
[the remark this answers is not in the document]
Taepper: True, then I should also rename the classes to not contain the namespace at the start again. Do you think this would mess up the change list too much and I do it separately? (And I do #139 in the near future)

src/silo/zstdfasta/zstd_decompressor.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void ZstdDecompressor::decompress(const std::string& input, std::string& output) {
+   auto return_code = ZSTD_decompress_usingDDict(
+      zstd_context, output.data(), output.length(), input.data(), input.size(), zstd_dictionary
+   );
+   if (ZSTD_isError(return_code)) {
+      std::string error_name = ZSTD_getErrorName(return_code);
+      throw std::runtime_error(
+         "Error '" + error_name + "' in dependency when decompressing using zstd."
```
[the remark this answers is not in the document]
Taepper: What would you suggest?

src/silo/zstdfasta/zstdfasta_table.cpp
fengelniederhammer: The `preprocessing_database` wrapper definitely makes sense. I think this is just some detail where we can remove duplication. Maybe it can also be a normal function (since macros in C++ are kinda nasty)?

src/silo/zstdfasta/zstdfasta_table_reader.cpp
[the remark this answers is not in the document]
fengelniederhammer: I see, thanks.

include/silo/zstdfasta/zstd_decompressor.h
```diff
@@ -12,15 +12,23 @@ class ZstdDecompressor {
    ZSTD_DCtx* zstd_context;
 
   public:
+   ZstdDecompressor(ZstdDecompressor&& other);
+   ZstdDecompressor& operator=(ZstdDecompressor&& other);
```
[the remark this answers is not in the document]
Taepper: Such that we can move objects, that contain a ZstdDecompressor

src/silo/preprocessing/sequence_info.cpp
```diff
@@ -49,15 +30,17 @@ std::vector<std::string> SequenceInfo::getSequenceSelects() {
    std::vector<std::string> sequence_selects;
    for (const std::string& name : nuc_sequence_names) {
       sequence_selects.emplace_back(fmt::format(
```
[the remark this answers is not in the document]
Taepper: Irrelevant in terms of optimization unless we have >different sequences

src/silo/zstdfasta/zstdfasta_writer.cpp
Taepper: I would still like to hold the internal interface consistent, where we have readers and writers for these file types

src/silo/zstdfasta/zstdfasta_table_reader.test.cpp
Taepper: If there comes up a need for that. But for now I think this is fine?

src/silo/query_engine/query_engine.cpp
Taepper: Actually yes, while the exception were still causing me some headache

src/silo/preprocessing/partition.test.cpp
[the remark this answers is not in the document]
Taepper: I would do this once the duckdb steps are not prone to change much more

include/silo/zstdfasta/zstdfasta_table_reader.h
```diff
@@ -0,0 +1,53 @@
...
+   duckdb::Connection& connection;
+   std::string table_name;
+   std::string where_clause;
+   std::string order_by_clause;
+   std::unique_ptr<duckdb::MaterializedQueryResult> query_result;
+   std::unique_ptr<duckdb::DataChunk> current_chunk;
+   std::unique_ptr<silo::ZstdDecompressor> decompressor;
+   size_t current_row;
+
+   // std::string DEBUG_dictionary;
```
fengelniederhammer, line 30: Can we delete this?

src/silo/preprocessing/preprocessing_config.cpp
```diff
@@ -52,6 +26,8 @@ PreprocessingConfig::PreprocessingConfig(
    const InputDirectory& input_directory_,
    const IntermediateResultsDirectory& intermediate_results_directory_,
    const OutputDirectory& output_directory_,
+   const PreprocessingDatabaseLocation& preprocessing_database_location_,
+   const NdjsonInputFilename& ndjson_input_filename_,
    const MetadataFilename& metadata_filename_,
    const PangoLineageDefinitionFilename& pango_lineage_definition_filename_,
    const PartitionsFolder& partition_folder_,
```
fengelniederhammer, line 33: `partition_folder_` and `sorted_partition_folder_` seem to be unused here now.

fengelniederhammer (approved):
Looks a lot better :+1: Maybe not perfect, but good enough. We can still improve some details later on the way. :)

Issue #220, opened 2023-10-25
Allow segments to be `null`


fengelniederhammer:
We don't want to require that all sequences have all segments or genes present. Instead allow that some segments/genes are `null`.

fengelniederhammer:
https://github.com/GenSpectrum/LAPIS-SILO/pull/239 contains example data.

PR #239, opened 2023-12-22
allow missing segments and genes

fengelniederhammer:
for #220. The input file contains metadata that are `null` (but all metadata are present) and segments and genes that are `null


include/silo/zstdfasta/zstdfasta_table_reader.h
```diff
@@ -42,9 +44,9 @@ class ZstdFastaTableReader {
 
    std::optional<std::string> nextSkipGenome();
 
-   std::optional<std::string> next(std::string& genome);
+   std::optional<std::string> next(std::optional<std::string>& genome);
```
Taepper, line 47: Hmm I am not sure whether it is just me, but I find it a bit confusing, that we now have two optionals here, that refer to different things. But I also did not find a way around that that is elegant.
Of course the idiomatic thing to do, would be to return a `std::optional` of a custom data type, that does now have a key and an optional string. But this was initially not chosen, as it can be often much more efficient (especially for long strings), to write them directly into a preallocated string reference, given as parameter.

Also I am not sure, whether there the performance penalty is already paid by writing into a `std::optional<std::string>`

Taepper: Would be happy to merge it like this, though. What do you think?

Taepper (approved):

include/silo/zstdfasta/zstdfasta_table_reader.h
fengelniederhammer: As discussed, we leave it as it is for now and maybe investigate again later whether we have performance issues.

PR #243, opened 2024-01-09
Fix amino acid insertions

JonasKellerer:
Resolves the issue, that for the large dataset, amino acid insertions are not read in properly. It seems to be an issue with duckdb.


JonasKellerer:
I now moved the fix to the insertion column. 
With the large dataset it takes roughly the same time.

src/silo/common/string_utils.cpp
```diff
@@ -19,4 +19,12 @@ std::vector<std::string> splitBy(const std::string& value, const std::string_vie
    return splits;
 }
 
+std::string removeSymbol(const std::string& value, const std::string_view symbol) {
+   std::string result = value;
+   const auto symbol_position = result.find(symbol);
+   if (symbol_position != std::string::npos) {
+      result.erase(symbol_position, symbol.size());
+   }
+   return result;
```
Taepper, line 28: This only removes up to one quote. Lets also cover the case where on string is completely quoted

```
return value.erase(
      std::remove(value.begin(), value.end(), '\"'),
      value.end()
   );
```

Taepper: (remove all " from the string, not only the first one found)

Taepper (approved):

src/silo/common/string_utils.cpp
JonasKellerer: I changed it, so it removes all. Additionally, I added tests to verify the behavior.

PR #245, opened 2024-01-15
fix: orderBy descending and Limit leading to wrong results

Taepper:
fixes https://github.com/GenSpectrum/LAPIS/issues/572


fengelniederhammer (approved):
LGTM

PR #249, opened 2024-01-18
build: make linter throw errors again and fix them 

### LINKED ISSUE #202: Check linter in CI
The linter does not seem to throw an error in the CI, when clang-tidy finds a problem. This should be fixed.

fengelniederhammer:
resolves #202


Taepper (approved):
LGTM! Although I think we once wanted to not make the linter check the imports, but rather stick to the IWYU (include-what-you-use) provided imports (many of which the linter would want to discard). This was a suggestion by @JonasKellerer and I ran iwyu periodically (2-3 months) as it is a little bit of an effort. I would leave it as it is now in this PR, but generally we should think about which way we want to do it

fengelniederhammer:
Generally I agree with IWYU - I didn't check too much though, and also it was mostly my IDE that complained about the unused imports, so I cleaned up a couple of unused includes on the way.

Issue #112, opened 2023-06-02
Efficient download of large sets of sequences


chaoran-chen:
We would like to enable users to retrieve large sets of sequences (for example, several million) from our database efficiently. The current approach of reconstructing the sequence from the indices is not very efficient.

- It should work for all types of sequences: aligned nucleotide sequences, not aligned nucleotide sequences, and aligned AA sequences.
- The sequences should be returned in a compressed format. Unlike with the metadata/aggregated data, the data are too big to be moved around in an uncompressed format. zstd is a good and popular format, gzip should probably also be offered.

@Taepper mentioned recently that it might be possible to precompress the sequences and concatenate the compressed data directly (i.e., without decompressing and recompressing). I quickly looked into the [zstd compression format](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#zstandard-frames): on first glance, I can't see why we wouldn't be able to precompress and store each sequence in an own _data block_ and just concatenate the _data blocks_. We can probably use a custom dictionary and as long as all sequences are precompressed with the same dictionary, it should work - right? Hopefully, something similar is possible for gzip?

For LAPIS v1, we currently store each sequence individually with zstd compression. For the >15 million sequences, the aligned nucleotide sequences need 3.4 GB, the unaligned nucleotide sequences need 27 GB (maybe we can improve by using a better dictionary?), and the aligned AA sequences need 5.5 GB. That's not too much, so I think that we can even hold them all in RAM for the start. Later, we can add a low-RAM option where they are kept on disk.

I imagine that with this approach, we can _stream_ the responses. That means we don't need to first construct the whole response and can almost immediately start returning results. The CPU & RAM usage should be minimal.

Do you think that this works and is a good approach?

chaoran-chen:
Not ready to use now but potentially interesting for the future: There is a proposal to enable servers and browsers/clients to share compression dictionaries.

https://github.com/WICG/compression-dictionary-transport
https://chromestatus.com/feature/5124977788977152

PR #255, opened 2024-01-23
Unaligned Sequences

Taepper:
feat: introduce storage of unaligned sequences from either ndjson file or fasta file and make them queryable via the Fasta action

The last required action

 #112 is now very easy (for unaligned), where we can stream the sequences almost directly from the backing file


.github/workflows/ci.yml
```diff
@@ -35,13 +35,10 @@ jobs:
     name: Build And Run linter
     runs-on: ubuntu-latest
     steps:
-      -
-        uses: actions/checkout@v4
-      -
-        name: Set up Docker Buildx
+      - uses: actions/checkout@v4
```
JonasKellerer, line 38: If we change the style here, I would favor changing it for the whole file. So also in the section endToEndTests

.run/silo--api.run.xml
```diff
@@ -1,10 +1,15 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="silo --api" type="CMakeRunConfiguration" factoryName="Application" PROGRAM_PARAMS="--api --dataDirectory output/" REDIRECT_INPUT="false" ELEVATE="false" USE_EXTERNAL_CONSOLE="false" EMULATE_TERMINAL="false" WORKING_DIR="file://.." PASS_PARENT_ENVS_2="true" PROJECT_NAME="SILO" TARGET_NAME="siloApi" CONFIG_NAME="Debug" RUN_TARGET_PROJECT_NAME="SILO" RUN_TARGET_NAME="siloApi">
-    <envs>
-      <env name="SPDLOG_LEVEL" value="debug" />
-    </envs>
-    <method v="2">
-      <option name="com.jetbrains.cidr.execution.CidrBuildBeforeRunTaskProvider$BuildBeforeRunTask" enabled="true" />
-    </method>
-  </configuration>
+    <configuration default="false" name="silo --api" type="CMakeRunConfiguration" factoryName="Application"
```
JonasKellerer, line 2: This does not change the functionality, does it? My Clion just revertet this back to the previous format. So I would not add this change here.

.run/silo--preprocessing.run.xml
```diff
@@ -1,10 +1,15 @@
 <component name="ProjectRunConfigurationManager">
-    <configuration default="false" name="silo --preprocessing" type="CMakeRunConfiguration" factoryName="Application" PROGRAM_PARAMS="--preprocessing --preprocessingConfig=testBaseData/test_preprocessing_config.yaml --databaseConfig=testBaseData/exampleDataset/database_config.yaml" REDIRECT_INPUT="false" ELEVATE="false" USE_EXTERNAL_CONSOLE="false" EMULATE_TERMINAL="false" WORKING_DIR="file://.." PASS_PARENT_ENVS_2="true" PROJECT_NAME="SILO" TARGET_NAME="siloApi" CONFIG_NAME="Debug" RUN_TARGET_PROJECT_NAME="SILO" RUN_TARGET_NAME="siloApi">
+    <configuration default="false" name="silo --preprocessing" type="CMakeRunConfiguration" factoryName="Application"
```
JonasKellerer, line 2: Same as above

endToEndTests/test/queries/manySequences.json
```diff
@@ -0,0 +1,44 @@
+{
```
JonasKellerer, line 1: Can we rename this file to fasta_manySequences? Then it will be next to the other fastaAligned test.

endToEndTests/test/queries/oneRowTwoUnalignedSequences.json
```diff
@@ -0,0 +1,21 @@
+{
```
JonasKellerer, line 1: Same here. It would help to find the test. But on thinking about it: maybe we could put the test files now in folders. So those which test the endpoints each in one, and then the filters in another. But this can also be done in another PR.

endToEndTests/test/queries/oneSequenceUnaligned.json
```diff
@@ -0,0 +1,20 @@
...
+    "filterExpression": {
+      "type": "StringEquals",
+      "column": "gisaid_epi_isl",
+      "value": "EPI_ISL_1749899"
+    }
+  },
+  "expectedQueryResult": [
+    {
+      "gisaid_epi_isl": "EPI_ISL_1749899",
+      "main": "some_very_short_string"
```
JonasKellerer, line 17: Should the e2e tests also work with the data from `exampleDataset`? Then we would need this string also in the fasta file.

README.md
```diff
@@ -23,7 +23,7 @@ decided for the approach with less maintenance effort, since it will automatical
 ### Running SILO locally with CLion
```
JonasKellerer, line 23: I think it would be nice do document the temp folder somewhere. I'm not sure if this should be here, or in a separate documentation. We could also add a ticket for that.

JonasKellerer (approved):
Thank you. Nice addition.

My comments are mostly for small refactors. However, the `add_sequences_to_results_for_partition` function was quite hard to read. If possible, splitting into smaller function would help.

include/silo/preprocessing/preprocessing_config_reader.h
```diff
@@ -17,13 +17,13 @@ struct OptionalPreprocessingConfig {
     */
    std::optional<std::filesystem::path> output_directory;
    /**
-    * The directory where the intermediate results will be stored
+    * The directory where temporary results will be stored
     * that are not relevant for an end user.
     */
-   std::optional<std::filesystem::path> intermediate_results_directory;
+   std::optional<std::filesystem::path> temp_directory;
```
fengelniederhammer, line 23: Please change accordingly in the LAPIS docs: https://github.com/GenSpectrum/LAPIS/blob/main/lapis2-docs/src/content/docs/maintainer-docs/references/preprocessing.mdx

.run/silo--api.run.xml
Taepper: Huh, my Clion changed it actually to what I have there now. There was an update to clang tidy, do you have that as well?

.github/workflows/ci.yml
Taepper: Huh, my Clion changed it actually to what I have there now. There was an update to clang tidy, do you have that as well?
Can you double check, I will revert it, if spurious on my part

include/silo/preprocessing/preprocessing_config_reader.h
fengelniederhammer: We can also make it part of https://github.com/GenSpectrum/LAPIS/issues/581

Taepper: Reverted it

.github/workflows/ci.yml
JonasKellerer: Hm... It did the same for me. So we can leave this as is.

PR #257, opened 2024-01-23
feat: return only aliased pango lineages

### LINKED ISSUE #242: Return alias of pango lineage in response
LAPIS v1 returns the alias of a pango lineage, if requested. For example an aggregated request with fields `NEXTCLADEPANGOLINEAGE` returns `{ nextcladePangoLineage: 'JN.1.1.1', count: 2 },`. 

Currently, SILO returns `{ count: 2, nextcladePangoLineage: 'B.1.1.529.2.86.1.1.1.1' },`. 

When I recall correctly, we decided, that SILO should also return the alias, if not queried with an additional option?

JonasKellerer:
Resolves #242


Taepper (approved):
LGTM

PR #260, opened 2024-01-24
fix: allow sql keywords for metadata field names

### LINKED ISSUE #259: Allow SQL Keywords as Metadata Names
Currently, the metadata field name "group" causes a parsing error.

fengelniederhammer:
resolves #259


Taepper (approved):
LGTM. I really like the testing Scenarios! We might also try to generate test files directly in the tests (possibly with duckdb), if we want to test various more generic things for larger files

PR #262, opened 2024-01-25
fix: test and address sanatizer after merge of unaligned and sql statements

JonasKellerer:
They did not show up in the ci, but if you run the tests locally.

### DIFF
--- a/include/silo/preprocessing/preprocessing_database.h
+++ b/include/silo/preprocessing/preprocessing_database.h
@@ -11,6 +11,7 @@ namespace silo {
 
 class ZstdFastaTable;
 class ReferenceGenomes;
+class CompressSequence;
 
 namespace preprocessing {
 
@@ -18,23 +19,24 @@ class Partitions;
 
 class PreprocessingDatabase {
   public:
-   static constexpr std::string_view COMPRESS_NUC = "compressNuc";
-   static constexpr std::string_view COMPRESS_AA = "compressAA";
+   std::unique_ptr<CompressSequence> compress_nucleotide_function;
+   std::unique_ptr<CompressSequence> compress_amino_acid_function;
 
   private:
    duckdb::DuckDB duck_db;
    duckdb::Connection connection;
 
   public:
-   PreprocessingDatabase(const std::optional<std::filesystem::path>& backing_file);
+   PreprocessingDatabase(
+      const std::optional<std::filesystem::path>& backing_file,
+      const ReferenceGenomes& reference_genomes
+   );
 
    duckdb::Connection& getConnection();
 
    void refreshConnection();
 
    Partitions getPartitionDescriptor();
-
-   static void registerSequences(const silo::ReferenceGenomes& reference_genomes);
 
    std::unique_ptr<duckdb::MaterializedQueryResult> query(std::string sql_query);
 
--- a/src/silo/preprocessing/preprocessing_database.cpp
+++ b/src/silo/preprocessing/preprocessing_database.cpp
@@ -12,6 +12,7 @@
 #include "silo/common/fasta_reader.h"
 #include "silo/preprocessing/partition.h"
 #include "silo/preprocessing/preprocessing_exception.h"
+#include "silo/preprocessing/sql_function.h"
 #include "silo/storage/reference_genomes.h"
 #include "silo/zstdfasta/zstd_compressor.h"
 #include "silo/zstdfasta/zstdfasta_reader.h"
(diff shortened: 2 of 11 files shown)


src/silo/preprocessing/preprocessing_compressor.cpp
```diff
@@ -0,0 +1,94 @@
...
+               .local()
+               .compress(
+                  uncompressed.GetData(), uncompressed.GetSize(), buffer.data(), buffer.size()
+               );
+         return StringVector::AddStringOrBlob(
+            result, buffer.data(), static_cast<uint32_t>(size_or_error_code)
+         );
+      }
+   );
+};
```
fengelniederhammer, line 64:
```suggestion
}
```

include/silo/preprocessing/preprocessing_compressor.h
```diff
@@ -0,0 +1,35 @@
...
+class ZstdCompressor;
+}  // namespace silo
+
+namespace silo::preprocessing {
+
+class Compressor {
+  public:
+   static void initialize(const silo::ReferenceGenomes& reference_genomes);
+
+   static void compressNuc(
```
fengelniederhammer, line 18: Do those function really need to be static? As far as I can judge this: Those function are registered as user defined functions in the particular DuckDB instance that lives in the `PreprocessingDatabase`. I.e. the function need the lifetime of the `PreprocessingDatabase`.
 -> Would it be enough to let the `PreprocessingDatabase` hold an instance of the compressor - making the compressor functions non-static?

Maybe this requires a slight refactoring such that the `preprocessing_db` in the `Preprocessor` is initialized after reading the reference genomes (because the constructor needs the reference genomes to construct and hold the compressors such that they live as long as the DuckDB instance).

JonasKellerer: I think the problem is not with these functions, that they are static. (It is the same as having them in a namespace). But rather, that we have the static maps, which they use. 
But I like your idea. This way, we would not have to reset them. Static variables are always a problem.

fengelniederhammer: Yes sure, the static caches are the problem. But most probably, by the caches non-static, those functions here will also have to be non-static.

fengelniederhammer: In the end it's always the same story: Everything that's stored statically somewhere produces issues in testing.

Taepper: I did not want to have them static as well. The call to connection.CreateVectorizedFunction is the culprit here.

I needed to refactor it to a static function (and also those ugly static maps) because I could not register a non-static function as an UDF. Maybe you are able to?

fengelniederhammer: Ok, maybe we'll give it a shot :)

Taepper (commented):
Wow, I like this!

Now that you managed to get it done without the `static` functions, we could also get rid of the circumvention that we need to pass the `sequence_name` and instead register a separate function for each sequence?

src/silo/preprocessing/preprocessing_database.cpp
```diff
@@ -30,97 +31,26 @@ using duckdb::StringVector;
...
-   );
-
-   connection.CreateVectorizedFunction(
-      std::string(COMPRESS_AA),
-      {LogicalType::VARCHAR, LogicalType::VARCHAR},
-      LogicalType::BLOB,
-      Compressors::compressAA
-   );
+   compress_nucleotide_function->applyTo(connection);
+   compress_amino_acid_function->applyTo(connection);
```
Taepper, line 53: `applyTo` seems a little unintuitive here, I think?

JonasKellerer: I am open for suggestions. How about 'addToConnection'?

Taepper (approved):
Great! Yes, let's do that in another PR. Can you open an issue for that?

include/silo/preprocessing/preprocessing_database.h
```diff
@@ -11,31 +11,33 @@ namespace silo {
...
 namespace preprocessing {
 
 class Partitions;
 
 class PreprocessingDatabase {
   public:
-   static constexpr std::string_view COMPRESS_NUC = "compressNuc";
-   static constexpr std::string_view COMPRESS_AA = "compressAA";
+   std::unique_ptr<CompressSequence> compress_nucleotide_function;
+   std::unique_ptr<CompressSequence> compress_amino_acid_function;
```
Taepper, line 23: Does this actually need to be a `unique_ptr`? I don't think this is a blocker here or something, but I just wanted to mention it, as when we extend it to multiple functions, we will not need to store a `vector<unique_ptr>`/`map<unique_ptr>`, but can store the objects in the container directly

JonasKellerer:
I added a new ticket for the sql functions for each of the sequences #282

include/silo/preprocessing/preprocessing_database.h
JonasKellerer: As discussed: It was only a pointer, so we can extend this in the future with more of the SqlFunctions. However, we should address this in #282

PR #264, opened 2024-01-26
refactor: use node testrunner instead of mocha

fengelniederhammer:
I've tried using the [node test runner](https://nodejs.org/api/test.html) to run the e2e tests, but I'm not sure whether it's worth the trouble. The changes are quite straight forward. There is a small benefit of having on less dependency.


JonasKellerer (approved):
LGTM :)

PR #267, opened 2024-01-30
Optimize bitmaps for more memory-efficiency

Taepper:
The very memory efficient mode is now implemented:
The most numerous bitmap is deleted instead of flipped. When looking up values, all other bitmaps are considered instead.

A good next step would be a hybrid mode, where the deletion is only preferred over flipping, if there is substantial memory gains, as the performance impact is quite high

### DIFF
--- a/include/silo/common/aa_symbols.h
+++ b/include/silo/common/aa_symbols.h
@@ -39,7 +39,7 @@ class AminoAcid {
       STOP,  // Stop codon
       X,     // Any amino acid
    };
-   static constexpr uint32_t COUNT = static_cast<uint32_t>(Symbol::X) + 1;
+   static constexpr uint32_t COUNT = 25;
 
    static constexpr std::string_view SYMBOL_NAME = "Amino Acid";
    static constexpr std::string_view SYMBOL_NAME_LOWER_CASE = "amino acid";
@@ -77,6 +77,14 @@ class AminoAcid {
       AminoAcid::Symbol::Y,     // Tyrosine
       AminoAcid::Symbol::STOP,  // Stop codon, Star-character in sequence
    };
+
+   static constexpr std::array<Symbol, 3> INVALID_MUTATION_SYMBOLS{
+      AminoAcid::Symbol::B,  // Aspartic acid or Asparagine
+      AminoAcid::Symbol::Z,  // Glutamine or Glutamic acid
+      AminoAcid::Symbol::X,  // Any amino acid
+   };
+
+   static_assert(INVALID_MUTATION_SYMBOLS.size() + VALID_MUTATION_SYMBOLS.size() == SYMBOLS.size());
 
    static constexpr Symbol SYMBOL_MISSING = Symbol::X;
 
--- a/src/silo/query_engine/actions/fasta_aligned.cpp
+++ b/src/silo/query_engine/actions/fasta_aligned.cpp
@@ -59,21 +59,20 @@ std::string reconstructSequence(
       reconstructed_sequence[position_id] = SymbolType::symbolToChar(symbol);
    }
 
-   tbb::
-      parallel_for(
-         tbb::blocked_range<size_t>(0, sequence_store.positions.size()),
-         [&](const auto local) {
-            for (auto position_id = local.begin(); position_id != local.end(); position_id++) {
-               const Position<SymbolType>& position = sequence_store.positions.at(position_id);
-               for (const auto symbol : SymbolType::SYMBOLS) {
-                  if (symbol != position.symbol_whose_bitmap_is_flipped &&
-                      position.bitmaps.at(symbol).contains(sequence_id)) {
-                     reconstructed_sequence[position_id] = SymbolType::symbolToChar(symbol);
-                  }
+   tbb::parallel_for(
+      tbb::blocked_range<size_t>(0, sequence_store.positions.size()),
+      [&](const auto local) {
+         for (auto position_id = local.begin(); position_id != local.end(); position_id++) {
+            const Position<SymbolType>& position = sequence_store.positions.at(position_id);
+            for (const auto symbol : SymbolType::SYMBOLS) {
+               if (!position.isSymbolFlipped(symbol) && !position.isSymbolDeleted(symbol)
+                   && position.getBitmap(symbol)->contains(sequence_id)) {
+                  reconstructed_sequence[position_id] = SymbolType::symbolToChar(symbol);
                }
             }
          }
-      );
+      }
+   );
 
    for (const size_t position : sequence_store.missing_symbol_bitmaps.at(sequence_id)) {
       reconstructed_sequence[position] = SymbolType::symbolToChar(SymbolType::SYMBOL_MISSING);
--- a/src/silo/query_engine/actions/mutations.cpp
+++ b/src/silo/query_engine/actions/mutations.cpp
@@ -63,63 +63,73 @@ std::unordered_map<std::string, typename Mutations<SymbolType>::PrefilteredBitma
 }
 
 template <typename SymbolType>
-void Mutations<SymbolType>::addMutationsCountsForPosition(
+void Mutations<SymbolType>::addPositionToMutationCountsForMixedBitmaps(
    uint32_t position,
    const PrefilteredBitmaps& bitmaps_to_evaluate,
    SymbolMap<SymbolType, std::vector<uint32_t>>& count_of_mutations_per_position
 ) {
    for (const auto& [filter, sequence_store_partition] : bitmaps_to_evaluate.bitmaps) {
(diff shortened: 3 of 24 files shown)


src/silo/query_engine/actions/mutations.cpp
```diff
@@ -69,29 +69,88 @@ void Mutations<SymbolType>::addMutationsCountsForPosition(
...
+   }
+
+   correctMutationCountForDeletedSymbolBitmaps(
+      position, bitmaps_to_evaluate, count_of_mutations_per_position
+   );
+}
+
+template <typename SymbolType>
+// NOLINTNEXTLINE(readability-function-cognitive-complexity)
+void Mutations<SymbolType>::correctMutationCountForDeletedSymbolBitmaps(
```
fengelniederhammer, line 112: Why can't we incorporate those steps in the function above instead of the `if isSymbolDeleted { continue }`? IMO that would make the flow a little more obvious.

src/silo/query_engine/actions/fasta_aligned.cpp
```diff
@@ -59,21 +59,19 @@ std::string reconstructSequence(
...
-                      position.bitmaps.at(symbol).contains(sequence_id)) {
-                     reconstructed_sequence[position_id] = SymbolType::symbolToChar(symbol);
-                  }
+   tbb::parallel_for(
+      tbb::blocked_range<size_t>(0, sequence_store.positions.size()),
+      [&](const auto local) {
+         for (auto position_id = local.begin(); position_id != local.end(); position_id++) {
+            const Position<SymbolType>& position = sequence_store.positions.at(position_id);
+            for (const auto symbol : SymbolType::SYMBOLS) {
+               if (!position.isSymbolFlipped(symbol) && position.getBitmap(symbol)->contains(sequence_id)) {
```
fengelniederhammer, line 68: Why don't we need to account for deleted symbols here?

fengelniederhammer (commented):
I just have a couple of question, but otherwise seems fine.

Is it really worth the trouble though? The reduction of the size numbers in the tests doesn't seem overwhelming...

src/silo/query_engine/filter_expressions/or.cpp
```diff
@@ -33,7 +34,10 @@ std::string Or::toString(const silo::Database& database) const {
       children.begin(),
       children.end(),
       std::back_inserter(child_strings),
-      [&](const std::unique_ptr<Expression>& child) { return child->toString(database); }
+      [&](const std::unique_ptr<Expression>& child) {
+         SPDLOG_TRACE("- {}", child->toString(database));
```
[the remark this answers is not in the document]
Taepper: Ah, yes. I hunted down a bug there, where the toString was throwing an error haha

src/silo/storage/position.cpp
```diff
@@ -0,0 +1,160 @@
...
+      const uint32_t count =
+         isSymbolFlipped(symbol) ? sequence_count - bitmap.cardinality() : bitmap.cardinality();
+      count_sum += count;
+      if (count > max_count) {
+         max_symbol = symbol;
+         max_count = count;
+      }
+   }
+   if (symbol_whose_bitmap_is_deleted.has_value()) {
+      if (sequence_count - count_sum > max_count) {
```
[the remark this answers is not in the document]
Taepper: count_sum is the number of sequences with a symbol different from `X` at that position. sequence_count will therefore always be bigger than count_sum.
We want to know, whether 
But actually, this really is not returning what is necessarily the most numerous symbol, as `count_sum` does not contain the number of symbols that have `N` at that position. I guess it should be disallowed for Positions with Deleted symbols, as are the other functions

src/silo/storage/position.test.cpp
```diff
@@ -11,54 +11,219 @@
...
 
-void deserializeFromFile(const std::string& filename, silo::Position<silo::Nucleotide>& position) {
+void deserializeFromFile(const std::string& filename, Position<Nucleotide>& position) {
    std::ifstream input_file(filename, std::ios::binary);
    ::boost::archive::binary_iarchive input_archive(input_file);
    input_archive >> position;
    input_file.close();
 }
 
+TEST(Position, flipsMostNumerousCorrectlyFromUnset) {
```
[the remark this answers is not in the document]
Taepper: I took the `FromUnset` from a previous test by someone else. Are you happy with `FromUnoptimized`, which I would prefer as not only Flipped is not set, but also Deleted

src/silo/query_engine/actions/fasta_aligned.cpp
Taepper: True, we can do that. It is not incorrect, but it is a small optimisation as we would not need to call the bitmap::contains function (which will always return false)

include/silo/common/aa_symbols.h
```diff
@@ -39,7 +39,7 @@ class AminoAcid {
       STOP,  // Stop codon
       X,     // Any amino acid
    };
-   static constexpr uint32_t COUNT = static_cast<uint32_t>(Symbol::X) + 1;
+   static constexpr uint32_t COUNT = 25;
```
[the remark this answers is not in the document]
Taepper: I did find it a lot clearer. It is checked nonetheless, as the std::array of all symbols needs to have the correct size

src/silo/query_engine/actions/mutations.cpp
Taepper: I refactored it a little, which should make the flow more obvious

Taepper:
I see your point that the effect seems small for the test data set if one looks only at storage consumption. There the static overhead mostly hides the effect because of the few sequences. Although, looking at the `values_store_in_X_container` shows the rather big savings already. It will be much more apparent for bigger datasets, where the effect will be closer to 90%.

src/silo/storage/position.test.cpp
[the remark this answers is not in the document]
fengelniederhammer: Sounds good :+1:

src/silo/query_engine/actions/fasta_aligned.cpp
Taepper: Yes, exactly. I think saying that we start with the reference sequence **of the partition** is a good way to put it. That is, we change the reference sequence with regard to all members of `indexing_differences_to_reference_sequence`. This then changed all positions where the optimized symbol is different from the reference sequence we started from.

src/silo/query_engine/actions/mutations.cpp
```diff
@@ -63,35 +63,92 @@ std::unordered_map<std::string, typename Mutations<SymbolType>::PrefilteredBitma
...
+         }
+         const uint32_t symbol_count =
+            current_position.isSymbolFlipped(symbol)
+               ? filter->andnot_cardinality(*current_position.getBitmap(symbol))
+               : filter->and_cardinality(*current_position.getBitmap(symbol));
+
+         count_of_mutations_per_position[symbol][position] += symbol_count;
+
+         const auto deleted_symbol = current_position.getDeletedSymbol();
+         if (deleted_symbol.has_value() && symbol != *deleted_symbol) {
```
fengelniederhammer, line 86: Is this different from `!current_position.isSymbolDeleted(symbol)`?

src/silo/storage/position.cpp
```diff
@@ -0,0 +1,160 @@
...
+      ));
+   }
+
+   std::optional<typename SymbolType::Symbol> max_symbol = std::nullopt;
+   uint32_t max_count = 0;
+
+   for (const auto& symbol : SymbolType::SYMBOLS) {
+      roaring::Roaring& bitmap = bitmaps[symbol];
+      bitmap.runOptimize();
+      bitmap.shrinkToFit();
```
fengelniederhammer, line 59: I don't know whether it's good to hide this side effect in a `getSomething` method.

src/silo/query_engine/actions/mutations.cpp
```diff
@@ -63,35 +63,92 @@ std::unordered_map<std::string, typename Mutations<SymbolType>::PrefilteredBitma
...
+      const auto deleted_symbol = sequence_store_partition.positions[position].getDeletedSymbol();
+      if (deleted_symbol == std::nullopt) {
+         continue;
+      }
+      for (const uint32_t idx : *filter) {
+         const roaring::Roaring& n_bitmap = sequence_store_partition.missing_symbol_bitmaps[idx];
+         if (n_bitmap.contains(position)) {
+            count_of_mutations_per_position[*deleted_symbol][position] -= 1;
+         }
+      }
```
fengelniederhammer, line 142: As discussed, let's inline this into `addPositionToMutationCountsForMixedBitmaps` and get rid of the separate `correct...` method.

fengelniederhammer (approved):
LGTM

PR #272, opened 2024-01-31
refactor: always have type members first in struct

Taepper:
Small refactor to always list type members first in the struct and class definitions


fengelniederhammer (approved):
LGTM

PR #277, opened 2024-02-04
Optimized preprocessing

Taepper:
Numerous optimizations that lead to both faster and much more memory-efficient preprocessing.

Most importantly, the unaligned sequences are now never held in memory, but directly hive-partitioned by partition_id into intermediate storage location. From there they will be copied to the save location. 

gisaid LAPIS builds now run with ~80 GB in preprocessing.

Various logging improvements during preprocessing


JonasKellerer:
Just for documentation of my findings: 
- The docker image of the linter was too large. I disabled the copying, since the caching did not work anyways. Now it works.
- The Unit tests fail due to a segmentation fault in metadata_info. I'm looking in to this now.

PR #278, opened 2024-02-05
feat: display database info after loading new database state

Taepper:
For easier information on the log, whether the database was loaded as expected


JonasKellerer (approved):
LGTM

PR #281, opened 2024-02-07
fix: Make C++ flags in CMake compatible for MacOS

Taepper:
Currently I need to take out some flags when building for Mac locally, this adds it to the `CMakeLists.txt` and is triggered only for Mac


JonasKellerer (approved):
LGTM. (But I can not test, since I dont have a mac :) )

PR #283, opened 2024-02-07
README: fix path

pflanze:
A small fix; `./.run` would be more consistent but I figured too many dots are just confusing, and the `./` is unnecessary for paths that are not executables.


fengelniederhammer (approved):
Thanks, looks good :+1:

PR #284, opened 2024-02-07
Cj build with conan

pflanze:
While trying to understand how the build process works, I found this duplication and thought that maybe it will be easier to digest that way. You could also just pick the first of the two commits of course if you prefer that (or feel free to ignore them anyway).


fengelniederhammer (approved):
LGTM

PR #285, opened 2024-02-07
Revert runtime compression

Taepper:
mutations endpoint was too slow for gisaid data


JonasKellerer:
I trued using it with the dataset we already processed on the testserver. However, it did not finish loading even after one hour. Should I have preprocessed the data again with these changes?

JonasKellerer (approved):
LGTM.

Loading of the data works. However it is a bit slow (2h for the covid dataset)

Taepper:
2 hours makes this functionality in its current form not usable. Instead I remove the deletions of bitmaps entirely from the preprocessing for now. It is nice that we can now have multiple positional compression formats available, but this should be planned more clearly

PR #286, opened 2024-02-08
Do not delete bitmaps for now

Taepper:
For now, this will increase preprocessing size. But only after the current maximum spike in memory usage


JonasKellerer (approved):
LGTM

PR #288, opened 2024-02-12
#248 Maybe for Amino Acid

### LINKED ISSUE #248: Add MAYBE for AminoAcids
Check or add that `HasAAMutation` and `AASymbolEquals` uses the `AmbiguityMode`.

Check or add tests that both work with MAYBE queries.

Taepper:
feat: generalizing symbol and mutation filters. Clear handling of ambiguous symbols

resolves #248


src/silo/query_engine/filter_expressions/has_mutation.cpp
```diff
@@ -22,63 +22,73 @@ class DatabasePartition;
...
-         nuc_sequence_name_or_default, position, ref_symbol
-      ));
-      return expression->compile(database, database_partition, NONE);
+   std::vector<typename SymbolType::Symbol> symbols =
+      std::vector(SymbolType::SYMBOLS.begin(), SymbolType::SYMBOLS.end());
+   if (mode == AmbiguityMode::UPPER_BOUND) {
+      // We can only be sure, that the symbol did not mutate, if the ref_symbol is at that position
+      symbols.erase(std::remove(symbols.begin(), symbols.end(), ref_symbol), symbols.end());
+   } else {
+      // Remove all symbols that could match the searched base
```
JonasKellerer, line 70: Do we need this comment?

src/silo/query_engine/filter_expressions/symbol_equals.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include "silo/query_engine/filter_expressions/negation.h"
+#include "silo/query_engine/filter_expressions/or.h"
+#include "silo/query_engine/operators/bitmap_selection.h"
+#include "silo/query_engine/operators/complement.h"
+#include "silo/query_engine/operators/index_scan.h"
+#include "silo/query_engine/operators/operator.h"
+#include "silo/query_engine/query_parse_exception.h"
+#include "silo/storage/database_partition.h"
+
+using silo::Nucleotide;
```
JonasKellerer, line 24: This is only used in line 188 and 190. I think we can remove it here.

src/silo/query_engine/filter_expressions/symbol_equals.cpp
```diff
@@ -0,0 +1,198 @@
...
+)
+    : sequence_name(std::move(sequence_name)),
+      position(position),
+      value(value) {}
+
+template <typename SymbolType>
+std::string SymbolEquals<SymbolType>::toString(const silo::Database& /*database*/) const {
+   const std::string nuc_sequence_name_prefix = sequence_name ? sequence_name.value() + ":" : "";
+   const char symbol_char = value.has_value() ? SymbolType::symbolToChar(*value) : '.';
+   return nuc_sequence_name_prefix + std::to_string(position + 1) + std::to_string(symbol_char);
```
JonasKellerer, line 42: is the fmt slower here, or can we use it to make this more readable?

JonasKellerer (approved):
LGTM.

It feels like we are missing some tests here. Especially to cover the logic of maybe in conjunction with symbol_equals.

src/silo/query_engine/filter_expressions/has_mutation.cpp
Taepper: I like it, as it is not obvious, and the function name approach would be kind of weird at this point in my opinion

PR #290, opened 2024-02-15
Fix inclusive date between for unsorted date column

fengelniederhammer:
The current behavior is:
* filters on an __unsorted__ date column __exclude__ the upper bound
* filters on a __sorted__ date column __include__ the upper bound

Expected behavior: Upper and lower bound should be inclusive.

Also this adds a unit test framework for testing queries directly in C++ code instead of relying on e2e tests for testing queries.


.gitignore
```diff
@@ -64,4 +64,5 @@ conanprofile
```
JonasKellerer, line 64: Reminder: Squash before merge

src/silo/test/date_between.test.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {{"primaryKey", "id"},
+     {"sorted_date", SORTED_DATE_VALUE},
+     {"unsorted_date", UNSORTED_DATE_VALUE}}},
+   {"alignedNucleotideSequences", {{"segment1", nullptr}}},
+   {"unalignedNucleotideSequences", {{"segment1", nullptr}}},
+   {"alignedAminoAcidSequences", {{"gene1", nullptr}}}
+};
+
+const auto DATABASE_CONFIG = DatabaseConfig{
+   "TODO",
```
JonasKellerer, line 27: Should this still be TODO?

JonasKellerer (approved):
LGTM

Maybe we can add tickets/epics to move the query tests out of the e2e tests into this new framework

PR #294, opened 2024-02-21
feat: reintroduce randomize for all query actions

Taepper:


src/silo/query_engine/actions/tuple.test.cpp
fengelniederhammer, line 1: Let's have a unit test with a seed?

src/silo/query_engine/actions/action.cpp
```diff
@@ -140,6 +150,45 @@ void from_json(const nlohmann::json& json, OrderByField& field) {
...
+   CHECK_SILO_QUERY(
+      !json.contains("offset") || json["offset"].is_number_unsigned(),
+      "If the action contains an offset, it must be a non-negative number"
+   )
+   return json.contains("offset") ? std::optional<uint32_t>(json["offset"].get<uint32_t>())
+                                  : std::nullopt;
+}
+
+std::optional<uint32_t> parseRandomizeSeed(const nlohmann::json& json) {
+   if (json.contains("randomize")) {
```
fengelniederhammer, line 172: This is a nice case where early returns improve readability.

```suggestion
   if (!json.contains("randomize")) {
      return std::nullopt;
   }
```

src/silo/test/randomize.test.cpp
```diff
@@ -0,0 +1,85 @@
...
+const QueryTestScenario AGGREGATE = {
+   "aggregateRandomize",
+   {{"action",
+     {{"type", "Aggregated"}, {"groupByFields", {"key"}}, {"randomize", {{"seed", 12321}}}}},
+    {"filterExpression", {{"type", "True"}}}},
+   {{{"count", 1}, {"key", "id3"}},
+    {{"count", 1}, {"key", "id1"}},
+    {{"count", 1}, {"key", "id4"}},
+    {{"count", 1}, {"key", "id5"}},
+    {{"count", 1}, {"key", "id2"}}}
```
fengelniederhammer, line 78: I don't like the formatting, this is almost impossible to read - I wonder whether it might be better to write plain JSON instead of nlohmann::json objects? Or can we fix the formatter to format this properly?

(Not an issue of this PR, just general thoughts)

src/silo/test/randomize.test.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "silo/test/query_fixture.test.h"
+
+using silo::ReferenceGenomes;
+using silo::config::DatabaseConfig;
+using silo::config::ValueType;
+using silo::test::QueryTestData;
+using silo::test::QueryTestScenario;
+
+const std::vector<nlohmann::json> DATA = {
+   {{"metadata", {{"key", "id1"}, {"col", "A"}}},
```
fengelniederhammer, line 14: Do we need "col"? If not, I think it's better to keep the datasets as minimal as possible.

```suggestion
   {{"metadata", {{"key", "id1"}}},
```

src/silo/test/randomize.test.cpp
fengelniederhammer, line 1: I think it would be good to add two e2e tests (`randomize: true` and `randomize: { seed: ... }`), too, for demonstration purposes.

src/silo/query_engine/actions/tuple.cpp
```diff
@@ -363,10 +365,31 @@ std::vector<Tuple::ComparatorField> Tuple::getCompareFields(
...
       actions::Tuple::getCompareFields(columns_metadata, order_by_fields);
+   if (randomize_seed) {
+      const size_t seed = *randomize_seed;
+      return [tuple_field_comparators, seed](const Tuple& tuple1, const Tuple& tuple2) {
+         if (tuple1.compareLess(tuple2, tuple_field_comparators)) {
+            return true;
+         }
+         if (tuple2.compareLess(tuple1, tuple_field_comparators)) {
+            return false;
+         }
```
fengelniederhammer, line 381: To me, it's not quite clear yet how randomize is supposed to interact with "order by"? Do they overwrite each other? What happens if both are specified? Let's add a test that shows the behavior?

Taepper: Randomize will function only as a tiebreaker. Will add a unit test

src/silo/query_engine/actions/tuple.test.cpp
Taepper: But the result would not be deterministic

src/silo/test/randomize.test.cpp
Taepper: For actions, filters, and result, or only results in particular?

src/silo/test/randomize.test.cpp
Taepper: Added it for a sort test, but forgot that implementation hehe

src/silo/test/randomize.test.cpp
fengelniederhammer: For all "jsons" in general.

fengelniederhammer (approved):
LGTM :+1:

PR #301, opened 2024-02-26
feat: make threads and max queued http connections available through optional parameter

JonasKellerer:


Taepper:
Are the queued http connections with 64 maybe a bit too pessimistic? What would be the cost of having a higher limit here? (I would think that queued requests are basically free?)

JonasKellerer:
I am currently testing how many we would need, and if our system can handle large numbers. And what would be the cost of having a high limit here.
This PR is here only, so I have the argument to pass on.
The default value 64 is the default value from POCO. On the server I am testing 10k.

Taepper:
These were also just random thoughts that might be helpful if you hadnt considered them, but looks sensible so far!

Issue #303, opened 2024-02-27
Implement basic request tracing capabilities


fengelniederhammer:
Related to #298, we need to be able to trace requests in the logs. We should at least be able to correlate incoming request and outgoing response. 

Idea: Implement the `X-Request-Id` header. 
* If it is set, SILO will take the value.
* If no value is set, SILO will generate a value
* The response should contain the request id
* At least one log statement should log the request ID along with some indication on the request itself
* At least one log statement shoud log the request ID along with some indication on the response.

PR #306, opened 2024-02-27
refactor: don't use an exception when an `if` does the job

fengelniederhammer:
related to #303 

## PR Checklist


Taepper (approved):
Thanks!

PR #341, opened 2024-03-05
fix: nucleotide symbol equals with dot

JonasKellerer:
From https://github.com/GenSpectrum/cov-spectrum-website/issues/938

### Summary
`'.'` should count if the symbol is equal to the reference genome. This PR should fix that.

## PR Checklist

### DIFF
--- a/src/silo/common/nucleotide_symbols.test.cpp
+++ b/src/silo/common/nucleotide_symbols.test.cpp
@@ -7,7 +7,7 @@ TEST(NucleotideSymbol, enumShouldHaveSameLengthAsArrayOfSymbols) {
 }
 
 TEST(NucleotideSymbol, conversionFromCharacter) {
-   EXPECT_EQ(silo::Nucleotide::charToSymbol('.'), silo::Nucleotide::Symbol::GAP);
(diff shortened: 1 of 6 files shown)


src/silo/common/nucleotide_symbols.test.cpp
```diff
@@ -7,7 +7,7 @@ TEST(NucleotideSymbol, enumShouldHaveSameLengthAsArrayOfSymbols) {
 }
 
 TEST(NucleotideSymbol, conversionFromCharacter) {
-   EXPECT_EQ(silo::Nucleotide::charToSymbol('.'), silo::Nucleotide::Symbol::GAP);
```
corneliusroemer, line 10: Oh I see, it worked as the tests suggested, just the test was wrong 😄

JonasKellerer:
I will now also add that the amino acid symbols are treated the same. (They give the correct result, but not explicitly by design.)

corneliusroemer:
@JonasKellerer great yes, amino acid syntax should be the same, `.` equal to reference.

JonasKellerer:
There is now a new ticket #342 to revisit `charToSymbol` so it does not return the same result for an invalid char and for `'.'`.

fengelniederhammer (approved):
LGTM :+1:

Issue #302, opened 2024-02-26
Multi-architecture/ARM Docker support


chaoran-chen:
It would be great if SILO's Docker image can run on ARM.

PR #344, opened 2024-03-06
fix: add bash dependency which is required by conan build of pkgconf and is not installed on alpine by default

Taepper:
Work towards #302

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -20,6 +20,7 @@ Checks: >
   -abseil-string-find-str-contains,
   -bugprone-easily-swappable-parameters,
   -bugprone-unchecked-optional-access,
+  -bugprone-unused-return-value,
   -readability-magic-numbers
 #  TODO(someone): clean up misc-non-private-member-variables-in-classes and add option back in
 # Not using google-readability-avoid-underscore-in-googletest-name because it also fails for test_name
--- a/src/silo/config/config_repository.test.cpp
+++ b/src/silo/config/config_repository.test.cpp
@@ -30,18 +30,20 @@ ConfigReaderMock mockConfigReader(const DatabaseConfig& config) {
 
 TEST(ConfigRepository, shouldReadConfigWithoutErrors) {
    const auto config_reader_mock = mockConfigReader(
-      {"main",
-       {
-          "testInstanceName",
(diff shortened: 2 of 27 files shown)


src/silo/config/config_repository.test.cpp
```diff
@@ -30,33 +30,37 @@ ConfigReaderMock mockConfigReader(const DatabaseConfig& config) {
...
-          },
-          "testPrimaryKey",
-       }}
+             .instance_name = "testInstanceName",
+             .metadata =
+                {
+                   {.name = "notPrimaryKey", .type = ValueType::STRING},
+                },
+             .primary_key = "testPrimaryKey",
+          }}
```
fengelniederhammer, line 63: Oh nice, I didn't know that C++ has named arguments. This looks way better than before.

.clang-tidy
```diff
@@ -20,6 +20,7 @@ Checks: >
   -abseil-string-find-str-contains,
   -bugprone-easily-swappable-parameters,
   -bugprone-unchecked-optional-access,
+  -bugprone-unused-return-value,
```
fengelniederhammer, line 23: Are we sure that we want to disable this check? Actually, it sounds useful, but I don't know how useful it actually is. How many methods return useless return values?

fengelniederhammer (approved):

src/silo/config/config_repository.test.cpp
Taepper: A newer cpp20 feature: https://pdimov.github.io/blog/2020/09/07/named-parameters-in-c20/

.clang-tidy
Taepper: Yes it is, the problem is that assignment operators lead to false positives (they didn't before)

Taepper: One would need to rewrite
`action = json.get<std::unique_ptr<Mutations<Nucleotide>>>();`
to
`(void) (action = json.get<std::unique_ptr<Mutations<Nucleotide>>>());`
to avoid this error when not using an assignment further...

Taepper: I add a description

fengelniederhammer: Ok, the assignment thing is ugly. Then let's have it disabled :+1:

PR #353, opened 2024-03-19
fix: specify namespace fmt in calls to format_to

Taepper:
This broke compatibility with a late 2023 Ubuntu (23.10).


JonasKellerer (approved):
LGTM

PR #357, opened 2024-03-23
feat: return mutation as object that does not need to be reparsed

Taepper:
See https://github.com/loculus-project/loculus/pull/1446

### DIFF
--- a/endToEndTests/test/queries/aaMutDistribution.json
+++ b/endToEndTests/test/queries/aaMutDistribution.json
@@ -14,156 +14,234 @@
     {
       "count": 37,
       "mutation": "T19R",
+      "mutationFrom": "T",
+      "mutationTo": "R",
+      "position": 19,
       "proportion": 0.38144329896907214,
       "sequenceName": "S"
     },
     {
       "count": 37,
       "mutation": "G142D",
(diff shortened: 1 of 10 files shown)


fengelniederhammer:
I was about to bring up that discussion anyway, I really like the change :+1: 

In LAPIS we will most probably keep the old `mutation` field for backwards compatibility and convenience. Do we want to do the same here? Or do we rather want to strip the redundant data from the SILO responses?

Taepper:
That is a good point, just returning both might be viable. What do you think @chaoran-chen. Keeping both fields only in LAPIS or also in SILO?

fengelniederhammer:
(Also just to keep in mind: this PR should not be merged without preparation in LAPIS if it still contains breaking changes.)

Taepper:
Similar to https://github.com/GenSpectrum/LAPIS/issues/710 it might be better to include both in SILO, then we can order by all fields that are also returned by LAPIS

Taepper:
Should be ready to merge now, as it is non-breaking for LAPIS

Taepper:
Some QOL changes introduced to query.test.js, where the actual result is directly written to a file that can be easier compared and copied

PR #358, opened 2024-03-25
feat: improve validation error message of some actions on orderByFields

fengelniederhammer:
related to https://github.com/GenSpectrum/LAPIS/issues/710

### Summary

Actually one would need to do the same for `Aggregated` and `Details`, too. But this is supposed to be a quick fix and there we have a dynamic list that wasn't that easy to get into an error message.

## PR Checklist


Taepper (approved):
LGTM

PR #359, opened 2024-03-25
Testing environment improvements

pflanze:
- Ignore directories generated by the preprocessing for tests, and running the test runner

- Make silo_test check that it's running in the project root. While one can deduce the reasons for tests failing by looking at exceptions and drawing the right conclusions, this can easily be overlooked and other people may assume the test has to run from the subdir and run into the same as me.

### DIFF
--- a/src/main.test.cpp
+++ b/src/main.test.cpp
@@ -2,10 +2,19 @@
 #include <gtest/gtest.h>
 #include <spdlog/sinks/null_sink.h>
 #include <spdlog/spdlog.h>
+#include <filesystem>
 
 #include "silo/common/log.h"
 
 int main(int argc, char* argv[]) {
+   try {
+      if (!std::filesystem::exists("testBaseData/exampleDataset")) {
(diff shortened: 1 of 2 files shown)


src/main.test.cpp
```diff
@@ -2,12 +2,21 @@
...
 
 #include "silo/common/log.h"
 
 int main(int argc, char* argv[]) {
+   try {
+      if (!std::filesystem::exists("testBaseData/exampleDataset")) {
+         throw std::runtime_error("must be run in root of repository");
+      }
+   } catch (std::exception& e) {
+      std::cerr << e.what() << std::endl;
```
Taepper, line 15: Maybe turn this into a log message:
`SPDLOG_ERROR(...)`

fengelniederhammer: Why do we need this try-catch block anyway? What's wrong with simply letting the exception through?

Taepper (approved):

PR #360, opened 2024-03-25
Small fixes

pflanze:
Various small issues--cosmetic warnings and safety against code changes.

### DIFF
--- a/src/silo/storage/column_group.cpp
+++ b/src/silo/storage/column_group.cpp
@@ -20,6 +20,7 @@ class Nucleotide;
 }  // namespace silo
 
 namespace silo::storage {
+using silo::config::ColumnType;
 
 uint32_t ColumnPartitionGroup::fill(
    duckdb::Connection& connection,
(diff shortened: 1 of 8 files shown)


pflanze:
Sorry, needs clang-format. Which version is the correct one again?

Taepper:
Our CI runs clang-tidy-19

src/silo/storage/column_group.cpp
```diff
@@ -206,6 +206,8 @@ ColumnPartitionGroup ColumnPartitionGroup::getSubgroup(
          result.nuc_insertion_columns.insert({item.name, nuc_insertion_columns.at(item.name)});
       } else if (item.type == silo::config::ColumnType::AA_INSERTION) {
          result.aa_insertion_columns.insert({item.name, aa_insertion_columns.at(item.name)});
+      } else {
+          throw std::runtime_error("missing case");
```
fengelniederhammer, line 210: Let's throw a more specific error message what was missing?

pflanze: This case should never happen, if it does then it's a bug, no? Actually, let me check if that can't be turned into a switch statement, then there will be compile time checking and I can change the throw into an abort.

pflanze: To expand on that / summarize my rationale with regards to using abort and the usage of a closure:

- It's best to have exhaustive matching checks; those are provided
  when using `switch`, but not when using if-else chains.

- Even if the compiler can show whether the case coverage in a switch
  statement is exhaustive, it warns about e.g. leaving a non-void
  function without a return statement (when falling off the
  `switch`). Also, maybe in some situation / for some people the
  compilers won't do compile-time checking, hence a runtime fallback
  is good. But, since it really should never happen, using an
  exception seems overkill: it is more verbose to write and read, one
  then cannot use `noexcept` (well, one can, the exception is then
  turned into an abort anyway), and will most likely cost some more
  space in the binary. And if the binary aborts (again, will never
  happen as long as we build the way we currently do and heed the
  warnings), attaching it to a debugger will clear up where it
  happens.

- Putting a `default` case into the `switch` would defeat the
  exhaustiveness check, hence have to put the `abort` after it; hence
  can't use `break` as that would lead directly to the `abort` in the
  normal flow. `return` is fine for cases where the control flow in
  the current function ends here anyway; in other cases, would have to
  use a label and `goto`, or, a local function (closure) to allow to
  `return` from the case branches. AFAIK compilers will completely
  optimize away a closure like this thus there should be no difference
  in the generated code.

After some IRC discussion this is what I took as the best idiom for
doing exhaustive matching in today's C++ (some say one should never
use abort, but I feel that given that it's normally compiler-proven to
never execute it's fine--it may not even be compiled into the
code). I'll continue to watch out if there's anything better.

src/silo/storage/column_group.cpp
```diff
@@ -190,23 +194,39 @@ ColumnPartitionGroup ColumnPartitionGroup::getSubgroup(
...
+            case ColumnType::NUC_INSERTION:
+               result.nuc_insertion_columns.insert({item.name, nuc_insertion_columns.at(item.name)}
+               );
+               return;
+            case ColumnType::AA_INSERTION:
+               result.aa_insertion_columns.insert({item.name, aa_insertion_columns.at(item.name)});
+               return;
+         }
+         abort();
+      })();
```
Taepper, line 229: Why was this lambda necessary?

Taepper (approved):

src/silo/storage/column_group.cpp
pflanze: (For those reading later) In addition to the comments further above: lambda immediately called is a way (quite commonly used both in C++ and other languages) to enable use of `return` with a specified return boundary.

PR #374, opened 2024-03-27
Manual image caching

Taepper:
faster builds by copying [@corneliusroemer](https://github.com/corneliusroemer) image caching for our dependency images, which rarely change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,14 +3,14 @@ name: LAPIS-SILO
 on:
   push:
 
-
 concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
 env:
-  DOCKER_IMAGE_NAME: ghcr.io/GenSpectrum/LAPIS-SILO
-  DOCKER_TEST_IMAGE_NAME: silo/unittests
+  DOCKER_DEPENDENCY_IMAGE_NAME: ghcr.io/genspectrum/lapis-silo-dependencies
+  DOCKER_LINTER_DEPENDENCY_IMAGE_NAME: ghcr.io/genspectrum/lapis-silo-linter-dependencies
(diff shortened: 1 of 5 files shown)


Taepper:
an example of the working cache:
https://github.com/GenSpectrum/LAPIS-SILO/actions/runs/8463811779

corneliusroemer:
Nice! Would be great to quantify the speed ups :D

How long did it take before and now when there are cache hits for everything?

I didn't see the arm part here, do you use separate docker files for that?

corneliusroemer:
How come the linter takes so long? Is there so much work done?

Also, I noticed you often compile with 4 parallel threads. GitHub actions runners only have 2. It might improve performance to reduce to 2 in CI

JonasKellerer (approved):
LGTM

The ci did not run the unit test image. (from the link you posted) Was that aborted by you?

corneliusroemer:
> The ci did not run the unit test image. (from the link you posted) Was that aborted by you?

@JonasKellerer it did, or is that something else?

image

https://github.com/GenSpectrum/LAPIS-SILO/actions/runs/8463828054/job/23187878018

JonasKellerer:
grafik from Alex comment. There the step of the unit image did not run. But maybe thats an older run? But if it did run all is fine :)

Taepper:
I cancelled it after I saw that it hit the cache in the necessary job. I then just squash and rebased the commit for the PR. Then it rebuilt both dependency images as expected, as duckdb->0.10.1 dependency change was already on the changed main

Taepper:
@corneliusroemer the quantification is not as straight forward, as the cache was hit also before _but only sometimes_

Mostly the first dependency jobs, which take 20 mins each (although in parallel) should now be saved, if no dependency were changed. In the attached screenshot the upper pipeline would be shortened by 21m2s, and the lower one hit the cache by itself by chance

image

PR #383, opened 2024-04-03
fix error handling

pflanze:
This fixes error reporting for some kinds of invalid queries.

We ran into this while working on the boolean columns.

Does this warrant tests (I didn't really save the queries we made, sadly, but may be able to reconstruct them)?

### DIFF
--- a/src/silo/query_engine/filter_expressions/string_equals.cpp
+++ b/src/silo/query_engine/filter_expressions/string_equals.cpp
@@ -1,4 +1,5 @@
 #include "silo/query_engine/filter_expressions/string_equals.h"
+#include <fmt/format.h>
 
 #include <optional>
 #include <utility>
@@ -35,6 +36,12 @@ std::unique_ptr<silo::query_engine::operators::Operator> StringEquals::compile(
    const silo::DatabasePartition& database_partition,
    Expression::AmbiguityMode /*mode*/
 ) const {
+   CHECK_SILO_QUERY(
+      database_partition.columns.string_columns.contains(column) ||
+         database_partition.columns.indexed_string_columns.contains(column),
+      fmt::format("the database does not contain the column '{}'", column)
+   );
+
    if (database_partition.columns.indexed_string_columns.contains(column)) {
       const auto& string_column = database_partition.columns.indexed_string_columns.at(column);
       const auto bitmap = string_column.filter(value);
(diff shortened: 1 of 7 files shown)


pflanze:
Sorry, will check the failing e2e tests later. I only ran the integration tests, need to include e2e in my script.

src/silo/query_engine/filter_expressions/string_equals.cpp
```diff
@@ -46,21 +53,17 @@ std::unique_ptr<silo::query_engine::operators::Operator> StringEquals::compile(
...
-      if (embedded_string.has_value()) {
-         return std::make_unique<operators::Selection>(
-            std::make_unique<operators::CompareToValueSelection<common::SiloString>>(
-               string_column.getValues(), operators::Comparator::EQUALS, embedded_string.value()
-            ),
-            database_partition.sequence_count
-         );
-      }
-      return std::make_unique<operators::Empty>(database_partition.sequence_count);
+   // database_partition.columns.string_columns.contains(column)
```
fengelniederhammer, line 56: Let's delete this?

fengelniederhammer (changes requested):
I think the `DateBetween` and `IntBetween` and filters could use the same check. 

```
{
  "action": {
    "limit": 100,
    "type": "Aggregated"
  },
  "filterExpression": {
    "children": [
      {
        "column": "division",
        "from": "Aargau",
        "to": "Aargau",
        "type": "DateBetween"
      }
    ],
    "type": "And"
  }
}
``` 
fails with 
```
{
  "error": "Internal Server Error",
  "message": "map::at"
}
``` 

Otherwise, this is a nice improvement :+1: 

In the long run, it would be nice if we could also test error messages in the unit tests. IIRC, that's currently not possible.

src/silo/query_engine/filter_expressions/string_equals.cpp
Taepper: I would like to turn this into:
`database_partition.columns.string_columns.contains(column) `

pflanze:
I've added @fengelniederhammer 's test example to the commit message, so that it's easy to find if in the future someone wants to write actual tests. I think it's ready to re-review.

pflanze:
I've now also added single quotes around the column names in the error messages.

Taepper (approved):

PR #384, opened 2024-04-08
Boolean columns

### LINKED ISSUE #219: Implement column type boolean
Pathoplexus needs the field `isLatestVersion` which is a boolean. SILO currently does not support columns of type boolean.

pflanze:
resolves  #219


pflanze:
Looks like the e2e tests fail with hash based ordering issues. How to solve?

And the linter fails with cognitive complexity: fair enough, it does strike me that the code is missing some abstractions for these things in general. But this will need a bit more study on my part and it's not the right time to do this now. Alternatively could introduce a macro or template here as a hack but I'm not sure the linter will be happier about it?

src/silo/storage/database_partition.cpp
```diff
@@ -124,6 +124,13 @@ void DatabasePartition::validateMetadataColumns() const {
          );
       }
    }
+   for (const auto& col : columns.bool_columns) {
```
fengelniederhammer, line 127: Regarding cognitive complexity: 
* Either we ignore it here (we unfortunately also did this in many other places
* Or (my preferred way) we extract methods for every `for` in this method (seems easy enough on a quick glimpse).

include/silo/storage/column/bool_column.h
fengelniederhammer, line 1: I wonder why we don't make bool columns always indexed. There are only 3 possible values, so we only need to store 3 bitmaps. Or did you actively decide against indexed bool columns?

fengelniederhammer (commented):
Looks good so far :+1: 
* We need to update the docs, though not now. I created https://github.com/GenSpectrum/LAPIS/issues/734.
* Should we add unit tests that test the `BooleanEquals` filter?

fengelniederhammer:
And please squash your commits before merging such that the message conforms to https://www.conventionalcommits.org/.

pflanze:
> Looks like the e2e tests fail with hash based ordering issues. 

That wasn't actually the case; it didn't get the new column in the test results because the database_config.yaml for the ndjson tests wasn't updated (which I didn't notice because my test runner didn't run the ndjson tests; I'll make a PR with a proper test runner in the future).

I have also pushed a commit that replaces exampleDatasetAsNdjson/database_config.yaml with a symlink that points to exampleDataset/database_config.yaml, since the two files are identical. In the future if the files ever diverge, the symlink could be replaced with the changed content again.

pflanze:
> Should we add unit tests that test the `BooleanEquals` filter?

There are 3 tests that do so?: endToEndTests/test/queries/booleanEquals*.json

fengelniederhammer:
> I guess you really mean unit tests?

Yes, I mean unit tests in https://github.com/GenSpectrum/LAPIS-SILO/tree/003c7ca53cb7915493e1a0871519a397dde77de6/src/silo/test

include/silo/storage/column/bool_column.h
fengelniederhammer: I created #385 to do this later.

pflanze:
Changed the commit messages to properly follow conventional commit spec, and only using issue number in the last commit to avoid spamming the linked GitHub issue.

This concludes the work, if there are no objections I will merge.

pflanze:
I don't know what failed in the CI. It appears to be something building the docker container? The tests run fine for me locally and the content didn't change from before the last force push.

PR #391, opened 2024-04-22
add .dir-locals.el for project-specific Emacs settings

pflanze:
### Summary

This adds Emacs configuration specific to this project. Emacs automatically reads this file when editing source files in this repository and indentation settings are then automatically correct.

I don't think this needs documentation, it's used automatically for Emacs users (it asks to confirm using the file on first use), and other users don't care.

## PR Checklist

- [ ] ~~All necessary documentation has been adapted or there is an issue to do so.~~
- [ ] ~~The implemented feature is covered by an appropriate test.~~

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,6 @@
+;; The 'nil' configuration applies to all modes.
+((nil . ((indent-tabs-mode . nil)
+         (tab-width . 2) ;; used by what?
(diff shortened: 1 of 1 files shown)


pflanze:
Pushed commit message change for conventional commits.

fengelniederhammer (approved):

PR #399, opened 2024-04-25
Runtime Config

Taepper:
feat: tidying up CLI and file configuration for runtime config. Added option for specifying the port

### Summary

The port on which SILO listens to requests should be able to be specified.

Furthermore, I edited the way the configs are parsed such that there is a default `RuntimeConfig` (default constructor), which has 2 member functions for overwriting its values from 1. the config file and 2. the CLI arguments

## PR Checklist

### DIFF
--- a/include/silo_api/runtime_config.h
+++ b/include/silo_api/runtime_config.h
@@ -3,12 +3,25 @@
 #include <filesystem>
 #include <optional>
 
+#include <Poco/Util/AbstractConfiguration.h>
+
+#include "silo/preprocessing/preprocessing_config.h"
+
 namespace silo_api {
 
-struct RuntimeConfig {
-   std::optional<std::filesystem::path> data_directory;
+static const std::string DATA_DIRECTORY_OPTION = "dataDirectory";
+static const std::string MAX_CONNECTIONS_OPTION = "maxQueuedHttpConnections";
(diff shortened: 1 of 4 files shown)


fengelniederhammer (commented):
Maybe we should also document somewhere (in the README) that the runtime config exists? I just noticed that this is not mentioned there.

Taepper:
> Maybe we should also document somewhere (in the README) that the runtime config exists? I just noticed that this is not mentioned there.

Yes, I will do so! What is with the StartUp Config. Would you have any objections if I merge all the information from that into the RuntimeConfig?

Taepper:
Then we have one config for --api and one config for --preprocessing. I would also change the way the preprocessing config currently works, to be consistent with this. Therefore, we consistently have

`default configuration values < config file < CLI arguments`

Cornelius and Christian both mentioned that environment variables should be added to those lists. Do you object to have that between config and CLI for the final precendences:

`default configuration values < config file < environment variables < CLI arguments`

pflanze:
Commit 70523298b0cbb513f703e567bd9db9281d29f871 works for me, and the code looked good to me. Documentation will be good. I don't think it needs any tests at this time in the project's history since the CI and my test runner will exercise the important options already.

pflanze (approved):
LGTM, and works for SILO_PORT.

PR #402, opened 2024-04-30
fix: floatEquals and floatBetween with null values

JonasKellerer:
### Summary

In https://github.com/GenSpectrum/LAPIS/pull/760 it was found that filtering for null values does not work correctly. This PR targets this issue. The problem was that null is represented as nan in the code. However, equals comparisons with nan always return false. 

## PR Checklist

### DIFF
--- a/src/silo/query_engine/operators/selection.cpp
+++ b/src/silo/query_engine/operators/selection.cpp
@@ -1,6 +1,7 @@
 #include "silo/query_engine/operators/selection.h"
 
 #include <array>
+#include <cmath>
 #include <compare>
 #include <iomanip>
 #include <iterator>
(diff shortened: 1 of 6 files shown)


src/silo/test/float_equals.test.cpp
fengelniederhammer, line 1: actually it's not only `float_equals`, but also `float_between` :)

src/silo/query_engine/operators/selection.cpp
```diff
@@ -131,8 +132,14 @@ bool CompareToValueSelection<T>::match(uint32_t row_id) const {
    assert(column.size() > row_id);
    switch (comparator) {
       case Comparator::EQUALS:
+         if (std::isnan(value)) {
+            return std::isnan(column[row_id]);
```
fengelniederhammer, line 136: If I rebase this to main, this doesn't compile:
```
 error: no matching function for call to ‘isnan(const value_type&)’
  145 |             return !std::isnan(column[row_id]);

[...]

/usr/include/c++/11/cmath:611:3: note: candidate: ‘constexpr bool std::isnan(float)’
  611 |   isnan(float __x)
      |   ^~~~~
/usr/include/c++/11/cmath:611:15: note:   no known conversion for argument 1 from ‘const value_type’ {aka ‘const silo::common::OptionalBool’} to ‘float’
  611 |   isnan(float __x)
      |         ~~~~~~^~~
/usr/include/c++/11/cmath:619:3: note: candidate: ‘constexpr bool std::isnan(double)’
  619 |   isnan(double __x)
      |   ^~~~~
/usr/include/c++/11/cmath:619:16: note:   no known conversion for argument 1 from ‘const value_type’ {aka ‘const silo::common::OptionalBool’} to ‘double’
  619 |   isnan(double __x)
      |         ~~~~~~~^~~
/usr/include/c++/11/cmath:624:3: note: candidate: ‘constexpr bool std::isnan(long double)’
  624 |   isnan(long double __x)
      |   ^~~~~
/usr/include/c++/11/cmath:624:21: note:   no known conversion for argument 1 from ‘const value_type’ {aka ‘const silo::common::OptionalBool’} to ‘long double’
  624 |   isnan(long double __x)


```

fengelniederhammer (changes requested):

src/silo/test/int_equals.test.cpp
fengelniederhammer, line 1: Let's also rename this file?

src/silo/query_engine/operators/selection.cpp
fengelniederhammer, line 1: Reminder: squash before merge

fengelniederhammer (approved):
LGTM

PR #403, opened 2024-05-01
fix: pango lineage filter with null values

JonasKellerer:
### Summary

In https://github.com/GenSpectrum/LAPIS/pull/760 it was found that filtering for null values does not work correctly. This PR targets this issue. The problem was that null was not allowed for pangolineage filters.

## PR Checklist

### DIFF
--- a/src/silo/test/pango_lineage_filter.test.cpp
+++ b/src/silo/test/pango_lineage_filter.test.cpp
@@ -1,0 +1,116 @@
+#include <nlohmann/json.hpp>
+
+#include <optional>
+
+#include "silo/test/query_fixture.test.h"
+
+using silo::PangoLineageAliasLookup;
+using silo::ReferenceGenomes;
+using silo::config::DatabaseConfig;
+using silo::config::ValueType;
+using silo::test::QueryTestData;
+using silo::test::QueryTestScenario;
+
+static const std::string SOME_BASE_PANGO_LINEAGE = "BASE.1";
+static const std::string SOME_SUBLINEAGE = "CHILD.1";
+
+nlohmann::json createDataWithPangoLineageValue(const std::string& primaryKey, std::string value) {
+   return {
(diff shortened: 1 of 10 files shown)


src/silo/test/pango_lineage_filter.test.cpp
```diff
@@ -0,0 +1,116 @@
...
+           {.name = "pango_lineage", .type = ValueType::PANGOLINEAGE, .generate_index = true}},
+       .primary_key = "primaryKey"}
+};
+
+const auto REFERENCE_GENOMES = ReferenceGenomes{
+   {{"segment1", "A"}},
+   {{"gene1", "*"}},
+};
+
+const auto ALIAS_LOOKUP = PangoLineageAliasLookup{{{"A", {"P.1"}}}};
```
fengelniederhammer, line 56:
```suggestion
const auto ALIAS_LOOKUP = PangoLineageAliasLookup{{{"A", {"P.1.1.1"}}}};
```
I think there is a certain logic that expects this to have 3 numbers.

src/silo/test/pango_lineage_filter.test.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+using silo::PangoLineageAliasLookup;
+using silo::ReferenceGenomes;
+using silo::config::DatabaseConfig;
+using silo::config::ValueType;
+using silo::test::QueryTestData;
+using silo::test::QueryTestScenario;
+
+static const std::string SOME_BASE_PANGO_LINEAGE = "P.1";
+static const std::string SOME_SUBLINEAGE = "A.1";
```
fengelniederhammer, line 15:
```suggestion
static const std::string SOME_BASE_PANGO_LINEAGE = "BASE.1";
static const std::string SOME_SUBLINEAGE = "CHILD.1";
```

fengelniederhammer (approved):
LGTM

Issue #129, opened 2023-06-15
Validate input data


fengelniederhammer:
When doing SILO preprocessing, validate the input data:
* Pango lineage alias
  * validate that it's a proper tree
* Metadata
  * primary keys are unique
  * validate data types of each field
  * Pango lineages have their alias defined
* sequences
  * primary key exists in metadata, is unique and every metadata primary key exists in sequences.
  * aligned sequences: verify length

JonasKellerer:
"primary key exists in metadata, is unique and every metadata primary key exists in sequences." exists already

fengelniederhammer:
Mind https://github.com/GenSpectrum/LAPIS-SILO/pull/215#discussion_r1413702056 and improve log messages

fengelniederhammer:
Try to reproduce #361 and fix it.

Taepper:
Also: Unit tests that execute the preprocessing in both paths

corneliusroemer:
Motivated by the bug hunt related to https://github.com/loculus-project/loculus/pull/2667

Extra validation ideas:
- run a self test after import, running a bunch of queries/assertions
  - if downloading all sequences, get correct number of sequences
  - if downloading all sequences, don't get duplicate sequence ids
  - ...

Taepper:
After #458 a lot of this should already be implemented.

We will add separate issues to verify and that check that proper error messaged are returned accordingly.

PR #429, opened 2024-05-17
Insertions out of metadata

### LINKED ISSUE #372: Move insertions out of metadata
* There should not be a metadata type for insertions
* In the tsv/fasta preprocessing format: Currently, insertions are provided to SILO in the metadata.tsv file. Think of how else insertions should be provided
* In NDJSON format, insertions can be provided separately from metadata already.
* This solves a couple of issues:
  * currently it's possible to aggregated by insertions
  * https://github.com/GenSpectrum/LAPIS-SILO/pull/354
  * "Workarounds" in the code, where you need to consider metadata types that are not insertions (which is the case quite often)

Taepper:
resolves #372 

### Summary

Insertions are no metadata, but are specific to an aligned sequence. This is now reflected in the code. The database config does no longer contain the metadata, but instead there is an nuc_insertions and aa_insertions required, which give all insertions for nucleotide sequences and amino acid sequences respectively

This blocked the refactoring of the preprocessing, as there was quite some extra logic for considering insertions

## PR Checklist


Taepper:
I will still add an option to the preprocessingConfig to change the location of the insertions files and make it relative to the input_directory

src/silo/preprocessing/sequence_info.cpp
```diff
@@ -79,12 +79,46 @@ std::string SequenceInfo::getAminoAcidSequenceSelect(
...
+            "not contained in the reference sequences.",
+            name_type,
+            name,
+            input_filename.string()
+         ));
+      }
+   }
+   for (const std::string& name : names_to_validate_against) {
+      if (std::find(names_to_validate.begin(), names_to_validate.end(), name) == names_to_validate.end()) {
+         // TODO(#220) handle the cases when segments are left out appropriately
```
fengelniederhammer, line 102: Is this still a TODO? #220 is already done.

fengelniederhammer (approved):
Looks good, overall the code got quite a bit simpler :+1: 

Do you also want to adapt the LAPIS docs?
https://lapis.cov-spectrum.org/gisaid/v2/docs/maintainer-docs/references/preprocessing#metadata-file
This still describes the old mechanism.

Also a unit test would be nice that check that the preprocessing works in both paths (also with more than one insertion for a single sequence), but maybe we can leave that to #129?

fengelniederhammer:
And please add `BREAKING CHANGE: ...` in the commit message footer to make sure that we get a new minor version (corresponding to a major version for pre 1.0)

src/silo/preprocessing/sequence_info.cpp
Taepper: I was also under the impression that this already works, but didn't want to remove the TODO without verifying everything first

fengelniederhammer: Then let's make it a TODO for #129?

Taepper: I would add another ticket that request that keys may be left out
But might turn into a won't fix (/ won't implement), as duckdb validation might not be able to cover left out keys in all cases

fengelniederhammer:
Oh and please don't merge yet. We'll have to check whether this is compatible with LAPIS. Most likely, we'll at least have to adapt the preprocessing configs in the deployments.

Taepper:
> Oh and please don't merge yet. We'll have to check whether this is compatible with LAPIS. Most likely, we'll at least have to adapt the preprocessing configs in the deployments.

At least the way it is designed, it should not be breaking for ndjson inputs. The only thing is that we are more strict on the requirement on the json fields of `nucleotideInsertions` and `aminoAcidInsertions`, which should have already been checked before

fengelniederhammer:
Yes, but I just checked, all the configs still contain the `insertions` metadata type, which needs to be removed.

fengelniederhammer:
I squashed and combined the two breaking change lines in the commit message into a single one, because Release Please seems to ignore all but the last of them.

PR #433, opened 2024-05-17
Empty input

### LINKED ISSUE #244: Start without data
As just discussed in Slack, it should be possible to start SILO properly with a schema but without data. Currently, it might throw errors such as "Error from SILO: Metadata field country not found." when it is started without a serialized state. Such an error message is very confusing.

Taepper:
resolves #244

### Summary

We want to be able to start without data, such that downstream packages do not need extra logic.

In particular the ingest with an empty ndjson (which does not have keys anymore) should not fail. This was challenging, because we never want to read the ndjson files into a table as a whole, but only the non-sequence columns of it, and the sequence columns should be compressed while reading them

## PR Checklist


include/silo/preprocessing/metadata_info.h
```diff
@@ -2,34 +2,37 @@
...
+   static std::vector<std::string> getMetadataFields(
+      const silo::config::DatabaseConfig& database_config
+   );
 
-   std::vector<std::string> getMetadataSelects() const;
+   static std::vector<std::string> getMetadataTypes(
+      const silo::config::DatabaseConfig& database_config
+   );
+
+   static std::string getMetadataStruct(const silo::config::DatabaseConfig& database_config);
```
fengelniederhammer, line 31: According to my IDE, this is unused.

include/silo/preprocessing/metadata_info.h
```diff
@@ -2,34 +2,37 @@
...
       const silo::config::DatabaseConfig& database_config
    );
 
-   std::vector<std::string> getMetadataFields() const;
+   static std::vector<std::string> getMetadataFields(
+      const silo::config::DatabaseConfig& database_config
+   );
 
-   std::vector<std::string> getMetadataSelects() const;
+   static std::vector<std::string> getMetadataTypes(
```
fengelniederhammer, line 27:
```suggestion
   static std::vector<std::string> getMetadataSqlTypes(
```

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -9,137 +9,173 @@
...
    const std::filesystem::path& metadata_file,
    const silo::config::DatabaseConfig& database_config
 ) {
    duckdb::DuckDB duck_db(nullptr);
    duckdb::Connection connection(duck_db);
    // Get the column names (headers) of the table
-   auto result =
-      connection.Query(fmt::format("SELECT * FROM '{}' LIMIT 0", metadata_file.string()));
+   auto result = connection.Query(fmt::format(
+      "SELECT * FROM read_csv_auto('{}', delim = '\t', header = true) LIMIT 0",
```
fengelniederhammer, line 43: What's actually the reason for using `read_csv_auto` instead of putting the plain filename there?

src/silo/preprocessing/metadata_info.cpp
```diff
@@ -9,137 +9,173 @@
...
-   if (result->RowCount() == 0) {
-      throw silo::preprocessing::PreprocessingException(fmt::format(
-         "File {} is empty, which must not be empty at this point", ndjson_file.string()
-      ));
-   }
-   if (result->RowCount() > 1) {
-      throw silo::preprocessing::PreprocessingException(
-         "Internal exception, expected Row Count=1, actual " + std::to_string(result->RowCount())
-      );
+      return;
```
fengelniederhammer, line 109: Is it ok to `return` here instead of throwing?

fengelniederhammer (changes requested):
Actually looks good, but IMO the code needs a little clean up.

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -183,7 +198,7 @@ void Preprocessor::buildPartitioningTableByColumn(const std::string& partition_b
       R"-(
 CREATE OR REPLACE TABLE partition_keys AS
 SELECT row_number() OVER () - 1 AS id, partition_key, count
-FROM (SELECT {} AS partition_key, COUNT(*) AS count
+FROM (SELECT CASE WHEN "{0}" IS NULL THEN 'NULL'::VARCHAR ELSE '_' || "{0}" END AS partition_key, COUNT(*) AS count
```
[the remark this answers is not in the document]
Taepper: I wanted to get rid off the join clause needing a special case for `NULL`, which is quite bug-prone, as it is required multiple times at very different spots in the file.
Therefore, I needed a value to map `NULL` to, for which I chose the string `'NULL'`. To not have a collision with a value in this column that is also the string `NULL`, I prepended all values with a `_` for the join key

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -274,114 +289,155 @@ FROM metadata_table;
...
-      partition_by_where = fmt::format(
-         "WHERE (metadata.\"{0}\" = partition_key_to_partition.partition_key) OR "
-         "(metadata.\"{0}\" IS NULL AND "
-         "partition_key_to_partition.partition_key IS NULL)",
-         database_config.schema.partition_by.value()
-      );
-   } else {
-      partition_by_select = "0 AS partition_id";
-      partition_by_where = "";
+   if (std::filesystem::is_empty(file_name)) {
```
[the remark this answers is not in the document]
Taepper: Huh, interesting, that the linter did not catch this

src/silo/preprocessing/metadata_info.cpp
Taepper: It is more explicit. Also it prevents some errors where the auto-detection cannot infer that there is a header file, or erroneously assumes ',' as delimiter and have all (tab-separated) columns in a single column instead

src/silo/preprocessing/preprocessor.cpp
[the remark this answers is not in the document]
Taepper: I have trouble to come up with a nice name for the function that turns `column` to `CASE WHEN "column" IS NULL THEN 'NULL'::VARCHAR ELSE '_' || "column"`

What do you think of:
`std::string escapeNullFromField(const std::string& field);`
`std::string makeKeyNonNull(const std::string& field);`
`std::string convertFieldToStringIfNull(const std::string& field);`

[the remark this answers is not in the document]
fengelniederhammer: Inspired by `makeKeyNonNull`: `makeNonNullKey`?

src/silo/preprocessing/metadata_info.cpp
Taepper: True. Let's better have the emptiness check outside

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -769,53 +708,54 @@ void Preprocessor::buildAminoAcidSequenceStore(
...
                      .read();
 
-                  aa_sequence_store.finalize();
+                  sequence_store.finalize();
                }
             }
          }
       );
-      SPDLOG_INFO("build - finished amino acid sequence {}", aa_name);
+      SPDLOG_INFO("build - finished nucleotide sequence {}", sequence_name);
```
fengelniederhammer, line 758: We should probably replace here: `nucleotide` -> `SymbolType::SYMBOL_NAME`?

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -38,7 +46,24 @@ Preprocessor::Preprocessor(
...
       preprocessing_db(preprocessing_config.getPreprocessingDatabaseLocation(), reference_genomes),
       reference_genomes_(reference_genomes),
-      alias_lookup_(std::move(alias_lookup)) {}
+      alias_lookup_(std::move(alias_lookup)) {
+   for (const auto& [seq_name, _] : reference_genomes_.raw_nucleotide_sequences) {
+      nuc_sequences.emplace_back(seq_name);
+   }
+   for (const auto& [seq_name, _] : reference_genomes_.raw_aa_sequences) {
+      aa_sequences.emplace_back(seq_name);
+   }
```
fengelniederhammer, line 55:
```suggestion
   nuc_sequences = reference_genomes_.getSequenceNames<Nucleotide>();
   aa_sequences = reference_genomes_.getSequenceNames<AminoAcid>();
```

fengelniederhammer (approved):
LGTM

corneliusroemer:
Can I already test this works? Are there any breaking changes or can I just use the docker image as is?

PR #436, opened 2024-05-21
docs: document how to test the changelog that will be generated

fengelniederhammer:
### Summary

example: `npm run release-please-dry-run -- --token=ghp_WcVX*****aJkLsg --target-branch=changelogTest`

## PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,3 +246,17 @@ To check commit messages of a branch to the commit where it branches off from `m
 ```shell
 npm run commitlint:merge-base
 ```
+
+### Testing The Generated Changelog
+
+To test the generated changelog, run
+
+```shell
+npm run release-please-dry-run -- --token=<GitHub PAT> --target-branch=<name of the upstream branch>
+```
+
+where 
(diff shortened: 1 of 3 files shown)


pflanze:
- Where does the example invocation take the repo url from? (Does it read it from the local Git clone?) It also doesn't list the config file, release-please-config.json.

- Unlike what the commit message suggests, the commit adds not only documentation, but also the installation and running in CI of the tool.

fengelniederhammer:
> * Where does the example invocation take the repo url from? (Does it read it from the local Git clone?) It also doesn't list the config file, release-please-config.json.

All of that is already bundled in `npm run release-please-dry-run`, because it will always be fixed for this repo. The only arguments that a user needs to supply are the GitHub PAT and the branch, since they are dynamic for this command in this context.

>     * Unlike what the commit message suggests, the commit adds not only documentation, but also the installation and running in CI of the tool.

I added it.

pflanze:
Aha got it, again, the entry in package.json is used by the invocation (I'm not a frequent npm user).

pflanze (approved):

PR #438, opened 2024-05-22
refactor: enable YAMLFiles as ConfigSources to have substructs

taepper:
## PR Checklist

~- [] All necessary documentation has been adapted or there is an issue to do so.~


include/silo/config/api_options.h
fengelniederhammer, line 1: Do we need this file?

src/silo/config/api_options.cpp
fengelniederhammer, line 1: Do we need this file?

include/silo_api/request_handler_factory.h
```diff
@@ -16,10 +16,13 @@ namespace silo_api {
 class SiloRequestHandlerFactory : public Poco::Net::HTTPRequestHandlerFactory {
   private:
    silo_api::DatabaseMutex& database;
-   const RuntimeConfig runtime_config;
+   const silo::config::RuntimeConfig runtime_config;
```
fengelniederhammer, line 19: Why are we not `using` the runtime config here?

src/silo/config/util/abstract_config_source.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+std::optional<int32_t> silo::config::AbstractConfigSource::getInt32(const Option& option) const {
+   const auto string_value = getString(option);
+   if (string_value == std::nullopt) {
+      return std::nullopt;
+   }
+   try {
+      return boost::lexical_cast<int32_t>(*string_value);
+   } catch (boost::bad_lexical_cast&) {
+      return std::nullopt;
```
fengelniederhammer, line 20: Or should we rather throw in this case? Spotting those type errors can be quite hard when SILO silently ignores them.

src/silo/config/util/abstract_config_source.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+std::optional<uint32_t> silo::config::AbstractConfigSource::getUInt32(const Option& option) const {
+   const auto string_value = getString(option);
+   if (string_value == std::nullopt) {
+      return std::nullopt;
+   }
+   try {
+      return boost::lexical_cast<uint32_t>(*string_value);
+   } catch (boost::bad_lexical_cast&) {
+      return std::nullopt;
```
fengelniederhammer, line 32: Same here: Throw instead of ignoring it?

src/silo/config/util/yaml_file.cpp
```diff
@@ -0,0 +1,51 @@
...
+   for (const auto& access : option.access_path) {
+      if (!current.IsDefined() || !current.IsMap()) {
+         return false;
+      }
+      current = current[access];
+   }
+   if (!current.IsDefined() || !current.IsMap()) {
+      return false;
+   }
+   return true;
```
fengelniederhammer, line 36:
```suggestion
   return current.IsDefined() && current.IsMap();
```

fengelniederhammer: I'm not even sure whether this is correct. `IsMap` seems strange here. I think a unit test would be good that covers this method and `getString`.

src/silo_api/command_line_arguments.cpp
```diff
@@ -1,26 +1,66 @@
...
+            const char char_in_lower_case =
+               static_cast<char>(std::tolower(static_cast<unsigned char>(character)));
+            current_result += char_in_lower_case;
+         } else {
+            current_result += character;
+         }
+      }
+      result.emplace_back(current_result);
+   }
+   return "SILO_" + boost::join(result, "-");
```
fengelniederhammer, line 25: This is the result of `siloApi --help`:

```
usage: siloApi OPTIONS
SILO - Sequence Indexing engine for Large Order of genomic data

--daemon                                                     Run application 
                                                             as a daemon.
--umask=mask                                                 Set the daemon's 
                                                             umask (octal, 
                                                             e.g. 027).
--pidfile=path                                               Write the process
                                                             ID of the 
                                                             application to 
                                                             given file.
-h, --help                                                   display help 
                                                             information on 
                                                             command line 
                                                             arguments
--preprocessingConfig=PATH                                   path to the 
                                                             preprocessing 
                                                             config file
--databaseConfig=PATH                                        path to the 
                                                             database config 
                                                             file
-dPATH, --SILO_data-directory=PATH                           path to the 
                                                             preprocessed data
--SILO_port=NUMBER                                           port to listen to
                                                             requests
--maxQueuedHttpConnections=NUMBER                            maximum number of
                                                             http connections
--threadsForHttpConnections=NUMBER                           number of threads
                                                             for http 
                                                             connections
-a, --api                                                    Execution mode: 
                                                             start the SILO 
                                                             web interface
-p, --preprocessing                                          Execution mode: 
                                                             trigger the 
                                                             preprocessing 
                                                             pipeline to 
                                                             generate a 
                                                             partitioned 
                                                             dataset that can 
                                                             be read by the 
                                                             database
-tMINUTES, --SILO_estimated-startup-time-in-minutes=MINUTES  Estimated time in
                                                             minutes that the 
                                                             initial loading 
                                                             of the database 
                                                             takes. As long as
                                                             no database is 
                                                             loaded yet, SILO 
                                                             will throw a 503 
                                                             error. This 
                                                             option allows 
                                                             SILO to compute a
                                                             Retry-After 
                                                             header for the 
                                                             503 response. 
``` 

* IMO the `SILO_` prefix looks weird. Should we get rid of it?
* The casing is not consistent (`SILO_estimated-startup-time-in-minutes` vs `maxQueuedHttpConnections`. Depending on which casing style we go for, this will be another breaking change. Actually I prefer to leave it camel case, or are there good arguments to make everything kebap case?

src/silo_api/command_line_arguments.cpp
```diff
@@ -1,26 +1,66 @@
...
 }
 
-int32_t CommandLineArguments::getInt32(const std::string& key) const {
-   return config.getInt(key);
+std::optional<int32_t> CommandLineArguments::getInt32(const Option& option) const {
+   if (hasProperty(option)) {
+      try {
+         return config.getInt(asUnixOptionString(option));
+      } catch (Poco::SyntaxException&) {
+         return std::nullopt;
```
fengelniederhammer, line 51: Same here, I think it would be better if we don't silently ignore this exception.

fengelniederhammer (changes requested):

src/silo_api/command_line_arguments.cpp
Taepper: The SILO prefix is mistakenly added. I am still thinking of a way to have these options nicely displayed automatically. I think that we want to

include/silo_api/request_handler_factory.h
Taepper: As it is a header file, I would prefer not to

src/silo_api/command_line_arguments.cpp
Taepper: (the SILO prefix is only meant for environment variables)

Taepper: Christian mentioned that kebab-case is the defacto UNIX standard

src/silo_api/api.cpp
fengelniederhammer, line 1: `.binding("threadsForHttpConnections"))` and `.binding("maxQueuedHttpConnections"))` also needs to use the `CommandLineArguments::asUnixOptionString`, doesn't it? Otherwise reading the options won't work?

(Although this is another breaking change)

fengelniederhammer (changes requested):
Actually looks good, but let's fix the remaining command line argument names.

src/silo_api/api.cpp
Taepper: Okay, then I would like to not have this breaking (also why I explicitly marked this as refactor:). Therefore I will keep the helper function for UNIX casing, but will remove its usage for now and add an issue instead.

This should be a refactor PR, which only helps to implement the breaking behavior in the future

fengelniederhammer, line 1: Let's also use `silo::config::MAX_CONNECTIONS_OPTION` and `silo::config::PARALLEL_THREADS_OPTION` for the argument definitions here, now it should be possible without breaking changes, isn't it?

fengelniederhammer (approved):

PR #439, opened 2024-05-22
feat: add serializationVersion to SILO output for smoother transitioning to new formats

### LINKED ISSUE #437: Versions for preprocessing output
#429 made apparent that we need some concept of versioning for the output of the SILO preprocessing, as it introduces a breaking change there. An already running instance of SILO will not be able to read the serialized state that is produced by a newer version of SILO.

We decided that the preprocessing output should be versioned.
* The `preprocessing output version` is an integer, starting with 1.
  * Semantic versioning is not feasible, since changes are either breaking or they are fully compatible (forwards and backwards).
  * If no version is found in the preprocessing output, SILO should assume `preprocessing output version 0`.
  * The version is determined by a constant in the code that has to be changed when implementing breaking changes in the preprocessing output
* Implementation suggestion for the API reading new database states:
  * Start with the newest version folder
  * Check data version file
    * if not found, then ignore
  * Check preprocessing output version
    * if incompatible (i.e. `output's version != SILO code's version`), then log a message at `warn` and ignore
  * if the data version file check and the preprocessing output version check was passed, then load that data version.

Taepper:
resolves #437 

## PR Checklist


src/silo_api/database_directory_watcher.test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+TEST(DatabaseDirectoryWatcher, invalidDirectoryNotValid) {
+   ASSERT_EQ(DatabaseDirectoryWatcher::checkValidDataSource("///"), std::nullopt);
+}
+
+TEST(DatabaseDirectoryWatcher, validBackwardsCompatible) {
+   auto under_test =
+      DatabaseDirectoryWatcher::checkValidDataSource("testBaseData/testDataDirectories/1234");
+   ASSERT_TRUE(under_test.has_value());
+   ASSERT_TRUE(under_test.value().isCompatibleVersion());
```
fengelniederhammer, line 19: This will change soon (and thus break the test). Let's check for `== 0` instead?

fengelniederhammer: (Same for the test below)

src/silo_api/database_directory_watcher.test.cpp
fengelniederhammer, line 1: Let's maybe add assertions that the `timestamp` value is as expected?

testBaseData/testDataDirectories/3123/data_version.silo
fengelniederhammer, line 1: Let's maybe rename the directory `3123` to `somethingInvalidInHere`?

testBaseData/testDataDirectories/3124/data_version.silo
fengelniederhammer, line 1: Let's maybe rename the directory `3124` to `notEvenYaml`?

src/silo_api/database_directory_watcher.cpp
```diff
@@ -92,26 +84,28 @@ std::optional<std::pair<std::filesystem::path, silo::DataVersion>> getMostRecent
...
+            "Selected newest data which is compatible ({}) in path {} for ingestion",
+            entry.second.toString(),
+            entry.first.string()
+         );
+         return entry;
+      }
+      SPDLOG_WARN(
+         "The database output {} is incompatible with the current SILO version.",
+         entry.second.toString()
+      );
```
fengelniederhammer, line 104:
```suggestion
      SPDLOG_WARN(
         "The database output {} is incompatible with the current SILO version ({}) - skipping it.",
         entry.second.toString(),
         silo::DataVersion::SILO_SERIALIZATION_VERSION.value
      );
```

fengelniederhammer (approved):
LGTM :+1:

fengelniederhammer:
How do we document this feature? 
* Something for maintainers would be good so that they know that the concept exists.
* We should write down something for developers so that they are aware that they might need to adapt the version number when making changes to the DB.

Let's at least create issues to remind us to create these docs, if we don't do it now?

PR #443, opened 2024-05-24
test: add test that checks that we do not regress re/ non-standard column names

Taepper:
~resolves #~

### Summary

A bug was mentioned by @corneliusroemer a while back that SILO crashed on certain numeric sequence names. When I fixed this I missed to implement a test, which guarantees that we do not regress in this regard. This is that missed test

## PR Checklist

~- [ ] The implemented feature is covered by an appropriate test.~


testBaseData/numericNames/database_config.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - name: version
+      type: int
+    - name: submissionId
+      type: string
+    - name: accessionVersion
+      type: string
+    - name: isRevocation
+      type: string
+    - name: versionStatus
+      type: string
```
fengelniederhammer, line 29: Are all those fields important for the test case? If not, I would suggest to remove them (only keeping the primary key).

testBaseData/numericNames/reference_genomes.json
```diff
@@ -0,0 +1,22 @@
...
+      "sequence": "AAAAAAAAAAAAAAAA"
+    }
+  ],
+  "genes": [
+    {
+      "name": "someLongGene",
+      "sequence": "AAAAAAAAAAAAAAAAAAAAAAAAA"
+    },
+    {
+      "name": "someShortGene",
```
fengelniederhammer, line 18: Let's also have a numeric gene?

src/silo/preprocessing/preprocessor.test.cpp
```diff
@@ -83,6 +83,29 @@ const Scenario NDJSON_WITH_SQL_KEYWORD_AS_FIELD = {
    )
 };
 
+const Scenario NDJSON_WITH_NUMERIC_NAMES = {
+   .input_directory = "testBaseData/numericNames/",
+   .expected_sequence_count = 2,
+   .query = R"(
+      {
+         "action": {
+            "type": "Aggregated",
```
fengelniederhammer, line 92: Is this a relevant query for this test case? Let's maybe do a `Fasta` action instead?

Taepper: The FastaAligned action cannot filter for the numeric metadata field

fengelniederhammer: Ah, I overlooked that there is also a numeric metadata field :+1:

fengelniederhammer (approved):

PR #450, opened 2024-05-28
chore: fix CLion run config

fengelniederhammer:
### Summary

https://github.com/GenSpectrum/LAPIS-SILO/pull/438 apparently still introduced a breaking change in the command line arguments

## PR Checklist


Taepper:
Ahh, I see I broke it here
```
static const std::string PREPROCESSING_CONFIG_OPTION = "preprocessing-config";
static const std::string RUNTIME_CONFIG_OPTION = "runtime-config";
static const std::string DATABASE_CONFIG_OPTION = "database-config";
```
but do we rather want to revert that, as not to have to bump a version immediately again

fengelniederhammer:
Ok, then let's do #451 instead.

PR #455, opened 2024-05-29
fix: Remove incorrect compile flags

### LINKED ISSUE #448: Spurious abort instead of exception
SILO sometimes aborts when exceptions are thrown within TBB parallelization blocks. This only occurs when SILO is compiled with build-type release

Taepper:
resolves #448

### Summary

Remove incorrect flags: `set(CMAKE_CXX_FLAGS_RELEASE "[...] -static-libstdc++ -static-libgcc")`

These should instead be declared as

`target_link_options(siloApi PUBLIC  -static-libstdc++ -static-libgcc)`

because `CMAKE_CXX_FLAGS..` are used for the compilation of every unit (.cpp) file. This links standard library code into the output multiple times.

But having that compilation flag is also not desirable, as the tbb's shard object (.so) file is itself linked against shared libgcc and libstdc++ as verified by

```
> ldd libtbb.so
	linux-vdso.so.1 (0x00007fffcbca7000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fc3db913000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fc3db8f3000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc3db6ca000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc3dbba2000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc3db5e3000)
```

And using different standard libraries (statically linked in our code, shared in tbb dependency) leads to problems as pointed out by `man gcc`
>        -static-libgcc
>
>           ...   
>
>           There are several situations in which an application should use the shared libgcc instead of the static version.  The most common of these is when
>           the application wishes to throw and catch exceptions across different shared libraries.  In that case, each of the libraries as well as the
>           application itself should use the shared libgcc.

.. and as observed in #448 

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
 set(CMAKE_CXX_FLAGS "-Wall")
 set(CMAKE_CXX_FLAGS_DEBUG "-g -fsanitize=address")
-set(CMAKE_CXX_FLAGS_RELEASE "-O3 -static-libstdc++ -static-libgcc")
+set(CMAKE_CXX_FLAGS_RELEASE "-O3")
 
 # Work-around only for MacOS
 if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")


pflanze:
I think this warrants a description on how to reproduce the error (best in the body of the commit message). It's probably worth including the minimal test case that you made, too, or alternatively give the diff of the change that you applied to silo to make the issue show up reliably (I'd also include that in the commit message, indented by 4 spaces).

pflanze:
I have added some more information to the commit message, from my notes. Merge if you're happy with that.

pflanze (approved):

Taepper:
Ah, yes that looks very nice and is certainly a better location than just having it in the PR body.

I will change 
> .. due to tcc being linked dynamically ..
to `tbb` and merge

pflanze:
Ah, I had forgotten that you had some info in the PR already when I edited the commit message. But except for the `ldd` output I think it's complete anyway. I also forgot to add a 

> This links standard library code into the output multiple times.

Actually it doesn't apparently, since the size of the .o files stayed the same when I looked. But that was a side concern only anyway.

(I forgot to add a `Refs: ` header. We're not very consistent yet anyway though.)

PR #461, opened 2024-06-03
chore: Use ubuntu as server base image

corneliusroemer:
- **chore(deps): use Ubuntu 24.04 as base image, update dependencies**
- **No install recommends for final image**

resolves #

## PR Checklist


Dockerfile
```diff
@@ -4,27 +4,34 @@ FROM $DEPENDENCY_IMAGE AS builder
...
 
-RUN  \
-    python ./build_with_conan.py --release --parallel 4\
+RUN python3 ./build_with_conan.py --release --parallel 4 \
     && cp build/silo_test . \
     && cp build/siloApi .
 
 
-FROM alpine:3.18 AS server
+FROM ubuntu:24.04 AS server
```
fengelniederhammer, line 12: Why would you want to use Ubuntu here? This adds a lot of things that we actually don't need at runtime, which 1. bloats the image from 50 MB to ~1 GB, 2. is a security risk, because the container has much more tools available.

(Might be that i missed a discussion in my absence)

Taepper:
Thank you, done in #519 !

PR #471, opened 2024-06-10
feat: allow null for sequenceName in insertion contains queries

JonasKellerer:
### Summary

Fixes the error, that users could not pass `segmentName: null` or `segmentName: ""` in `insertion_contains` queries. This is necessary for nucleotide insertion queries. https://github.com/GenSpectrum/LAPIS/issues/806 is related to this.

Do we want to treat an empty string the same as passing null? (Currently I have implemented it like it, but we don't have to.

Additionally, we can now test that a query throws an error.

## PR Checklist

### DIFF
--- a/include/silo/query_engine/query_parse_sequence_name.h
+++ b/include/silo/query_engine/query_parse_sequence_name.h
@@ -1,0 +1,42 @@
+#pragma once
+
+#include <optional>
+#include <string>
+
+#include "silo/database.h"
+
+namespace silo {
+
+template <typename SymbolType>
+std::string validateSequenceName(std::string sequence_name, const silo::Database& database) {
+   CHECK_SILO_QUERY(
+      database.getSequenceStores<SymbolType>().contains(sequence_name),
+      fmt::format(
+         "Database does not contain the {} Sequence with name: '{}'",
+         SymbolType::SYMBOL_NAME,
+         sequence_name
+      )
+   );
+   return sequence_name;
+}
+
+template <typename SymbolType>
+std::string validateSequenceNameOrGetDefault(
+   std::optional<std::string> sequence_name,
+   const silo::Database& database
+) {
+   if (sequence_name.has_value()) {
+      return validateSequenceName<SymbolType>(sequence_name.value(), database);
+   }
+
+   CHECK_SILO_QUERY(
+      database.getDefaultSequenceName<SymbolType>().has_value(),
+      "The database has no default " + std::string(SymbolType::SYMBOL_NAME_LOWER_CASE) +
+         " sequence name"
+   );
+
+   const auto default_sequence_name = database.getDefaultSequenceName<SymbolType>().value();
+   return validateSequenceName<SymbolType>(default_sequence_name, database);
+}
+
+}  // namespace silo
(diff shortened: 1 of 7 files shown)


Taepper:
But isn't this now diverging from the other filters work?

JonasKellerer:
As discussed: we currently dont want to support explicit null values for segmentName (silimar to symbol_equals.cpp). To pass an insertionContains filter, the segmentName must be left out.

include/silo/query_engine/query_parse_sequence_name.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
```
JonasKellerer, line 1: Reminder: squash before merge

src/silo/query_engine/filter_expressions/insertion_contains.cpp
```diff
@@ -55,28 +56,14 @@ std::unique_ptr<silo::query_engine::operators::Operator> InsertionContains<Symbo
...
-      // NOLINTNEXTLINE(bugprone-unchecked-optional-access) -- the previous statement checks it
-      validated_sequence_name = *database.getDefaultSequenceName<SymbolType>();
-   }
+   std::string valid_sequence_name =
+      getValidSequenceNameOrDefault<SymbolType>(sequence_name, database);
+
    const std::map<std::string, SequenceStorePartition<SymbolType>&>& sequence_stores =
       database_partition.getSequenceStores<SymbolType>();
-   CHECK_SILO_QUERY(
-      sequence_stores.contains(validated_sequence_name),
```
Taepper, line 73: Is there a specific reason you removed this check?

Taepper: A misconfigured "default sequence" would lead to an error here. Unless we can circumvent that elsewhere

JonasKellerer: I will enable the check again, and add a ticket to fix this when we have time.

JonasKellerer: Ticket: #475

include/silo/query_engine/query_parse_sequence_name.h
```diff
@@ -0,0 +1,42 @@
...
+   }
+
+   CHECK_SILO_QUERY(
+      database.getDefaultSequenceName<SymbolType>().has_value(),
+      "The database has no default " + std::string(SymbolType::SYMBOL_NAME_LOWER_CASE) +
+         " sequence name"
+   );
+
+   const auto default_sequence_name = database.getDefaultSequenceName<SymbolType>().value();
+   return validateSequenceName<SymbolType>(default_sequence_name, database);
```
Taepper, line 39: Oh no, now I see that my previous comment on the check being necessary was misguided, as it will be verified here even when it takes the default sequence name! Sorry for the extra work. So I guess we can remove the check outside this function afterall.

LGTM!

Taepper (approved):

Taepper:
Thank you very much, it looks great!

Taepper:
Sorry, I mistakenly cancelled the workflow instead of a different one. Restarted it now

PR #478, opened 2024-06-11
fix image tags to actually tag latest on main branch for dependency images

### LINKED ISSUE #445: linter-dependencies latest tag
If running Dockerfile_linter manually it uses the latest tag of the linter-dependencies image by default. But we never tag it as such in the CI

Taepper:
resolves #445 

### Summary

Before, the latest tag was only added to the metadata when we are on the main branch : (github.ref == 'refs/heads/main')

Together with the fact that we build the dependency images only when we did not hit the cache lead to the fact that the latest tag was actually never applied to our dependency images.

The change was to introduce the extra logic as by Cornelius in _the other_ repo.

## PR Checklist


fengelniederhammer (approved):
LGTM

PR #481, opened 2024-06-12
fix: more efficient ndjson emptiness check

Taepper:
resolves issue where SILO occasionally uses a lot more memory for preprocessing since 0.2.3

### Summary

Before we did a `COUNT(*)` on the entire file, which is costly. Now we do the `COUNT(*)` only on the first row of the file.

## PR Checklist


JonasKellerer (approved):
LGTM

Works on production.

PR #482, opened 2024-06-13
refactor: lambda-callbacks instead of fill columns

Taepper:
resolves #

### Summary

instead of the columns pulling the next row from the file_reader, the file_reader
 instead pushes the value to the columns using callbacks.

This is in preparation for parallelizing the building of column functions

## PR Checklist


davidgicev (approved):

JonasKellerer (approved):
LGTM

PR #483, opened 2024-06-17
Process chunks column-wise in the table reader

### LINKED ISSUE #291: Parallel building of Columns
Columns are currently not built in parallel.

This is because:
- Columns with dictionary encoding need be synchronized when inserting values
- Columns are built 1 value at a time. That means that we read one row from the input and add all its values to all columns

This means that we would need to grab a mutex for multiple columns per row when building the columns from the metadata.

We can make parallel building and grabbing of mutexes a viable approach by reading larger groups of rows at a time and then add groups of values to each vector. This means that we do not need to grab a mutex per value, but only one mutex per group of values.

We should write a wrapper class for the duckdb preprocessor's `metadata_table` that reads many rows at a time (~1000)

davidgicev:
resolves #291

### Summary

We now use tbb::parallel_for to build the different columns in parallel per input chunk.

## PR Checklist

### DIFF
--- a/src/silo/common/table_reader.cpp
+++ b/src/silo/common/table_reader.cpp
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include <fmt/format.h>
+#include <oneapi/tbb/parallel_for.h>
 #include <spdlog/spdlog.h>
 #include <duckdb.hpp>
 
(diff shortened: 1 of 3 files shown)


src/silo/common/table_reader.cpp
```diff
@@ -26,24 +34,38 @@ silo::TableReader::TableReader(
...
+      if (current_chunk->size() == 0) {
+         continue;
+      }
+      current_chunk_size = current_chunk->size();
+
+      tbb::parallel_for(
+         tbb::blocked_range<uint32_t>(0, column_functions.size()),
+         [&](const auto& local) {
+            for (size_t column_idx = local.begin(); column_idx != local.end(); column_idx++) {
+               auto column = current_chunk->data[column_idx + 1];
```
Taepper, line 54: Is this copy really necessary? Does anything speak against turning this into a reference (possible const?, if GetValue allows it)

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -651,15 +651,41 @@ void Preprocessor::buildMetadataStore(
...
+         order_by_clause
+      );
+
+      int64_t fill_time;
+      {
+         const silo::common::BlockTimer timer(fill_time);
+         const size_t number_of_rows = table_reader.read();
+         database.partitions.at(partition_id).sequence_count += number_of_rows;
       }
+      SPDLOG_INFO(
```
Taepper, line 684: Let's change the loglevel to `SPDLOG_DEBUG` and we can keep this in :)

Taepper:
resolves #291

Taepper:
And also rebase onto current main for a better change list :)

fengelniederhammer:
What's the new feature in this PR? Or should the commit message say `refactor` rather than `feat`?

Taepper (approved):

PR #500, opened 2024-06-20
build: share library for test and api

Taepper:
### Summary

Currently two object files are generated for every source file. See the linter doing his work twice:
https://github.com/GenSpectrum/LAPIS-SILO/actions/runs/9599350811/job/26473097915
But also the [normal build step](https://github.com/GenSpectrum/LAPIS-SILO/actions/runs/9599350009/job/26472879416) lists every source file twice, once for each target:
```
#9 367.9 [ 75%] Building CXX object CMakeFiles/siloApi.dir/src/silo/storage/column/date_column.cpp.o
#9 368.8 [ 76%] Building CXX object CMakeFiles/silo_test.dir/src/silo/storage/column/date_column.cpp.o
```
Here cmake is very conservative as different flags could be used to compile the same source file twice.
But we are the ones in control here and can adjust the `CMakeLists.txt` to compile the files once as a library, but without archiving that library using the [Object Library feature](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries). The object library feature does exactly what we want. As we do not want to distribute the library, expensively archiving all sources into a `.a`-archive is not being done. Still, all object files are only compiled once.

We also need to only list the relevant linked libraries once, as `target_link_libraries` correctly inherits them from the shared library.

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,9 @@ file(GLOB_RECURSE SRC_TEST "src/*.test.cpp")
 file(GLOB_RECURSE SRC_SILO "src/*.cpp")
 list(REMOVE_ITEM SRC_SILO ${SRC_TEST})
 
+set(SRC_SILO_WITHOUT_MAIN ${SRC_SILO})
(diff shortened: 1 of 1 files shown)


fengelniederhammer (approved):
LGTM

PR #502, opened 2024-06-25
fix: don't abort when reading table in chunks

fengelniederhammer:
### Summary

The bug was introduced in #483 and only occured when the table reader tried to read the second chunk. Thus: also add a preprocessor test with a dataset that is large enough so that the bug occurs.

## PR Checklist


Taepper (approved):
LGTM

PR #504, opened 2024-06-28
chore: run clang-tidy (linter) on all files when on main or manually triggered

### LINKED ISSUE #498: clang-tidy on all files
Add workflow that is triggered by push to main that runs clang-tidy on all source files

Taepper:
resolves #498 

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,3 +233,25 @@ jobs:
               cmake --build build/Debug --target ${file%.cpp}.o
             fi
           done
+
+  linterAll:
+    name: Build/Run linter on all files
+    needs: dockerImage
+    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'trigger-linter')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Determine HEAD_SHA
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            echo "HEAD_SHA=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+          else
+            echo "HEAD_SHA=${{ github.sha }}" >> $GITHUB_ENV
+          fi
+
+      - shell: bash
+        name: Build all files with clang-tidy
+        run: |
+          docker build -f ./Dockerfile_linter \
+            --build-arg DEPENDENCY_IMAGE=ghcr.io/genspectrum/lapis-silo-dependencies:commit-${{ env.HEAD_SHA }}-amd64 . 
--- a/Dockerfile_linter
+++ b/Dockerfile_linter
@@ -8,4 +8,4 @@ WORKDIR /src
 COPY . ./
 
 RUN cmake -D BUILD_WITH_CLANG_TIDY=ON -D CMAKE_BUILD_TYPE=Debug -B build/Debug
-RUN cmake --build build/Debug
+RUN cmake --build build/Debug --parallel $THREADS


.github/workflows/clang-tidy-full.yml
```diff
@@ -0,0 +1,55 @@
...
+  waitForDependencyImage:
+    name: Wait for Docker Image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for Docker Image
+        uses: lewagon/wait-on-check-action@v1.3.1
+        with:
+          ref: ${{ github.ref }}
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          check-name: Build linter dependencies
```
fengelniederhammer, line 23: Do we need this job?
```
on:
  workflow_run:
    workflows: [Build Linter Image]
``` 
should already do the waiting.

.github/workflows/clang-tidy-full.yml
```diff
@@ -0,0 +1,55 @@
...
+          echo "$files"
+          IFS=$'\n'
+          for file in $files; do
+            echo "Check ending for file: $file"
+            if [[ $file == *.cpp ]]; then
+              echo "Now linting the file: $file"
+              echo "cmake --build build --target ${file%.cpp}.o"
+              cmake --build build --target ${file%.cpp}.o
+            fi
+          done
```
fengelniederhammer, line 55: Can't we simply use the old approach when running the linter on all files (i.e. simply building the code)?

Taepper: Ah, yes, that was supposed to be the case. I was still experimenting with the action triggers and waits! Sorry for not marking it as draft!

.github/workflows/clang-tidy-full.yml
Taepper: True. I did not know you can still define branch as a condition. But this does not work for the diffed files unfortunately as I need the PR number in the action to get the list of changed files. And I can only get it in the PR trigger event

.github/workflows/ci.yml
```diff
@@ -230,3 +230,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/genspectrum/lapis-silo-dependencies:commit-${{ github.sha }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - shell: bash
+        name: Configure and run clang-tidy on changed files
+        run: |
+          mv /src/build .
```
fengelniederhammer, line 247: Why do we need this `mv`? Can't we `cd` into the correct directory?

.github/workflows/ci.yml
```diff
@@ -230,3 +230,20 @@ jobs:
...
+          cmake -DBUILD_WITH_CLANG_TIDY=on -D CMAKE_BUILD_TYPE=Debug -B build/Debug
```
fengelniederhammer, line 248: Does the space make a difference?
```suggestion
          cmake -D BUILD_WITH_CLANG_TIDY=on -D CMAKE_BUILD_TYPE=Debug -B build/Debug
```

.github/workflows/ci.yml
```diff
@@ -230,3 +230,20 @@ jobs:
...
+          cmake --build build/Debug
```
fengelniederhammer, line 249: Comparing to `build_with_conan`, we're usually running this with `--parallel`. Should we also do that here?

fengelniederhammer (commented):
Could you please rebase to get rid of the superfluous commits?

.github/workflows/ci.yml
```diff
@@ -233,3 +233,25 @@ jobs:
...
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            echo "HEAD_SHA=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+          else
+            echo "HEAD_SHA=${{ github.sha }}" >> $GITHUB_ENV
+          fi
+
+      - shell: bash
+        name: Configure and run clang-tidy on changed files
+        run: |
+          docker build -f ./Dockerfile_dependencies \
```
fengelniederhammer, line 256: Shouldn't this take the linter Dockerfile?
```suggestion
          docker build -f ./Dockerfile_linter \
```

.github/workflows/ci.yml
```diff
@@ -233,3 +233,25 @@ jobs:
...
+      - name: Determine HEAD_SHA
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            echo "HEAD_SHA=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+          else
+            echo "HEAD_SHA=${{ github.sha }}" >> $GITHUB_ENV
+          fi
+
+      - shell: bash
+        name: Configure and run clang-tidy on changed files
```
fengelniederhammer, line 254: This name looks like copy paste, I think we can have something more suitable for this job here
```suggestion
        name: Build all files with clang-tidy
```

fengelniederhammer (changes requested):

fengelniederhammer (approved):

PR #509, opened 2024-07-16
Fix the slow preprocessing with more recent releases

Taepper:
## PR Checklist


fengelniederhammer (approved):
LGTM, let's try it.

Issue #426, opened 2024-05-16
Specify unaligned sequences separately in the DatabaseConfig


taepper:
Currently, we expect an unaligned sequence for every aligned sequence. It should be possible to specify this

Taepper:
This came up again in #767 

As it is auto-generated with the prefix `unaligned_`, this now leads to a breaking API change, as we do not want to hard-code the renaming of the output

taepper:
We do not want to change the `DatabaseConfig` format, it will be specified differently in the python bindings and query interface

PR #510, opened 2024-07-16
default nucleotide and amino acid sequence

### LINKED ISSUE #454: Default nucleotide and amino acid sequence
The default nucleotide and amino acid sequence, that is used when it is omitted in queries, should be better configurable.

- it is not possible to set a default amino acid sequence. This should be added to the database config.
- the default nucleotide sequence is `main`.  The default should be unset. A null should be treated as unset.
- It should be verified that the default sequence exists in the provided dataset.

fengelniederhammer:
resolves #454

### Summary

Changes in this PR:
* when no default nucleotide sequence is set in the config, set it to absent (instead of defaulting to "main")
* read the default amino acid sequence from the config (instead of hardcoding it to absent)
* validate that the values are actually present in the reference genomes (if set)
* if there is only one sequence of a type in the reference genomes, also set the default sequence in the database config
* add a unit test that checks that relevant mutation queries work without a `sequenceName` field, if the config has the default sequences.

## PR Checklist


src/silo/config/database_config.cpp
```diff
@@ -218,8 +244,13 @@ DatabaseConfig DatabaseConfigReader::readConfig(const std::filesystem::path& con
 ) -> decltype(ctx.out()) {
    return fmt::format_to(
       ctx.out(),
-      "{{ default_nucleotide_sequence: '{}', schema: {} }}",
-      database_config.default_nucleotide_sequence,
+      "{{ default_nucleotide_sequence: '{}', default_amino_acid_sequence: '{}', schema: {} }}",
```
Taepper, line 247: This will print `'null'` instead of `null`

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -126,6 +128,42 @@ Database Preprocessor::preprocess() {
    );
 }
 
+void Preprocessor::checkConfig() {
```
Taepper, line 131: I find this function name a little confusing

Taepper (changes requested):
That was quick! A few minor comments

src/silo/preprocessing/preprocessor.cpp
fengelniederhammer: Because of the name or because of the content? I'm also not super happy with the name. Suggestions are welcome ;)

Taepper: The name made me think that it would not mutate the config
Maybe finalize / resolve?

Otherwise an extra class ValidatedDatabaseConfig which also takes the ReferenceGenomes in its constructor would be viable

or we can leave it as a TODO for #426 , where we want to specify the unaligned sequences (and I would also say the aligned nuc and aa sequences) in the database config to reduce this dependency. Then the reference genomes is no longer a config for the sequences in the database but only a map for looking up the references

fengelniederhammer: Ok good point. `check` doesn't imply mutating. I renamed it to `finalizeConfig`.

`ValidatedDatabaseConfig` is only partially true, because you also need to validate it against the data (e.g. that all metadata fields are present etc.). So I'm not sure whether that would be a good name either.
If we want to redesign this anyway, then I'd propose to leave the rest to #426.

Taepper (approved):

PR #512, opened 2024-07-17
Report limit/offset error

### LINKED ISSUE #511: Streaming: report an error when the unimplemented offset or limit matters
Instead of silently ignoring offset and limit, complain to the user if it was being requested.

pflanze:
resolves #511

### Summary

I can't quickly add tests for this, as writing them is a large part of the work implementing offset+limit (it needs large dataset, or reducing the cut off via unit tests). Thus the point pushing this fix quickly would be void. I suggest quickly testing this manually instead.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/actions/action.cpp
+++ b/src/silo/query_engine/actions/action.cpp
@@ -8,6 +8,7 @@
 #include <random>
 #include <utility>
 
+#include <fmt/format.h>
 #include <spdlog/spdlog.h>
 #include <nlohmann/json.hpp>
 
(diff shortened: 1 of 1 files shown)


src/silo/query_engine/actions/action.cpp
```diff
@@ -117,15 +118,13 @@ QueryResult Action::executeAndOrder(
    // sorting) for small result sets, and streaming without those
    // features for larger ones.
 
+   size_t num_rows_at_least = 0;
```
Taepper, line 121: Why do we compute the `num_rows_at_least` and not the `num_rows`? Seems like a premature optimization to me, or did you see that the performance hit of the prior was relevant?

pflanze: Because before this temporary hack, the number would never be used, only `is_large`. I just renamed the variable from the previous local `num_rows` variable to spell out what it contains in the end.

Just writing this code was easier/faster than doing any measurement or proper analysis. Maybe reading this code in the future is better than reading code that calculates the whole num_rows, too, because otherwise some people start pondering if there's a potential for optimization there and lose time on that? 

(Maybe `std::vector<OperatorResult> bitmap_filter` isn't ideal either and should be in an object, then that object could have a (potentially cached) `size_t numRows()` method and if relevant, `bool atLeastNumRows(size_t n)`, but then making objects is a bit painful in C++...)

Taepper: Okay

Taepper (approved):

src/silo/query_engine/actions/action.cpp
pflanze: So, should I:

- move the for loop to a separate function (not a wrapper class)
- leave it as is
- change the for loop to not break and do the comparison after, and add a comment that optimizing this isn't expected to give any benefit
- make a wrapper

These are in the order of my preference, most preferred first.

PR #519, opened 2024-07-22
chore: change base image to ubuntu

Taepper:
### Summary

Tests revealed that there is a tremendous difference in speed between alpine and ubuntu images. The 182MB vs 62MB are negligible. 

debian images have a similar memory footprint to ubuntu at 220MB for debian:12 and 178MB for debian:12-slim images

## PR Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,19 +5,20 @@ FROM $DEPENDENCY_IMAGE AS builder
 COPY . ./
 
 RUN  \
-    python ./build_with_conan.py --release --parallel 4\
+    python3 ./build_with_conan.py --release --parallel 4\
     && cp build/silo_test . \
     && cp build/siloApi .
 
 
-FROM alpine:3.20 AS server
+FROM ubuntu:22.04 AS server
 
 WORKDIR /app
 COPY docker_default_preprocessing_config.yaml ./default_preprocessing_config.yaml
 COPY docker_runtime_config.yaml ./runtime_config.yaml
 COPY --from=builder /src/siloApi ./
 
-RUN apk update && apk add onetbb=2021.12.0-r0 curl jq
+RUN apt update \
+    &&  apt install -y libtbb-dev=2021.5.0-7ubuntu2 curl jq
 
 # call /info, extract "seqeunceCount" from the JSON and assert that the value is not 0. If any of those fails, "exit 1".
 HEALTHCHECK --start-period=20s CMD curl --fail --silent localhost:8081/info | jq .sequenceCount | xargs test 0 -ne || exit 1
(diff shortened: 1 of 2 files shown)


Taepper:
The chore label is definitely wrong. I guess this should be a `feat` (of the generated image)?

Dockerfile_dependencies
```diff
@@ -1,16 +1,14 @@
...
 
 ARG TARGETPLATFORM
 
-RUN apk update && apk add --no-cache py3-pip \
-    build-base=0.5-r3 \
-    cmake=3.29.3-r0 \
-    bash=5.2.26-r0 \
-    linux-headers=6.6-r0 \
-    boost-build=1.84.0-r0 \
-    onetbb=2021.12.0-r0
```
fengelniederhammer, line 11: Why don't we need to install tbb here? I would expect that tbb needs to be installed so that the build later in `Dockerfile` succeeds? Is it maybe already installed?

fengelniederhammer (approved):
LGTM, let's try it.

Dockerfile
```diff
@@ -5,19 +5,20 @@ FROM $DEPENDENCY_IMAGE AS builder
...
+FROM ubuntu:22.04 AS server
 
 WORKDIR /app
 COPY docker_default_preprocessing_config.yaml ./default_preprocessing_config.yaml
 COPY docker_runtime_config.yaml ./runtime_config.yaml
 COPY --from=builder /src/siloApi ./
 
-RUN apk update && apk add onetbb=2021.12.0-r0 curl jq
+RUN apt update \
+    &&  apt install -y libtbb-dev=2021.5.0-7ubuntu2 curl jq
```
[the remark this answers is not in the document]
fengelniederhammer: We already had this discussion and I'm not sure what the better approach is.

Fixing the version has the huge benefit of reproducable builds, but it leaves you with maintaining the versions manually.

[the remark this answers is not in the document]
pflanze: Ah, if you're happy to update the image, then that makes sense. You may want to get some notification about security upgrades then, though.

Dockerfile
```diff
@@ -5,19 +5,20 @@ FROM $DEPENDENCY_IMAGE AS builder
...
-FROM alpine:3.20 AS server
+FROM ubuntu:22.04 AS server
 
 WORKDIR /app
 COPY docker_default_preprocessing_config.yaml ./default_preprocessing_config.yaml
 COPY docker_runtime_config.yaml ./runtime_config.yaml
 COPY --from=builder /src/siloApi ./
 
-RUN apk update && apk add onetbb=2021.12.0-r0 curl jq
+RUN apt update \
```
fengelniederhammer, line 20: Also running `apt upgrade` would also be good to update the installed stuff.

Taepper: Hmm, but isn't this precisely contrary to your previously mentioned point of reproducible builds?

Not that I am against this, but if we do want the upgrade, maybe we should rethink our decision on fixed versions afterall?

fengelniederhammer: This is also about general security upgrades for installed system dependencies that SILO doesn't use (e.g. ssl)

PR #522, opened 2024-07-22
Report sort error

pflanze:
resolves "also reject sorting" request on Slack.

### Summary

IIRC we never discussed the further plans for sorting when streaming. Probably refusing with an error ("unimplemented") is really the right thing to do so far. It's a bit weird and potentially problematic in that FASTA/FASTA aligned queries will succeed with sorting first, then when the data set grows until the result set becomes >10'000 rows, they will start to fail. But that may still be better than silently wrong sorting. It would be good to warn the user in advance (while there are still <10'000 rows) somehow; again something that NDJSON cannot deliver, while JSON could contain richer data than just the rows. Documentation will have to be enough, or perhaps the dashboard or other UI components could warn about such queries.
I worry a bit more now that Details should also be changed to stream, and will expose the same conundrum.

As per discussion (Fabian), for Details we might set the cut-off a lot higher though since/if the involved data is much smaller, and might make the problem go away for most organisms.

There is no test as it's still impossible to test without large data or future work, thus will test manually.

User documentation should be added, but where?

## PR Checklist

### DIFF
--- a/src/silo_api/query_handler.cpp
+++ b/src/silo_api/query_handler.cpp
@@ -50,9 +50,11 @@ void QueryHandler::post(
             );
          }
       }
-   } catch (const silo::QueryParseException& ex) {
+   } catch (const silo::QueryException& ex) {
       response.setContentType("application/json");
-      SPDLOG_INFO("Query is invalid: " + query + " - exception: " + ex.what());
+      SPDLOG_INFO(
+         "Query is invalid: {} - exception during {}: {}", query, ex.duringString(), ex.what()
+      );
       response.setStatusAndReason(Poco::Net::HTTPResponse::HTTP_BAD_REQUEST);
       std::ostream& out_stream = response.send();
       out_stream << nlohmann::json(ErrorResponse{.error = "Bad request", .message = ex.what()});
--- a/include/silo/query_engine/query_parse_exception.h
+++ b/include/silo/query_engine/query_parse_exception.h
@@ -12,6 +12,6 @@ namespace silo {
 
 class [[maybe_unused]] QueryParseException : public std::runtime_error {
   public:
-   [[maybe_unused]] QueryParseException(const std::string& error_message);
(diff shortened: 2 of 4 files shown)


Taepper:
I thought the second part of streaming was now to introduce the config cutoff option?

pflanze:
> I thought the second part of streaming was now to introduce the config cutoff option?

This is not "the second part of streaming", but the second part of implementing the (small) followup requests for the first part.

"The second part of streaming" will contain the parts that make it possible to automatically test.

pflanze:
I forgot to update the comment in the code that it's not a hack anymore, will do later, have to leave.

src/silo_api/query_handler.cpp
```diff
@@ -50,9 +50,11 @@ void QueryHandler::post(
             );
          }
       }
-   } catch (const silo::QueryParseException& ex) {
+   } catch (const silo::QueryException& ex) {
       response.setContentType("application/json");
-      SPDLOG_INFO("Query is invalid: " + query + " - exception: " + ex.what());
+      SPDLOG_INFO(
+         "Query is invalid: {} - exception during {}: ", query, ex.duringString(), ex.what()
```
fengelniederhammer, line 56:
```suggestion
         "Query is invalid: {} - exception during {}: {}", query, ex.duringString(), ex.what()
```

fengelniederhammer (approved):
LGTM

fengelniederhammer:
I think we can make the `fix: action: use more appropriate new QueryEvaluationException` a `refactor`, since it's internal only.

src/silo_api/query_handler.cpp
pflanze: Oh, no compile time check. (Another reason to use Rust.) Good catch!

fengelniederhammer: I was wondering why the log statement didn't include the message... :D

PR #529, opened 2024-07-31
Allow more preprocessing sequence names

### LINKED ISSUE #465: Allow more diverse sequence names
A `-` in a sequence name currently breaks the preprocessing. This should be fixed and be made more robust overall by removing the sequence names from most of the SQL statements

Taepper:
resolves #465 

### Summary

We now correctly escape all sequence names that appear in table names, UDFs, column names, etc.

Currently we still disallow whitespace, as duckdb's tsv reader will not generate column names from this which we need for binding

## PR Checklist


src/silo/database.cpp
```diff
@@ -427,27 +427,33 @@ void Database::saveDatabaseState(const std::filesystem::path& save_directory) {
...
    ::boost::archive::binary_oarchive nuc_sequences_archive(nuc_sequences_file);
-   nuc_sequences_archive << nuc_sequences_map;
+   nuc_sequences_archive << nuc_sequence_names;
+   nuc_sequences_archive << nuc_reference_sequences;
 
    auto aa_sequences_map = getAASequences();
    std::ofstream aa_sequences_file =
       openOutputFileOrThrow(versioned_save_directory / "aa_sequences.silo");
    ::boost::archive::binary_oarchive aa_sequences_archive(aa_sequences_file);
+   aa_sequences_archive << aa_sequence_names;
```
fengelniederhammer, line 441: This looks like another bump in the serialization version.

include/silo/storage/reference_genomes.h
```diff
@@ -11,27 +11,31 @@
...
 
 struct ReferenceGenomes {
-   std::map<std::string, std::vector<Nucleotide::Symbol>> nucleotide_sequences;
-   std::map<std::string, std::vector<AminoAcid::Symbol>> aa_sequences;
-   std::map<std::string, std::string> raw_nucleotide_sequences;
-   std::map<std::string, std::string> raw_aa_sequences;
+   std::vector<std::string> nucleotide_sequence_names;
+   std::vector<std::string> aa_sequence_names;
+   std::vector<std::vector<Nucleotide::Symbol>> nucleotide_sequences;
+   std::vector<std::vector<AminoAcid::Symbol>> aa_sequences;
```
fengelniederhammer, line 17: Is it really a good idea to store sequence names and sequences separately? It bloats the code in several places and detaches information that actually belongs together (and thus should be attached).

src/silo/config/preprocessing_config.cpp
```diff
@@ -57,25 +57,22 @@ std::optional<std::filesystem::path> PreprocessingConfig::getNdjsonInputFilename
...
 }
 
-std::filesystem::path PreprocessingConfig::getNucFilenameNoExtension(std::string_view nuc_name
-) const {
+std::filesystem::path PreprocessingConfig::getNucFilenameNoExtension(size_t sequence_idx) const {
    std::filesystem::path filename = input_directory;
-   filename /= fmt::format("{}{}", nucleotide_sequence_prefix, nuc_name);
+   filename /= fmt::format("{}{}", nucleotide_sequence_prefix, sequence_idx);
    return filename;
 }
```
fengelniederhammer, line 64: IMO it's not good to expose the sequences index to the outside. Users will have to provide sequence files using this index and thus it relies on the order in which you provide it in the JSON file.

As a user I would expect that those reference genomes yield the same result and that I have to provide the same input files:
```json
{
  "nucleotideSequences": [
    {
      "name": "first",
      "sequence": "T"
    },
    {
      "name": "second",
      "sequence": "T"
    }
  ],
  "genes": []
}
``` 

```json
{
  "nucleotideSequences": [
    {
      "name": "second",
      "sequence": "T"
    },
    {
      "name": "first",
      "sequence": "T"
    }
  ],
  "genes": []
}
``` 

If I understand the code correctly, that's currently not the case, because `nuc_0.fasta` would be the `first` sequence with the first reference genomes, but the `second` sequence with the second reference genomes.

src/silo/test/amino_acid_insertion_contains.test.cpp
```diff
@@ -40,10 +40,8 @@ const auto DATABASE_CONFIG = DatabaseConfig{
        .primary_key = "primaryKey"}
 };
 
-const auto REFERENCE_GENOMES = ReferenceGenomes{
-   {{"segment1", "A"}, {"segment2", "T"}},
-   {{"gene1", "*"}, {"gene2", "*"}},
-};
+const auto REFERENCE_GENOMES =
+   ReferenceGenomes{{"segment1", "segment2"}, {"gene1", "gene2"}, {"A", "T"}, {"*", "*"}};
```
fengelniederhammer, line 44: This looks very unintuitive. It's harder to see what belongs together than before.

fengelniederhammer (commented):
Let's talk about these changes. The internal parts look reasonable, but I tried to point out some of my pain points in the comments.

src/silo/config/preprocessing_config.cpp
Taepper: The thing is, that I wanted to allow all sequence names, that are provided by the user and work with duckdb. This means, that I would not want to create folder-names that contain these possibly weird symbols in all cases.

Why I thought referencing everything by index makes sense, is because we wanted to move the sequence definitions into the database_config anyways, where we would also just have a list of sequences, whose position would determine the index. As I wanted to not blow up the PR again, I first took the indexes from the reference_genomes

src/silo/test/amino_acid_insertion_contains.test.cpp
Taepper: That is true, but I think we have this problem in many places where we create objects in-line

include/silo/storage/reference_genomes.h
Taepper: I introduced this change rather late while working on the PR. This cleaned up the code a lot in many places

src/silo/test/amino_acid_insertion_contains.test.cpp
fengelniederhammer: At least here the previous version looks a lot more intuitive :)

PR #534, opened 2024-08-05
feat: new operator for matching strings against regular expressions

### LINKED ISSUE #517: String search with regular expressions
We want to be able to filter String columns with regular expressions.

We will implement this with Google's RE2 library

Taepper:
resolves #517 

### Summary

Introduce a new Filter that filters String fields with regular expressions. Google's RE2 library is used for matching each string value against the user-provided regular expression

## PR Checklist

### DIFF
--- a/endToEndTests/test/queries/stringSearch_basic_regex.json
+++ b/endToEndTests/test/queries/stringSearch_basic_regex.json
@@ -1,0 +1,21 @@
+{
+  "testCaseName": "StringSearch with a basic regex",
+  "query": {
+    "action": {
+      "type": "Aggregated"
+    },
+    "filterExpression": {
+      "type": "Not",
+      "child": {
+        "type": "StringSearch",
+        "column": "gisaid_epi_isl",
+        "searchExpression": "EPI"
+      }
+    }
+  },
+  "expectedQueryResult": [
+    {
+      "count": 0
+    }
+  ]
+}
--- a/include/silo/query_engine/filter_expressions/string_search.h
+++ b/include/silo/query_engine/filter_expressions/string_search.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include <re2/re2.h>
+#include <nlohmann/json_fwd.hpp>
+
+#include "silo/query_engine/filter_expressions/expression.h"
+
+namespace silo {
+class Database;
+class DatabasePartition;
+namespace query_engine {
+namespace operators {
+class Operator;
+}  // namespace operators
+}  // namespace query_engine
+}  // namespace silo
+
+namespace silo::query_engine::filter_expressions {
+
+class StringSearch : public Expression {
+  private:
+   std::string column;
+   std::unique_ptr<re2::RE2> search_expression;
+
+  public:
+   explicit StringSearch(std::string column, std::unique_ptr<re2::RE2> search_expression);
+
+   std::string toString() const override;
+
+   [[nodiscard]] std::unique_ptr<silo::query_engine::operators::Operator> compile(
+      const Database& database,
+      const DatabasePartition& database_partition,
+      AmbiguityMode mode
+   ) const override;
+};
+
+// NOLINTNEXTLINE(readability-identifier-naming)
+void from_json(const nlohmann::json& json, std::unique_ptr<StringSearch>& filter);
+
+}  // namespace silo::query_engine::filter_expressions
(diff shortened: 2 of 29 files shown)


include/silo/query_engine/filter_expressions/string_search.h
```diff
@@ -0,0 +1,46 @@
...
+class Operator;
+}  // namespace operators
+}  // namespace query_engine
+}  // namespace silo
+
+namespace silo::query_engine::filter_expressions {
+
+class StringSearch : public Expression {
+  private:
+   std::string column;
```
pflanze, line 28: `column_name`

include/silo/query_engine/filter_expressions/string_search.h
```diff
@@ -0,0 +1,46 @@
...
+   std::unique_ptr<re2::RE2> search_expression;
+
+  public:
+   explicit StringSearch(std::string column, std::unique_ptr<re2::RE2> search_expression);
```
pflanze, line 32: `column_name`?

src/silo/query_engine/filter_expressions/expression.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <string>
 
+#include <re2/re2.h>
```
pflanze, line 5: Why is this include needed *here*?

src/silo/query_engine/filter_expressions/string_search.cpp
```diff
@@ -0,0 +1,113 @@
...
+      row_count
+   );
+}
+
+}  // namespace
+
+std::unique_ptr<silo::query_engine::operators::Operator> StringSearch::compile(
+   const silo::Database& /*database*/,
+   const silo::DatabasePartition& database_partition,
+   Expression::AmbiguityMode /*mode*/
```
pflanze, line 61: Is this an unused parameter of a shared interface, or forgotten, or for the future? (Add comment?)

pflanze (changes requested):
Looks good except for the mentioned points.

endToEndTests/test/queries/stringSearch_basic_regex.json
```diff
@@ -0,0 +1,21 @@
...
+  "query": {
+    "action": {
+      "type": "Aggregated"
+    },
+    "filterExpression": {
+      "type": "Not",
+      "child": {
+        "type": "StringSearch",
+        "column": "gisaid_epi_isl",
+        "searchExpression": ".*EPI.*"
```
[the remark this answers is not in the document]
pflanze: Some statement that matches are rare, but one does exist in Java: "Yes, [Java's] matches() always acts as if the regex were anchored at both ends. To get the traditional behavior, which is to match any substring of the target, you have to use find() (as others have already pointed out). Very few regex tools offer anything equivalent to Java's matches() methods, so your confusion is justified. The only other one I can think of offhand is the [XML Schema](http://www.regular-expressions.info/xml.html) flavor."

https://stackoverflow.com/questions/1894624/is-regex-in-java-anchored-by-default-with-both-a-and-character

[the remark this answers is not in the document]
pflanze: Other thoughts:

* I think just because SQL `LIKE` implicitly anchors (and doesn't have anchoring syntax) is not enough of a reason to deviate from the usual approach with regular expressions. You could deviate, but I think you need another reason to justify it, and we should state the implicit anchoring in the docs with emphasis, and that reason.

* If it is to be made configurable, I suggest to do it in the following way:

    * In the frontend, have a checkbox *right next* to the entry field for the regex; have it *checked* for implicit anchoring (it *adds* something to normal regex behaviour; it could be checked by default if deemed right, though).
    * Instead of adding two modes to the backend, if the checkbox is checked, have the frontend surround the regex with `^` and `$` before sending the query to the backend. (Because it simplifies the backend and protocol, and the "mode" is always directly kept with the regex, meaning if e.g. the regex is printed to a log file, that "mode" is automatically part of it.) Or, being more explicit to the user, have an "anchor" button that inserts `^` and `$` visibly into the regex input field.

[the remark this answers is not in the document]
Taepper: After some considerations, also looking at similar implementations, we decided for unanchored regexes by default.

src/silo/query_engine/filter_expressions/expression.cpp
Taepper: This is related to your point raised in #536, this include is needed for the `make_unique` of a `StringSearch` object. As the header file only forward declares the re2, to which it owns a unique_ptr, we need to be able to resolve a `sizeof(RE2)` in this file. This is changed including `<re2/re2.h>` in `string_search.h`

pflanze (approved):
This should be documented later in whatever place user docs go.

PR #535, opened 2024-08-06
Update to Clang tidy 20

Taepper:
## PR Checklist


pflanze (approved):
LGTM.

pflanze:
As discussed, the plan is to go with this as long as clang-format can be kept at version 17. If that doesn't work anymore, either revert the change to clang-tidy 20, or figure out how to configure clang-format ~exhaustively so that changing clang-format version does not yield formatting changes.

Discussion of the latter: https://stackoverflow.com/questions/59395507/how-to-maintain-a-clang-format-file-for-different-clang-format-versions
Problematic point re clangd: https://github.com/clangd/clangd/issues/1999

Could potentially also generate the .clang-format file from our explicit options and public metadata about the options accepted by the newest version of clang-format; but hopefully someone has already written such a generator, research this.

PR #549, opened 2024-08-28
Missing unaligned sequences bug

Taepper:
resolves #

## PR Checklist


src/silo/query_engine/actions/fasta.cpp
```diff
@@ -128,6 +129,10 @@ void addSequencesFromResultTableToJson(
       const size_t end_of_partition_in_result = results.size();
       for (size_t idx = start_of_partition_in_result; idx < end_of_partition_in_result; idx++) {
          auto current_key = table_reader.next(genome_buffer);
+         if (!current_key.has_value()) {
+            throw std::runtime_error("Internal Error");
```
fengelniederhammer, line 133: Some more details in the error message would be good (even if it's just for the developer reading this to understand why we throw an error here)

fengelniederhammer (approved):
LGTM

Issue #395, opened 2024-04-24
Verify that the temp folder is empty for preprocessing


Taepper:
This currently can lead to a bug, that *may* lead to wrong results for unaligned sequences endpoint.

This is caused by some sequences being present multiple times in the hive-partitioned files for unaligned sequences, which are copied from the temp folder to the output folder. If the temp folder is empty before running the preprocessing this error does not happen

Taepper:
Testing that the error no longer occurs and upgrading should also fix this

Taepper:
duckdb fix is not yet in the latest release.

Check and error if number of sequences in the temp folder is not as expected

Taepper:
The duckdb update should fix this. We do not want to delete (possibly misspecified) folders

PR #552, opened 2024-08-30
fix: manually overwrite already existing unaligned sequence files in the temp folder

Taepper:
### Summary

In case the directory already exists when running preprocessing, it is manually deleted as the currently used duckdb version does not correctly overwrite the files and may leave duplicate entries in the output

## PR Checklist

### DIFF
--- a/src/silo/preprocessing/preprocessor.cpp
+++ b/src/silo/preprocessing/preprocessor.cpp
@@ -555,6 +555,24 @@ void Preprocessor::createUnalignedPartitionedSequenceFile(
    const std::filesystem::path save_location =
       preprocessing_config.getIntermediateResultsDirectory() /
       fmt::format("unaligned_nuc_{}", sequence_idx);
+   // duckdb OVERWRITE and OVERWRITE_OR_IGNORE is broken in the current version,
+   // therefore we manually delete the save_location directory in case it already exists
+   if (std::filesystem::exists(save_location)) {
+      if (!std::filesystem::is_directory(save_location)) {
+         const std::string error_message = fmt::format(
+            "The temp directory contains an erroneous file {}. Delete this file then rerun "
+            "preprocessing.",
+            save_location.string()
+         );
+         SPDLOG_ERROR("{}", error_message);
+         throw silo::preprocessing::PreprocessingException(error_message);
+      }
+      SPDLOG_INFO(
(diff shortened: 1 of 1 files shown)


fengelniederhammer:
Linking #395, as this is another occurence of the same problem.

fengelniederhammer (approved):

src/silo/preprocessing/preprocessor.cpp
```diff
@@ -555,6 +555,24 @@ void Preprocessor::createUnalignedPartitionedSequenceFile(
    const std::filesystem::path save_location =
       preprocessing_config.getIntermediateResultsDirectory() /
       fmt::format("unaligned_nuc_{}", sequence_idx);
+   // duckdb OVERWRITE and OVERWRITE_OR_IGNORE is broken in the current version,
```
corneliusroemer, line 558: It would be very cool if you could:
1. add the link to the duckdb issue in a code comment here, or if none exists open one
2. include a link to the duckdb issue in the PR description
3. explicitly state which is the current duckdb version, and since approximately which duckdb version this bug is present in duckdb

corneliusroemer: Why don't we we pin duckdb to the highest non-bugged version?

corneliusroemer (commented):
We should create an issue for this bug - so one can link to it.

src/silo/preprocessing/preprocessor.cpp
Taepper: I added some info, including the link to the [change log](https://github.com/GenSpectrum/LAPIS-SILO/blob/main/CHANGELOG.md). Unfortunately, I saw that it did not appear in the release notes. Updating duckdb is not always trivial, esp. in the versions before 1.0.0 the build script often changed and the recipes in the package manager we use (conan) needed to be changed [multiple](https://github.com/conan-io/conan-center-index/commit/ba3fe9f0ee53b80620e74acdab461c283d03a145) [times](https://github.com/conan-io/conan-center-index/commit/0b2998d329462c48b561abfdd2b6b9f155a8f58a), given that we also use some duckdb extensions, which the maintainer of that recipe often does not check...

The most recent update was delayed, because I thought that #394 would make some of these migrations more deterministic, given that the query results of unsorted or randomly sampled queries often changed.

Issue #562, opened 2024-09-09
Remove tsv/fasta input format


Taepper:
Supporting the legacy input format with tsvs and fastas makes the code in the preprocessing a lot more complicated than it needs to be. We also have some complicated null logic because tsv does not have native support for nulls

fengelniederhammer:
Summary of the recent discussion:
* We want to remove the TSV/FASTA input format.
  * It is essentially unused. All real-world users use the NDJSON input format.
  * The preprocessing code will be a lot easier to maintain, since there will only be one major execution path.
  * NDJSON offers proper handling of null values (TSV requires the workaround of using empty strings)
* We won't do it now
  * If there are any bigger changes in the preprocessing code, then we should first do this issue
    * currently there are none planned.
  * Ideally we can wait until after the Rust experiment:
    * if successful, we can simply delete the C++ part altogether
    * if not successful, we only remove the TSV/FASTA execution path from the C++ code
* We will offer a script / CLI tool to convert the TSV/FASTA input to NDJSON instead.
  * Thus, we still "support" that input format, because it's actually useful. Users are more likely to have TSV and FASTA files instead of NDJSON.

PR #569, opened 2024-09-12
fix: validate that either a ndjson or tsv file was given as input

### LINKED ISSUE #564: Preprocessing errors with unclear/unactionable error: "duckdb threw with error: Parser Error: unterminated quoted string at or near "'pragma_collations""
I'm working on the silo import script in loculus.

Things are working ok until silo throws a (to the end user) unclear error:

```
Preprocessing exception when retrieving the fields of the metadata file 'pragma_collations', duckdb threw with error: Parser Error: unterminated quoted string at or near "'pragma_collations"
LINE 1: SELECT * FROM read_csv_auto('pragma_collations\0\0\0', delim = '	', header...
```

What's the end-user-understandable version of this error? I suppose this is more something I'm doing wrong in the data I'm passing as opposed to a silo bug. But there should still a) either be validation and an early informative error to the user or b) the exception should be wrapped into something more end-user-understandable.

Repro:

Run this in bash
```bash
gh repo clone loculus-project/loculus
cd loculus
git checkout origin/unclear-silo-error
cd kubernetes/loculus
bash error.sh
```

You can also attach then directly to the silo container:

```
docker exec -it $CONTAINER_NAME /bin/bash
```

And e.g. inside run `/app/siloApi -p`

Taepper:
resolves #564 

### Summary

We did not check that either a ndjson or tsv file is actually specified as the input. 

The resulting duckdb error is not parsable for users. Instead we now return a proper error message, also hinting at the possible CLI arguments that can be given instead of amending the preprocessing_config

## PR Checklist

### DIFF
--- a/src/silo/config/preprocessing_config.cpp
+++ b/src/silo/config/preprocessing_config.cpp
@@ -14,11 +14,19 @@ void PreprocessingConfig::validate() const {
    if (!std::filesystem::exists(input_directory)) {
       throw preprocessing::PreprocessingException(input_directory.string() + " does not exist");
    }
-   if (ndjson_input_filename.has_value() && metadata_file) {
+   if (ndjson_input_filename.has_value() && metadata_file.has_value()) {
       throw preprocessing::PreprocessingException(fmt::format(
          "Cannot specify both a ndjsonInputFilename ('{}') and metadataFilename('{}').",
          ndjson_input_filename.value().string(),
          metadata_file.value().string()
+      ));
+   }
+   if (!ndjson_input_filename.has_value() && !metadata_file.has_value()) {
+      throw preprocessing::PreprocessingException(fmt::format(
+         "Neither a ndjsonInputFilename ('{}') nor a metadataFilename ('{}') was specified as "
+         "preprocessing option.",
+         NDJSON_INPUT_FILENAME_OPTION.toCamelCase(),
+         METADATA_FILENAME_OPTION.toCamelCase()
       ));
    }
 }
@@ -211,7 +219,7 @@ void PreprocessingConfig::overwrite(const silo::config::AbstractConfigSource& co
       ctx.out(),
       "{{ input directory: '{}', pango_lineage_definition_file: {}, output_directory: '{}', "
       "metadata_file: '{}', reference_genome_file: '{}',  gene_file_prefix: '{}',  "
-      "nucleotide_sequence_file_prefix: '{}', unalgined_nucleotide_sequence_file_prefix: '{}', "
+      "nucleotide_sequence_file_prefix: '{}', unaligned_nucleotide_sequence_file_prefix: '{}', "
       "ndjson_filename: {}, "
       "preprocessing_database_location: {} }}",
       preprocessing_config.input_directory.string(),


src/silo/config/preprocessing_config.cpp
```diff
@@ -14,13 +14,21 @@ void PreprocessingConfig::validate() const {
...
          metadata_file.value().string()
       ));
    }
+   if (!ndjson_input_filename.has_value() && !metadata_file.has_value()) {
+      throw preprocessing::PreprocessingException(fmt::format(
+         "Neither a ndjsonInputFilename ('{}') nor a metadataFilename('{}') was specified in the "
+         "preprocessing config or via the --metadataFilename / --ndjsonInputFilename option.",
+         ndjson_input_filename.value().string(),
+         metadata_file.value().string()
+      ));
```
fengelniederhammer, line 30: There is another access to undefined options in there ;) 

I also would not mention how to do it, since it might change (and the current version forgot environment variables). Users should simply consult the documentation (maybe that's worth adding to the error message).

```suggestion
      throw preprocessing::PreprocessingException(fmt::format(
         "Neither a {} nor a {} was specified.",
         METADATA_FILENAME_OPTION.toString(),
         NDJSON_INPUT_FILENAME_OPTION.toString()
      ));
```

fengelniederhammer (changes requested):

src/silo/config/preprocessing_config.cpp
```diff
@@ -14,13 +14,21 @@ void PreprocessingConfig::validate() const {
...
+   if (ndjson_input_filename.has_value() && metadata_file.has_value()) {
       throw preprocessing::PreprocessingException(fmt::format(
          "Cannot specify both a ndjsonInputFilename ('{}') and metadataFilename('{}').",
          ndjson_input_filename.value().string(),
          metadata_file.value().string()
       ));
    }
+   if (!ndjson_input_filename.has_value() && !metadata_file.has_value()) {
+      throw preprocessing::PreprocessingException(fmt::format(
+         "Neither a ndjsonInputFilename ('{}') nor a metadataFilename ('{}') was specified as "
```
corneliusroemer, line 26: Here you could list the various ways in which they can be passed, but not super important (CLI, env, config file)

corneliusroemer (commented):
How is it possible that duckdb threw such a cryptic/mojibake error?

This looks a bit like out of bounds access, which isn't great.

Validating is good, but I wonder how the error could end up so mangled. Can you point me to the place where duckdb crashed/threw?

corneliusroemer:
If there is a bug in the CLI option part, might make sense to add some basic tests here (E2E, once you get to it :) )

this stuff should be super easy to test for regressions

src/silo/config/preprocessing_config.cpp
Taepper: I removed it after the above comment.

We have them described in the [README](https://github.com/GenSpectrum/LAPIS-SILO?tab=readme-ov-file#configuration-files). Although the part about default values and the configs therefore being optional is unfortunately no longer valid as of [my refactoring here](https://github.com/GenSpectrum/LAPIS-SILO/commit/847ec7e8af6eda73c95076c9b4428ed3fd5c3f9f#diff-b2ac6b3e28c6241fa5fce2e67f54fd37358cd2a7e7e0f3c5becab39bd4cf8015L39)...

Taepper:
> Can you point me to the place where duckdb crashed/threw?

The actual part that was broken / where the gibberish was generated is an unchecked optional access [here](https://github.com/GenSpectrum/LAPIS-SILO/blob/944f209c0bea902a85337fda2b92b65fe537159b/src/silo/preprocessing/preprocessor.cpp#L114). Note that the `->` symbol implicitly calls `.value()` on the optional.

I will add comments / additional action points to https://github.com/GenSpectrum/LAPIS-SILO/issues/562

corneliusroemer:
I see, and how come it produced mojibake? Because nullpointer access and C++ being unsafe language? I've only done stuff like that in Rust and there it would throw instead of producing gibberish.

Taepper:
I also thought it would throw an exception (which would still puzzle the end-user) as is documented for the `value` function [here](https://en.cppreference.com/w/cpp/utility/optional/value).

But [it turns out](https://en.cppreference.com/w/cpp/utility/optional/operator*) the shorthand syntax is just undefined behaviour :(

fengelniederhammer:
Probably this was the evil line?
https://github.com/GenSpectrum/LAPIS-SILO/blob/c34fa1f18e8985b4f2b10cec57c28fd68fa76b02/src/silo/preprocessing/preprocessor.cpp#L116

PR #571, opened 2024-09-12
Panic utilities improvements: ensure DEBUG_ASSERT compiles, add ASSERT_EQ etc., add tests

### LINKED ISSUE #559: More panic / assertment features
There could be some additional features provided by panic.h:

* ASSERT_EQ(a, b)
* allow format string and arguments for UNREACHABLE / UNIMPLEMENTED: UNREACHABLE("because {} is too small and hence it can't happen that ...", x)

Both of these are available in Rust, thus would make sense to have.

pflanze:
resolves #559

### Summary

- ensure that DEBUG_ASSERT's arguments compile: even someone who just compiles in release mode will see if their test expressions in DEBUG_ASSERT(test-expression) compile
- add ASSERT_EQ, ASSERT_LT etc.
- add tests for the important features (including panic features)

I didn't add tests before because I thought it would be necessary to test them (in particular the abort feature) via a script from outside which would have been painful; but turns out gtest has ASSERT_DEATH which allows to do this perfectly.

This PR still doesn't extend UNREACHABLE / UNIMPLEMENTED for arguments, which are a bit challenging to do (the 0-ary case does not need to call fmt::format at all, and CPP macros have no easy dispatch on the number of arguments, but I think I know how to do it anyway now; but will only approach this when asked).

## PR Checklist

- [ x] All necessary documentation has been adapted or there is an issue to do so.
- [ x] The implemented feature is covered by an appropriate test.

### DIFF
--- a/include/silo/common/panic.h
+++ b/include/silo/common/panic.h
@@ -70,21 +70,23 @@ namespace silo::common {
 #ifndef DEBUG_ASSERTIONS
 #warning \
    "DEBUG_ASSERTIONS is not set, should be 0 to ignore DEBUG_ASSERT, 1 to compile it in, assuming 0"
-#define DEBUG_ASSERT(e)
-#else                     // DEBUG_ASSERTIONS is defined
-#if DEBUG_ASSERTIONS == 0 /* never */
-#define DEBUG_ASSERT(e)
+#define DEBUG_ASSERTIONS 0
+#else
+#if DEBUG_ASSERTIONS == 0   /* never */
 #elif DEBUG_ASSERTIONS == 1 /* always */
-#define DEBUG_ASSERT(e)                                            \
-   do {                                                            \
-      if (!(e)) {                                                  \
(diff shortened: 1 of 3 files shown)


include/silo/common/panic.h
```diff
@@ -70,22 +115,50 @@ namespace silo::common {
 #ifndef DEBUG_ASSERTIONS
 #warning \
    "DEBUG_ASSERTIONS is not set, should be 0 to ignore DEBUG_ASSERT, 1 to compile it in, assuming 0"
-#define DEBUG_ASSERT(e)
-#else                     // DEBUG_ASSERTIONS is defined
-#if DEBUG_ASSERTIONS == 0 /* never */
-#define DEBUG_ASSERT(e)
+#define DEBUG_ASSERTIONS 0
+#else
```
Taepper, line 119: should this be an endif?

Taepper (approved):
LGTM

PR #576, opened 2024-09-16
refactor: remove forward declarations

### LINKED ISSUE #536: Rethink forward declarations
The [google style guide](https://google.github.io/styleguide/cppguide.html#Forward_Declarations) recommends to avoid forward-declarations whenever possible.

We should measure the change in compile time using forward-declarations vs. always including header-files and decide based on the results

Taepper:
resolves #536

### Summary

Following the [google style guide](https://google.github.io/styleguide/cppguide.html#Forward_Declarations) this PR removes all forward declarations that can be replaced with an include directive

## PR Checklist

### DIFF
--- a/include/silo/common/bidirectional_map.h
+++ b/include/silo/common/bidirectional_map.h
@@ -7,12 +7,10 @@
 #include <unordered_map>
 #include <vector>
 
+#include <boost/serialization/access.hpp>
+
 #include "silo/common/pango_lineage.h"
 #include "silo/common/types.h"
-
-namespace boost::serialization {
-class access;
-}  // namespace boost::serialization
 
 namespace silo::common {
 
--- a/include/silo/common/string.h
+++ b/include/silo/common/string.h
@@ -8,13 +8,9 @@
 #include <optional>
 #include <string>
 
-namespace boost::serialization {
-class access;
-}  // namespace boost::serialization
-namespace silo::common {
-template <typename V>
-class BidirectionalMap;
-}  // namespace silo::common
+#include <boost/serialization/access.hpp>
+
+#include "silo/common/bidirectional_map.h"
 
 namespace silo::common {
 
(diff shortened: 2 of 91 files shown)


src/silo/query_engine/filter_expressions/symbol_equals.cpp
```diff
@@ -137,9 +133,7 @@ std::unique_ptr<silo::query_engine::operators::Operator> SymbolEquals<SymbolType
       );
       return std::make_unique<operators::Complement>(
          std::make_unique<operators::IndexScan>(
-            std::move(logical_equivalent_of_nested_index_scan),
```
fengelniederhammer, line 140: `logical_equivalent_of_nested_index_scan` is now unused.

Why is it ok that we don't pass that argument to `IndexScan` now? AFAICS this is debug output, but maybe valuable?

src/silo/query_engine/filter_expressions/symbol_equals.cpp
```diff
@@ -172,9 +166,7 @@ std::unique_ptr<silo::query_engine::operators::Operator> SymbolEquals<SymbolType
    auto logical_equivalent =
       std::make_unique<SymbolEquals>(valid_sequence_name, position_idx, symbol);
    return std::make_unique<operators::IndexScan>(
-      std::move(logical_equivalent),
```
fengelniederhammer, line 175: `logical_equivalent` is now unused.

fengelniederhammer (approved):

src/silo/query_engine/filter_expressions/symbol_equals.cpp
Taepper: Ahh, I temporarily removed this (due to cyclic dependency) but wanted to insert it back in, thanks for catching it!

PR #590, opened 2024-09-19
feat: generalized wildcard queries, first finished draft

### LINKED ISSUE #458: Generalize wildcard lineage queries: Support more general hierarchies
### Motivation

@Taepper, @chaoran-chen and I have discussed generalizing the Pango wildcard search feature, as monophyletic clade search is important (non-wildcard searches are [paraphyletic](https://en.wikipedia.org/wiki/Paraphyly), which is usually not what one wants to do). In fact, if it was up to me, I would make wildcard/monophyletic search the default behaviour and make paraphyletic searches opt in, e.g. through `!` or maybe `^` operator instead of requiring `*`.

The current hierarchy feature is very restrictive: it requires a grammar of `ALIAS[.DECIMALNUMBER]{0,3}`

Many viruses have Pango-like lineage systems, but almost none use the exact spec according to the grammar above.

For example (using a semi-pseudo-regex-style-grammar):

- mpox: `ALIAS.[.DECIMALNUMBER]{0,2}` with `ALIAS=[ALPHABETICALCHARACTER]+`, i.e. like SARS-CoV-2, but with aliasing after 2 rather than 3 dot-number levels.
- RSV: `SUBTYPE[.ALIAS]*[.DECIMALNUMBER]*`
- flu: `ALIAS[.DECIMALNUMBER]*`

### Goal

Generalize the existing hierarchical search feature to be compatible with the above grammars.

### Proposal spec

The simplest way to encompass general monophyletic search on a tree of lineages is to not bother with the grammars at all, and simply treat lineage names as string labels, that are related in a (multi-)tree-like fashion. Initially, we probably want to think about these as a DAG (directed acyclical (multi)graph).

To take away the grammar complexity from SILO/LAPIS and work with all DAGS, all we need is a simple edge list as input config file. Done!

These are easy to generate programatically from the pango_alias.json file, and others, they'd be a bit more verbose, but that doesn't matter as a few thousand edges are not a problem at all for computers.

The simplest way to do it is (I'm using yaml for ease of reading/writing, but this could well be JSON):

```yaml
LINEAGE_LABEL:
  parent: PARENT_LABEL
  aliases: 
  - ALIAS_LABEL1
  - ALIAS_LABEL2
  other_metadata (optional): dict
```

Example for a subset of Pango:

```yaml
B:
  parent: null
  aliases: null
  other_metadata:
    designation_date: 2020-04-01
B.1:
  parent: B
  aliases: null
B.1.1:
  parent: B.1
  aliases: null
B.1.1.529:
  parent: B.1.1
  aliases: 
  - BA
BA.2:
  parent: B.1.1.529
  aliases:
  - B.1.1.1.529.2
BA.2.86:
  parent: BA.2
  aliases:
  - B.1.1.529.2.86
BA.2.86.1:
  parent: BA.2.86
  aliases:
  - JN
  - B.1.1.529.2.86
JN.1:
  parent: BA.2.86.1
  aliases:
  - BA.2.86.1.1
  - B.1.1.529.2.86.1.1
JN.1.11:
  parent: JN.1
  aliases:
  - BA.2.86.1.1.11
  - B.1.1.529.2.86.1.1.11
JN.1.11.1:
  parent: JN.1.11
  aliases:
  - KP
  - BA.2.86.1.1.11.1
  - B.1.1.529.2.86.1.1.11.1
KP.3:
  parent: JN.1.11.1
  aliases:
  - JN.1.11.1.3
  - BA.2.86.1.1.11.1.3
  - B.1.1.529.2.86.1.1.11.1.3
...
```

I think you get the point. Now one interesting possible extension is to include recombinants not as their own trees but give them two parents. One could then offer an _alternative_ wildcard query mode that treats a recombinant as a parent of _both_ of their parents (but only counting them once if both parents are included in a query).

The recombinant multi-parent mode would be cool, but is by far not as important as supporting general schemes.

I'm happy to provide further examples, e.g. for mpox, it's quite trivial to automate these.

Taepper:
resolves #458 

### Summary

This PR resolves the generalized lineage refactor. Currently the system only worked with lineages following the pango lineage format as it was designed for SARS-CoV-2.

Now this is generalized and just receives an arbitrary lineage definition file which outlines the parent-child relations for the provided lineages.

Breaking Changes:
- The preprocessing config fields has changed from pangoLineageDefinitionFilename to lineageDefinitionFilename
- We take a yaml lineage definition file instead of a pango alias key, a script is provided `scripts/alias2lineageDefinitions.py` that transforms a list of lineages and an alias into the requested file format
- We now validate on input and on query whether the provided lineage is in the defined lineages and throw errors accordingly

In the current version, we do not allow alias queries. This means that the user-provided query needs to e.g. contain a valid pango lineage for the respective instances. Searching for `B.1.1.529.1` does not work and instead a search for `BA.1` is required

## PR Checklist


Taepper:
By the way, the majority of the line-count is the new (more-verbose) lineage definition file: [testBaseData/exampleDataset/lineage_definitions.yaml](https://github.com/GenSpectrum/LAPIS-SILO/pull/590/files#diff-50fa23172ea10cff149c18ef4141afef670a8f5139cd68aeae9bbe47112b2680)

include/silo/storage/column/indexed_string_column.h
```diff
@@ -12,30 +12,39 @@
 #include <roaring/roaring.hh>
 
 #include "silo/common/bidirectional_map.h"
+#include "silo/common/lineage_tree.h"
 #include "silo/common/types.h"
+#include "silo/storage/lineage_index.h"
 
 namespace silo::storage::column {
 
+class IndexedStringColumn;  // TODO comment
```
fengelniederhammer, line 21: Let's do it :)

src/silo/storage/column/indexed_string_column.cpp
```diff
@@ -50,4 +81,12 @@ IndexedStringColumnPartition& IndexedStringColumn::createPartition() {
    return partitions.emplace_back(*lookup);
 }
 
+void IndexedStringColumn::generateLineageIndex(const common::LineageTreeAndIDMap& lineage_tree) {
+   *lookup = lineage_tree.lineage_id_lookup_map.copy();
+   for (auto& partition : partitions) {
+      std::reference_wrapper(partition.lookup) = *lookup;
+      partition.lineage_index = LineageIndex(lineage_tree.lineage_tree);
```
fengelniederhammer, line 88: Could you please explain why we need this reference wrapper here?

src/silo/config/util/config_repository.cpp
```diff
@@ -81,11 +86,6 @@ void validatePartitionBy(
    if (metadata_map.find(partition_by) == metadata_map.end()) {
       throw ConfigException("partition_by '" + partition_by + "' is not in metadata");
    }
-
-   const auto& partition_by_type = metadata_map[partition_by];
-   if (partition_by_type != ValueType::PANGOLINEAGE) {
-      throw ConfigException("partition_by '" + partition_by + "' must be of type PANGOLINEAGE");
-   }
```
fengelniederhammer, line 88: That means that we allow partitioning by arbitrary columns now?

src/silo/config/util/config_repository.cpp
```diff
@@ -26,22 +26,27 @@ std::map<std::string, ValueType> validateMetadataDefinitions(const DatabaseConfi
...
-            "PANGOLINEAGE"
+            "' lineage_index is set, generate_index must also be set."
+         );
+      }
+
+      const auto must_be_string = metadata.lineage_index;
+      if (metadata.type != ValueType::STRING && must_be_string) {
+         throw ConfigException(
+            "Metadata '" + metadata.name +
+            "' lineage_index is set, but the column must be of type STRING."
```
fengelniederhammer, line 49:
```suggestion
      const auto must_be_string = metadata.lineage_index;
      if (metadata.type != ValueType::STRING && must_be_string) {
         throw ConfigException(
            "Metadata '" + metadata.name +
            "' lineage_index is set, but the column is not of type STRING."
```

fengelniederhammer, line 49: This statement is effectively never hit, because the other two statements above already catch it -> Let's move it to the top.

src/silo/config/util/config_repository.cpp
```diff
@@ -26,22 +26,27 @@ std::map<std::string, ValueType> validateMetadataDefinitions(const DatabaseConfi
...
-            "' generate_index is set, but generating an index is only allowed for types STRING and "
-            "PANGOLINEAGE"
+            "' generate_index is set, but generating an index is only allowed for types STRING"
          );
       }
 
-      const auto must_generate_index_on_type = metadata.type == ValueType::PANGOLINEAGE;
-      if (!metadata.generate_index && must_generate_index_on_type) {
+      const auto must_generate_index = metadata.lineage_index;
+      if (!metadata.generate_index && must_generate_index) {
```
fengelniederhammer, line 38: This case is not covered by a unit test

src/silo/config/util/config_repository.cpp
```diff
@@ -26,22 +26,27 @@ std::map<std::string, ValueType> validateMetadataDefinitions(const DatabaseConfi
...
          throw ConfigException(
             "Metadata '" + metadata.name +
-            "' generate_index is not set, but generating an index is mandatory for type "
-            "PANGOLINEAGE"
+            "' lineage_index is set, generate_index must also be set."
+         );
+      }
+
+      const auto must_be_string = metadata.lineage_index;
+      if (metadata.type != ValueType::STRING && must_be_string) {
```
fengelniederhammer, line 46: This case is not covered by a unit test

src/silo/database.cpp
```diff
@@ -656,6 +647,10 @@ void Database::initializeColumn(config::ColumnType column_type, const std::strin
 void Database::initializeColumns() {
    for (const auto& item : database_config.schema.metadata) {
       initializeColumn(item.getColumnType(), item.name);
+      if (item.lineage_index) {
+         ASSERT(columns.indexed_string_columns.contains(item.name));
+         columns.indexed_string_columns.at(item.name).generateLineageIndex(lineage_tree);
+      }
```
fengelniederhammer, line 653: Why don't we make this part of `initalizeColumn`, in the `case config::ColumnType::INDEXED_STRING`? Then we also don't need the `ASSERT`, since the referenced value is created there.

fengelniederhammer (changes requested):
The commit message should reference the issue (ideally with `resolves #...` in the footer - then release please will automatically pick it up and mention it in the changelog)

Also this is a breaking change - the commit message should mention it.

Also I wonder whether it might be good to still have a metadata type `lineage` or similar? Would there be any advantages?

include/silo/common/lineage_tree.h
```diff
@@ -0,0 +1,71 @@
...
+#include <vector>
+
+#include <boost/serialization/access.hpp>
+
+#include "silo/common/bidirectional_map.h"
+#include "silo/preprocessing/lineage_definition_file.h"
+
+namespace silo::common {
+
+// The tree is allowed to be disconnected
```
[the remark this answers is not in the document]
Taepper: Yes, exactly. We could argue that we should then not call it a tree anymore, but I still find it more intuitive. Added the comment for the extra clarification..

include/silo/common/bidirectional_map.h
```diff
@@ -30,14 +29,45 @@ class BidirectionalMap {
...
+
+   Idx getOrCreateId(V value) {
+      if (value_to_id.contains(value)) {
+         return value_to_id.at(value);
+      }
+      const Idx identifier = id_to_value.size();
+      id_to_value.push_back(value);
+      value_to_id[value] = identifier;
+      return identifier;
+   }
```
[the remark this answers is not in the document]
Taepper: As this is templated code, it will not lead to duplicate symbols and has the additional advantage that the compiler will be helped at inlining :)
This follows the [google style guide](https://google.github.io/styleguide/cppguide.html#Self_contained_Headers):
> When a header declares inline functions or templates that clients of the header will instantiate, the inline functions and templates must also have definitions in the header, either directly or in files it includes.

Note that definition=implementation

testBaseData/exampleDataset/small_metadata_set.tsv
```diff
@@ -12,7 +12,7 @@ EPI_ISL_1749960	B.1.1.7	2021-04-15	Europe	Switzerland	Basel-Land	2021-02-03	59	0
 EPI_ISL_1361468	B.1.1.7	2021-03-06	Europe	Switzerland	Zürich	2021-01-20	50	0.98	
 EPI_ISL_1408062	B.1.1.7	2021-03-03	Europe	Switzerland	Valais	2020-11-24	50	0.97	false
 EPI_ISL_1597890	B.1.1.7	2021-03-21		Switzerland	Vaud	2021-01-25	51	0.96	true
-EPI_ISL_1682849	XA.1	2020-12-17	Europe	Switzerland	Thurgau	2021-01-21	52	0.95	false
```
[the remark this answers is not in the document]
Taepper: The lineage before was purely fictional and did not exist, now we return an error when preprocessing the previous state

src/silo/preprocessing/lineage_definition_file.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+LineageDefinitionFile LineageDefinitionFile::fromYAML(const std::string& yaml_string) {
+   YAML::Node yaml = YAML::Load(yaml_string);
+   LineageDefinitionFile file;
+   YAML::convert<LineageDefinitionFile>::decode(yaml, file);
+   return file;
+}
+
+LineageDefinitionFile LineageDefinitionFile::fromYAMLFile(const std::filesystem::path& yaml_path) {
+   std::ifstream file(yaml_path, std::ios::in | std::ios::binary);  // Open file in binary mode
```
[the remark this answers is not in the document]
Taepper: What exactly?

src/silo/common/lineage_tree.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+   UnionFind(size_t n) {
+      parent.resize(n);
+      rank.resize(n, 0);
+      for (uint32_t i = 0; i < n; ++i) {
+         parent[i] = i;
+      }
+   }
+
+   int find(uint32_t u) {
```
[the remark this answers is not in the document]
Taepper: I think UnionFind is very standardized and it is always called `find`

src/silo/common/lineage_tree.cpp
```diff
@@ -0,0 +1,155 @@
...
+         parent[rootV] = rootU;
+         rank[rootU]++;
+      }
+      return true;
+   }
+};
+
+bool containsCycle(int n, const std::vector<std::pair<Idx, Idx>>& edges) {
+   UnionFind uf(n);
+   for (const auto& edge : edges) {
```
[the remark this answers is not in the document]
Taepper: This is actually undirected cycle detection, so it is an arbitrary order. I therefore named it vertex1 and vertex2 to not imply directed behavior

src/silo/database.cpp
Taepper: Because we then would need to pass the extra information from the metadata, but I will give it a shot

src/silo/storage/column/indexed_string_column.cpp
fengelniederhammer: If you think it's beneficial also as a comment in the code. But mostly for my understanding of those two lines here (a PR comment would be enough ;) ).

src/silo/common/lineage_tree.cpp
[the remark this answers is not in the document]
fengelniederhammer: I find it confusing, but ok, then let's stick to the standard.

src/silo/database.cpp
fengelniederhammer: We can simply pass the whole metadata struct?

Taepper: Yes, I did that now :)

scripts/alias2lineageDefinitions.py
fengelniederhammer, line 1: I find this quite complicated. Wouldn't it be easier to do it in plain Python? 
* You get better IDE support compared to when writing SQL
* We would not need to DuckDB dependency in Python

fengelniederhammer: ChatGPT gave me this, which almost works:

```python
import json
import argparse
import yaml
import shutil
import sys
import os
import tempfile

# Define the argument parser
parser = argparse.ArgumentParser(description="Process a JSON file and convert it to a DataFrame.")
parser.add_argument('alias_key', type=str, help='Path to the alias_key in JSON format')
parser.add_argument('lineage_file', type=str, help='Path to the input_file containing all lineages')
parser.add_argument('--preserve-tmp-dir', action='store_true', help='Preserve the temporary directory to keep the intermediate files')

# Parse the arguments
args = parser.parse_args()

# Load the JSON data from the file path provided as argument
with open(args.alias_key, 'r') as file:
    alias_key = json.load(file)

# Create a list to store the reformatted data
reformatted_data = []

# Loop through the JSON and format it as desired
for key, value in alias_key.items():
    if value and isinstance(value, str):
        reformatted_data.append({"name": key, "alias": value})
    else:
        reformatted_data.append({"name": key, "alias": key})

# Load the lineage data from the file path provided as argument
with open(args.lineage_file, 'r') as file:
    lineages = file.read().splitlines()

# Create a dictionary to store the alias mappings
alias_dict = {item['name']: item['alias'] for item in reformatted_data}

# Function to unalias a lineage
def unalias_lineage(lineage):
    parts = lineage.split('.')
    if parts[0] in alias_dict:
        parts[0] = alias_dict[parts[0]]
    return '.'.join(parts)

# Unalias all lineages
unaliased_lineages = {lineage: unalias_lineage(lineage) for lineage in lineages}

# Function to find the immediate parent lineage
def find_immediate_parent(lineage):
    if '.' in lineage:
        return lineage.rsplit('.', 1)[0]
    return None

# Create the lineage definitions
lineage_definitions = {}
for lineage, unaliased in unaliased_lineages.items():
    parent = find_immediate_parent(unaliased)
    if parent:
        parent_unaliased = unalias_lineage(parent)
        lineage_definitions[lineage] = {"parents": [parent_unaliased]}
    else:
        lineage_definitions[lineage] = {"parents": []}

# Transform the data into YAML
yaml.dump(lineage_definitions, sys.stdout, default_flow_style=False)

if args.preserve_tmp_dir:
    temp_dir = tempfile.mkdtemp()
    print(f"Temporary directory: {temp_dir}", file=sys.stderr)
else:
    temp_dir = tempfile.mkdtemp()
    shutil.rmtree(temp_dir)
```

scripts/generate_new_lineage_definitions.bash
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash 
+curl https://raw.githubusercontent.com/cov-lineages/pango-designation/refs/heads/master/pango_designation/alias_key.json > alias_key.json 
+curl https://raw.githubusercontent.com/cov-lineages/pango-designation/refs/heads/master/lineages.csv | tail -n +2 | cut -d',' -f2 > lineages.csv
```
fengelniederhammer, line 3: Why not simply download the whole file, save it as is and do the tedious parsing in Python instead of weird bash commands?

Taepper: What python file? This would necessitate another python script as the current transformation script has nothing to do with all the additional information contained in the downloaded file. So this would require this script to then call another python file.. I do not find that simpler than the invocation of 2 commands. What exactly would be your alternate suggestion?

fengelniederhammer: What does it do? It cuts off a few lines and it basically strips down the CSV to a single column.

IMO it's more understandable to do that in Python code. `alias2lineageDefinitions.py` reads the file anyway, so it can also do those two simple transformations.

But if you think it's easier, we can also leave it here.

scripts/alias2lineageDefinitions.py
Taepper: I think that the SQL based script is less error-prone but effectively equivalent. I really like that I can just look at the intermediate tables and see the precise output of every single step in the script. What would the finished alternative look like?

fengelniederhammer: I don't like that I need to look at SQL statements to understand the code. Probably I'm just a lot less used to SQL :D

The finished script would look quite similar I guess. I resolves an alias in a couple places where it should print the alias instead (A.1.2.3.4 instead of AA.4). I assume it's easy to fix.

scripts/generate_new_lineage_definitions.bash
Taepper: I did make a second script that does some generalized version of this :)

src/silo/common/lineage_tree.cpp
```diff
@@ -0,0 +1,242 @@
...
+   ASSERT_LT(vertex_to, number_of_vertices);
+   adjacency_list.at(vertex_from).emplace_back(vertex_to);
+}
+
+std::optional<std::vector<Idx>> Graph::findCycleWitnessFromStartVertex(
+   Idx start_vertex,
+   std::vector<bool>& visited
+) const {
+   std::vector<Idx> recursion_stack;
+   std::vector<bool> in_recursion_stack(number_of_vertices);
```
fengelniederhammer, line 59: Do we really need this? Usages can probably be replaced by `recursion_stack.contains()`?

I think we can neglect the time that this algorithm needs compared to the other expensive things that SILO does. IMO performance optimization is irrelevant here, I'd rather simplify the code.

scripts/generate_new_lineage_definitions.bash
fengelniederhammer, line 1: As mentioned - IMO we should remove this here before merging and add it to the servers repo instead.

fengelniederhammer, line 1: I think this script would also benefit from making it Python. Then we would not need to store files on the disc, but instead we could simply store values in variable and have functions that return them.

fengelniederhammer (changes requested):
Actually looks good, but it's not quite ready to be merged yet.

Also please make sure that the changelog contains the issue.

src/silo/common/lineage_tree.cpp
Taepper: The usual "cycle-finding" algorithm actually only has the `in_recursion_stack` and `visited` vectors. They are most commonly referred to as `black` and `grey` vertices. I added the recursion_stack to 1st defend against stack-overflows and more importantly 2nd to show the witness cycle in the error message.

I think it can actually become quite resource consuming. I added a testcase with only 16'000 lineages which go from 32ms to 1.5seconds user-time when using your suggested change

src/silo/common/lineage_tree.cpp
```diff
@@ -0,0 +1,242 @@
...
+               parent_lineage,
+               lineage.lineage_name
+            ));
+         }
+         if (alias_mapping.contains(parent_id.value())) {
+            parent_id = alias_mapping.at(parent_id.value());
+         }
+         edge_list.emplace_back(parent_id.value(), my_id);
+      }
+   }
```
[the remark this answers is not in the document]
Taepper: I just did the same thing locally before :D

src/silo/preprocessing/sequence_info.test.cpp
```diff
@@ -19,7 +19,7 @@ TEST(SequenceInfo, validatesSuccessfulOnCorrectFile) {
 
 TEST(SequenceInfo, failWhenTooManyGenomesInReferences) {
    const auto reference_genomes =
-      ReferenceGenomes::readFromFile("testBaseData/exampleDataset/reference_genomes.json");
+      ReferenceGenomes::readFromFile("testBaseData/exampleDatasetAsTsv/reference_genomes.json");
```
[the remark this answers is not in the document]
Taepper: I want to keep the semantics the same, so all tests on the tsv dataset should remain on the tsv dataset :)

To keep this change "atomic"

include/silo/storage/lineage_index.h
```diff
@@ -0,0 +1,41 @@
...
+#include "silo/common/bidirectional_map.h"
+#include "silo/common/lineage_tree.h"
+#include "silo/common/types.h"
+
+namespace silo::storage {
+
+class LineageIndex {
+   friend class boost::serialization::access;
+
+   const common::LineageTree* lineage_tree;
```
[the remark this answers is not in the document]
Taepper: The references we had before (also in indexed_string_column.h) were nothing else than pointers, but much more error prone. I would like to keep this as a pointer and insert ASSERT statements that check for correct behavior

include/silo/storage/column/indexed_string_column.h
```diff
@@ -12,30 +12,46 @@
...
+         archive & lineage_index.value();
+      }
       // clang-format on
    }
 
    std::vector<Idx> value_ids;
    std::unordered_map<Idx, roaring::Roaring> indexed_values;
-   common::BidirectionalMap<std::string>& lookup;
+   std::optional<LineageIndex> lineage_index;
+   common::BidirectionalMap<std::string>* lookup;
```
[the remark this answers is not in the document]
Taepper: I think that shared_pointers can introduce a lot more complexity in debugging, and here we have a really simple object structure, which is guaranteed to be safe by design. Note that the (only) constructor is required to take the pointer as an argument. Also the only way to create the partition is from this same file, namely from the `IndexedStringColumn` (I will make the constructors private, which should already have been the case before). And the lifetime of the partition is bound to the lifetime of the `IndexedStringColumn` containing it, therefore the pointers will always be valid

src/silo/storage/reference_genomes.test.cpp
```diff
@@ -7,7 +7,8 @@
 
 TEST(ReferenceGenome, readFromFile) {
    auto under_test =
-      silo::ReferenceGenomes::readFromFile("testBaseData/exampleDataset/reference_genomes.json");
+      silo::ReferenceGenomes::readFromFile("testBaseData/exampleDatasetAsTsv/reference_genomes.json"
```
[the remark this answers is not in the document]
Taepper: I want to keep the semantics the same, so all tests on the tsv dataset should remain on the tsv dataset :)

To not make this PR change things it was not designed to

src/silo/database.cpp
```diff
@@ -567,16 +568,11 @@ Database Database::loadDatabaseState(const std::filesystem::path& save_directory
...
-            ::boost::archive::binary_iarchive input_archive(file_vec[partition_index]);
-            database.partitions[partition_index].serializeData(input_archive, 0);
-         }
-      }
-   );
+   for (size_t partition_index = 0; partition_index < database.partitions.size();
+        ++partition_index) {
+      ::boost::archive::binary_iarchive input_archive(file_vec[partition_index]);
+      database.partitions[partition_index].serializeData(input_archive, 0);
+   }
```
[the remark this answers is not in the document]
Taepper: oops

src/silo/preprocessing/lineage_definition_file.cpp
```diff
@@ -0,0 +1,129 @@
...
+         ));
+      }
+      return true;
+   }
+   static Node encode(const LineageName& lineage_name) {
+      Node node;
+      node = lineage_name.string;
+      return node;
+   }
+};
```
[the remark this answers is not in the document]
Taepper: Hmm, I just wanted to do this, but it is kind of weird, as the error message is very specific to the location where this is used. As this is not in the header file, I would say that this is just a implementation detail of the lineage_definition_file, therefore having it remain in here. I could put this in a function with a different name (and then also write a custom converter for vector<LineageName>, which calls that function), but putting it in the lineage_name.cpp file would mix concerns IMO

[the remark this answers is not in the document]
Taepper: Note that this would require us to also put this definition in the header file, "informing" all other files that include lineage_name.cpp what _YAML parsing is_ (including the YAML headers..)

src/silo/common/lineage_tree.cpp
fengelniederhammer: Ok, that's more than I expected. Then let's leave it as it is. (Still 1,5 s is quite few compared to the rest of the preprocessing.)

src/silo/preprocessing/sequence_info.test.cpp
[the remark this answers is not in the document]
fengelniederhammer: This is not a test on the tsv dataset. It just needs some reference genomes file and reads a NDJSON file. (And to be honest from the test as it is, I don't really know what it's testing. One would need to look at the files.)

include/silo/storage/lineage_index.h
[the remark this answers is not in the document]
fengelniederhammer: Ok, if the code actually gets simpler, maybe raw pointers are ok. Still they are a potential source of errors where using smart pointers already helps quite a bit.

src/silo/preprocessing/lineage_definition_file.cpp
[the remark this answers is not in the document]
fengelniederhammer: Ok, no big preference from my side, let's leave it as it is :+1:

include/silo/storage/column/indexed_string_column.h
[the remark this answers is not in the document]
Taepper: Re/ the mentioned complexity, for example see this recent blog post and discussion:
https://news.ycombinator.com/item?id=41717381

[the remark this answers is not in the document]
Taepper: But your point is totally valid, and I made an extensive comment at the top of `LineageIndex` file and making the constructors private is definitely the way to go, as then the API of the columns already guarantee that none of the raw pointer actually make it to the outside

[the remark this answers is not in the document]
Taepper: Having this visible when implementing the missing IndexedStringColumn tests with a LineageIndex (which you mentioned), I really liked that now the pointer passing completely disappeared

src/silo/config/database_config.cpp
```diff
@@ -144,7 +139,12 @@ struct convert<silo::config::DatabaseMetadata> {
       if (node["generateIndex"].IsDefined()) {
          metadata.generate_index = node["generateIndex"].as<bool>();
       } else {
-         metadata.generate_index = metadata.type == silo::config::ValueType::PANGOLINEAGE;
+         metadata.generate_index = false;
+      }
+      if (node["lineageIndex"].IsDefined()) {
+         metadata.lineage_index = node["lineageIndex"].as<bool>();
```
fengelniederhammer, line 145: As discussed: Let's rename it to `generateLineageIndex`
```suggestion
      if (node["generateLineageIndex"].IsDefined()) {
         metadata.generateLineageIndex = node["generateLineageIndex"].as<bool>();
```

fengelniederhammer:
I created https://github.com/GenSpectrum/LAPIS/issues/978 to update the documentation.

PR #591, opened 2024-09-20
Enable duckdb option partitioned_write_flush_threshold=1024 to reduce memory

taepper:
resolves spurious OOM crashes when preprocessing large datasets

### Summary

We now set the flag partitioned_write_flush_threshold to 1024, whereas its default value is 524288. Initial tests did not indicate a big slowdown of the preprocessing times.

We found this flag in a discussion of a similar issue [here](https://github.com/duckdb/duckdb/issues/9880#issuecomment-2266148994).

## PR Checklist


fengelniederhammer (changes requested):
Please adapt the commit message (with an eye on the changelog).

And maybe also put the link to that GitHub comment in the commit description? That would probably be valuable for someone trying to figure out in a year from now in the commit history, why we set that value.

fengelniederhammer (approved):

Issue #612, opened 2024-10-14
Implement lock for currently loaded directories


Taepper:
Currently there is no way for a user to know whether SILO (running as api) only requires the folder with the latest valid timestamp or also previous folder are required (due to being in transition from old to new / newer folders are corrupted due to a bug).

A locking mechanism in the form of a file could solve this problem

corneliusroemer:
There are good utilities for this that are part of UNiX so we don't need to reinvent, using flock for example, which makes it easy for multiple SILO instances running on the same data folder to each lock and prevent deletion.

See details from quick Chatgpt exploratory session: https://chatgpt.com/share/670e0ab2-5c78-8005-9373-246d6002ad47

theosanderson:
It's worth considering as part of implementing this whether having multiple SILO/LAPIS instances sharing a directory (with a load balancer, as Cornelius mentioned) is something you intend to support.

PR #613, opened 2024-10-14
fix: prevent setDatabase starvation which leads to no updates of currently loaded database

Taepper:
### Summary

resolves issue mentioned on slack, where the swap-over to the latest folder can sometimes fail.

Before, the setDatabase call requested a write-mutex. This was unnecessarily expensive. Instead, it is enough that we just atomically load and store the current database. Databases lifetimes for currently running transactions is now managed by a shared_ptr, which will correctly deallocate the old state once the last reference to the database is dropped (i.e. the last query using the old state has finished)

## PR Checklist

### DIFF
--- a/src/silo_api/database_directory_watcher.cpp
+++ b/src/silo_api/database_directory_watcher.cpp
@@ -116,7 +116,7 @@ void silo_api::DatabaseDirectoryWatcher::checkDirectoryForData(Poco::Timer& /*ti
    {
       try {
          const auto current_data_version_timestamp =
-            database_mutex.getDatabase().database.getDataVersionTimestamp();
+            database_mutex.getDatabase()->getDataVersionTimestamp();
          const auto most_recent_data_version_timestamp_found =
             most_recent_database_state->second.getTimestamp();
          if (current_data_version_timestamp >= most_recent_data_version_timestamp_found) {
@@ -137,6 +137,11 @@ void silo_api::DatabaseDirectoryWatcher::checkDirectoryForData(Poco::Timer& /*ti
    try {
       database_mutex.setDatabase(silo::Database::loadDatabaseState(most_recent_database_state->first
       ));
+      SPDLOG_INFO(
+         "New database with version {} successfully loaded.",
+         most_recent_database_state->first.string()
+      );
+      return;
    } catch (const std::exception& ex) {
       SPDLOG_ERROR(ex.what());
    } catch (const std::string& ex) {
@@ -150,7 +155,7 @@ void silo_api::DatabaseDirectoryWatcher::checkDirectoryForData(Poco::Timer& /*ti
       }
    }
    SPDLOG_INFO(
-      "New database with version {} successfully loaded.",
+      "Did not load new database with version {} successfully.",
       most_recent_database_state->first.string()
    );
 }
--- a/src/silo_api/database_mutex.cpp
+++ b/src/silo_api/database_mutex.cpp
@@ -1,31 +1,24 @@
 #include "silo_api/database_mutex.h"
 
-#include <mutex>
+#include <atomic>
 #include <utility>
 
 #include "silo/database.h"
 
 namespace silo_api {
 
-silo_api::FixedDatabase::FixedDatabase(
-   const silo::Database& database,
-   std::shared_lock<std::shared_mutex>&& mutex
-)
-    : lock(std::move(mutex)),
-      database(database) {}
-
 void silo_api::DatabaseMutex::setDatabase(silo::Database&& new_database) {
-   const std::unique_lock lock(mutex);
-   database = std::move(new_database);
+   auto new_database_pointer = std::make_shared<silo::Database>(std::move(new_database));
+
+   std::atomic_store(&database, new_database_pointer);
    is_initialized = true;
 }
 
-silo_api::FixedDatabase silo_api::DatabaseMutex::getDatabase() {
+std::shared_ptr<silo::Database> silo_api::DatabaseMutex::getDatabase() {
    if (!is_initialized) {
       throw silo_api::UninitializedDatabaseException();
    }
-   std::shared_lock<std::shared_mutex> lock(mutex);
-   return {database, std::move(lock)};
+   return std::atomic_load(&database);
 }
 
 }  // namespace silo_api
--- a/include/silo_api/database_mutex.h
+++ b/include/silo_api/database_mutex.h
@@ -1,19 +1,10 @@
 #pragma once
 
-#include <shared_mutex>
+#include <memory>
 
 #include "silo/database.h"
 
 namespace silo_api {
-
-class FixedDatabase {
-   std::shared_lock<std::shared_mutex> lock;
-
-  public:
-   FixedDatabase(const silo::Database& database, std::shared_lock<std::shared_mutex>&& mutex);
-
-   const silo::Database& database;
-};
 
 class UninitializedDatabaseException : public std::runtime_error {
   public:
@@ -22,15 +13,18 @@ class UninitializedDatabaseException : public std::runtime_error {
 };
 
 class DatabaseMutex {
-   std::shared_mutex mutex;
-   silo::Database database;
    bool is_initialized = false;
+   std::shared_ptr<silo::Database> database;
 
   public:
    DatabaseMutex() = default;
+   DatabaseMutex(const DatabaseMutex& other) = delete;
+   DatabaseMutex(DatabaseMutex&& other) = delete;
+   DatabaseMutex& operator=(const DatabaseMutex& other) = delete;
+   DatabaseMutex& operator=(DatabaseMutex&& other) = delete;
 
    void setDatabase(silo::Database&& new_database);
 
-   virtual FixedDatabase getDatabase();
+   virtual std::shared_ptr<silo::Database> getDatabase();
 };
 }  // namespace silo_api
--- a/src/silo_api/info_handler.cpp
+++ b/src/silo_api/info_handler.cpp
@@ -99,13 +99,13 @@ void InfoHandler::get(
 
    const auto fixed_database = database.getDatabase();
 
-   response.set("data-version", fixed_database.database.getDataVersionTimestamp().value);
+   response.set("data-version", fixed_database->getDataVersionTimestamp().value);
 
    const bool return_detailed_info = request_parameter.find("details") != request_parameter.end() &&
                                      request_parameter.at("details") == "true";
-   const nlohmann::json database_info =
-      return_detailed_info ? nlohmann::json(database.getDatabase().database.detailedDatabaseInfo())
-                           : nlohmann::json(database.getDatabase().database.getDatabaseInfo());
+   const nlohmann::json database_info = return_detailed_info
+                                           ? nlohmann::json(fixed_database->detailedDatabaseInfo())
+                                           : nlohmann::json(fixed_database->getDatabaseInfo());
    response.setContentType("application/json");
    std::ostream& out_stream = response.send();
    out_stream << database_info;
--- a/src/silo_api/query_handler.cpp
+++ b/src/silo_api/query_handler.cpp
@@ -35,9 +35,9 @@ void QueryHandler::post(
    try {
       const auto fixed_database = database_mutex.getDatabase();
 
-      auto query_result = fixed_database.database.executeQuery(query);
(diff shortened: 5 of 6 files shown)


src/silo_api/database_directory_watcher.cpp
```diff
@@ -137,6 +137,11 @@ void silo_api::DatabaseDirectoryWatcher::checkDirectoryForData(Poco::Timer& /*ti
    try {
       database_mutex.setDatabase(silo::Database::loadDatabaseState(most_recent_database_state->first
```
corneliusroemer, line 138: I would possibly add a log statement before starting key events as well. That makes it easier to see timings etc, these version changes are rare so logging liberally is fine

src/silo_api/database_directory_watcher.cpp
```diff
@@ -150,7 +155,7 @@ void silo_api::DatabaseDirectoryWatcher::checkDirectoryForData(Poco::Timer& /*ti
       }
    }
    SPDLOG_INFO(
-      "New database with version {} successfully loaded.",
+      "Did not load new database with version {} successfully.",
```
corneliusroemer, line 158: what happens if new version is not loaded successfully? does it retry later?

corneliusroemer (commented):
Thanks! I still have a few questions: 
1. Is my understanding correct that what happened on PPX is that: a) SILO didn't manage to switch to new version (for a reason I don't understand) and b) we somehow removed the data dir that was hence still in use (assuming the new one would be used but which wasn't actually).
1. How does the switch of versions happen? (Maybe there are some docs, in that case no need to reexplain). It's cool that you allow loading of new data without downtime, it's non trivial, there are quote a few pitfalls as a result.
1. What happens to new requests that are received while the db is being loaded? Are they rejected until the load is successful (you mention that the old version stays alive until the last request of the old version has completed)
1. What exactly happened in the PPX situation? The switch simply never happened? Why wasn't it retried? Will it be retried in the future, given this PR?
1. We should not remove old data version dirs until the change has happened successfully, this is why we hit a SILO error on PPX instead of the issue being just to show old data IIUC. How can we tell the old data version is no longer used, that we can hence safely delete the unused one?

src/silo_api/database_directory_watcher.cpp
fengelniederhammer: Yes, it will simply try again.

src/silo_api/database_directory_watcher.cpp
fengelniederhammer: What do you mean exactly? There is a log statement right before the `try` (in general this class consists of 50% log statements :D )

src/silo_api/database_directory_watcher.cpp
fengelniederhammer, line 158: We have log statements in every `catch`. Should we maybe simply add a clarifying "failed to load new database with version xyz" to the `SPDLOG_ERROR(ex.what());`?

fengelniederhammer (approved):
LGTM

Let's either create an issue that the changelog can reference or squash-merge this, such that the PR number ends up in the changelog.

fengelniederhammer:
>     2. How does the switch of versions happen? (Maybe there are some docs, in that case no need to reexplain). It's cool that you allow loading of new data without downtime, it's non trivial, there are quote a few pitfalls as a result.

SILO checks a certain directory every two seconds. It expects data versions in there (folders named after a timestamp). If it finds a folder with a higher number than the currently loaded, it attempts to load the contents.

>     3. What happens to new requests that are received while the db is being loaded? Are they rejected until the load is successful (you mention that the old version stays alive until the last request of the old version has completed)

They simply keep running. Loading the new DB happens in the background. Once the DB is loaded, the currently instance is swapped and new requests will see the new DB. 

>     4. What exactly happened in the PPX situation? The switch simply never happened? Why wasn't it retried? Will it be retried in the future, given this PR?

Is there something in the logs? It would actually be good to know.

>     5. We should not remove old data version dirs until the change has happened successfully, this is why we hit a SILO error on PPX instead of the issue being just to show old data IIUC. How can we tell the old data version is no longer used, that we can hence safely delete the unused one?

Currently, you cannot tell from the filesystem which data version is loaded. You could  ask the running SILO API.

src/silo_api/database_directory_watcher.cpp
corneliusroemer: Ah thanks it wasn't showing in default PR view without expanding context

src/silo_api/database_directory_watcher.cpp
corneliusroemer: Great, just out of curiosity, what's retry interval?

Taepper:
> Is my understanding correct that what happened on PPX is that: a) SILO didn't manage to switch to new version (for a reason I don't understand) and b) we somehow removed the data dir that was hence still in use (assuming the new one would be used but which wasn't actually).

Yes, that is correct.

> How does the switch of versions happen? (Maybe there are some docs, in that case no need to reexplain). It's cool that you allow loading of new data without downtime, it's non trivial, there are quote a few pitfalls as a result.

To switch the dataset, the "switcher" requests a unique lock. It is guaranteed that when it gets the lock, all other threads that also want this lock will block until we released our unique lock.

> What happens to new requests that are received while the db is being loaded? Are they rejected until the load is successful (you mention that the old version stays alive until the last request of the old version has completed)

Requests are still being served from the old state. The load and switch are two separate things.

> What exactly happened in the PPX situation? The switch simply never happened? Why wasn't it retried? Will it be retried in the future, given this PR?

The load of the old dataset was successful as can be seen in the log messages, but the switcher was _starved_ while waiting for its lock. It would retry if it failed, but the thread wanting to switch never really failed, it was just indefinitely waiting for a lock.

The requirement of a unique lock was unnecessarily expensive and complicated. Now, we no longer want a lock, but instead just atomically switch out the old dataset and the new dataset. We do not really care if a racy request still gets the old dataset or not (the header will show the updated dataset version after the switch was successful). Currently running queries rely on the fact that the data will stay in memory until they finished running. This is achieved by changing the return type to a `shared_ptr`, which will only release its memory once all references to the old dataset are gone

> We should not remove old data version dirs until the change has happened successfully, this is why we hit a SILO error on PPX instead of the issue being just to show old data IIUC. How can we tell the old data version is no longer used, that we can hence safely delete the unused one?

I added #612 for this.

src/silo_api/database_directory_watcher.cpp
corneliusroemer, line 138: Is there a timeout on this? When the incident happened, is this where it got stuck indefinitely? Just to help me understand what happened.

src/silo_api/database_directory_watcher.cpp
corneliusroemer: Ah 2 seconds

corneliusroemer:
> Loading the new DB happens in the background. Once the DB is loaded, the currently instance is swapped and new requests will see the new DB.

Does this mean that swapping can be blocked by if there happens to be continuously more than one active request? This would require a lot of requests and/or long ones but could happen at high load.

src/silo_api/database_mutex.cpp
```diff
@@ -1,31 +1,24 @@
...
 
-silo_api::FixedDatabase::FixedDatabase(
-   const silo::Database& database,
-   std::shared_lock<std::shared_mutex>&& mutex
-)
-    : lock(std::move(mutex)),
-      database(database) {}
-
 void silo_api::DatabaseMutex::setDatabase(silo::Database&& new_database) {
-   const std::unique_lock lock(mutex);
```
Taepper, line 18: This is the line where it got stuck before. A log message before would have helped a lot: "Acquiring unique lock to switch out datasets"

corneliusroemer:
>You could ask the running SILO API.

This works as long as there's only one instance running, if there are multiple pods being load balanced this would not be fool proof. But in our current PPX setup seems like a useful thing to try.

corneliusroemer:
Would it make sense to rename the classes to remove the `mutex` in it? IIUC, this PR drops the use of a muted.

Taepper:
> Does the new shared pointer implementation avoid the problem altogether in that new requests are served with the new version as soon as it has loaded successfully, even if there are still active old version requests? Just checking my understanding here.

Yes, precisely. Instead of waiting for all requests to finish, new requests now immediately get the new version. The lifetime management is done by the shared_ptr. (Without it, we would not know, when we would be able to release (i.e. remove from memory) the old database)

corneliusroemer (commented):
This might be too hard to do, but just as an idea, it'd be neat if it was possible to test that the version switch now works as expected:
1. Loading version 1
1. Faking/making running request A, that blocks until a later version becomes available 
1. Make version 2 available 
1. Make new request B and assert it uses version 2
1. Assert that request A completes successfully with data from version 1

fengelniederhammer:
> This might be too hard to do, but just as an idea, it'd be neat if it was possible to test that the version switch now works as expected:
> 
>     1. Loading version 1
> 
>     2. Faking/making running request A, that blocks until a later version becomes available
> 
>     3. Make version 2 available
> 
>     4. Make new request B and assert it uses version 2
> 
>     5. Assert that request A completes successfully with data from version 1

I just did that, it works. I added a 20 seconds sleep to the info endpoint and called it with curl, while preprocessing a new data version:

```
~curl 'localhost:8081/info' -vv
*   Trying 127.0.0.1:8081...
* Connected to localhost (127.0.0.1) port 8081 (#0)
> GET /info HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Tue, 15 Oct 2024 08:11:02 GMT
< Connection: Close
< X-Request-Id: 8693f837-3608-45d1-86a9-16d57f1af6b3
< data-version: 1728979701
< Content-Type: application/json
< 
* Closing connection 0
{"nBitmapsSize":3931,"numberOfPartitions":11,"sequenceCount":100,"totalSize":26589508}⏎  
``` 

```
[2024-10-15 10:10:30.935] [logger] [info] [database_directory_watcher.cpp:140] New database with version output/1728979701 successfully loaded.
[2024-10-15 10:11:02.467] [logger] [info] [logging_request_handler.cpp:15] Request Id [8693f837-3608-45d1-86a9-16d57f1af6b3] - Handling GET /info
[2024-10-15 10:11:14.127] [logger] [info] [database_directory_watcher.cpp:136] New data version detected: output/1728979861
[2024-10-15 10:11:14.127] [logger] [info] [database_config.cpp:211] Reading database config from output/1728979861/database_config.yaml
[2024-10-15 10:11:14.137] [logger] [info] [database.cpp:550] Finished loading partitions from output/1728979861/aa_sequences.silo
[2024-10-15 10:11:14.137] [logger] [debug] [database.cpp:668] build - initializing nucleotide sequences
[2024-10-15 10:11:14.631] [logger] [debug] [database.cpp:712] build - initializing amino acid sequences
[2024-10-15 10:11:14.882] [logger] [debug] [database.cpp:557] Loading partition data
[2024-10-15 10:11:16.962] [logger] [info] [database.cpp:580] Finished loading partition data
[2024-10-15 10:11:16.962] [logger] [debug] [database.cpp:728] Set data version to {timestamp: 1728979861, serializationVersion: 3}
[2024-10-15 10:11:16.962] [logger] [info] [database.cpp:583] Finished loading data_version from output/1728979861/data_version.silo
[2024-10-15 10:11:16.974] [logger] [info] [database.cpp:586] Database info after loading: sequence count: 100, total size: 26'589'508, N bitmaps size: 3'931
[2024-10-15 10:11:16.974] [logger] [info] [database_directory_watcher.cpp:140] New database with version output/1728979861 successfully loaded.
[2024-10-15 10:11:23.069] [logger] [info] [logging_request_handler.cpp:21] Request Id [8693f837-3608-45d1-86a9-16d57f1af6b3] - Responding with status code 200
```

You can see that the request arrived before the new data version was loaded. Then the new data version was loaded successfully, then the request finished and still shows the old data version.

corneliusroemer:
Nice @fengelniederhammer! I guess it's not straightforward to turn this into an automatic test 😄

fengelniederhammer:
> Nice @fengelniederhammer! I guess it's not straightforward to turn this into an automatic test 😄

I don't know how. You need to delay requests and make sure that the reload happens (and is finished!) during a running request. That seems pretty impossible to me.

PR #618, opened 2024-10-17
Pull request linting

### LINKED ISSUE #606: Pull request linting not working for large PRs
The github endpoint we are using to get the changed files in a PR only returns 30 files by default. These are also ordered lexicographically, making `src/` appear only after `/include` and `/endToEndTests`. Therefore the workflow can miss source files which are supposed to be checked.

Taepper:
resolves #606

### Summary

The github api for pull-requests is specified here:
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files

I now specify the `per_page` variable which is used for pagination to return 100 files (which is unfortunately the maximum) The default value is 30. This means that before no more than 30 files were returned by the endpoint in the CI action

## PR Checklist


fengelniederhammer (approved):
Seems to work as far as I could test it (on some other test branch)

PR #621, opened 2024-10-23
build: put conan-generated Find<lib>.cmake files in a separate directory from build/

Taepper:
### Summary

This follows the convention of putting helper .cmake files in to a separate directory as in: https://cmake.org/cmake/help/latest/variable/CMAKE_MODULE_PATH.html.

This additionally helps with developer setupt. There is now a separate folder (build/) to deleting all generated of the build system (cmake), and a folder (cmake/) of configuration files generated by the package manager (conan)

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,8 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 endif ()
 
 set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/build/")
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_BINARY_DIR}/generators/")
 
 # ---------------------------------------------------------------------------
 # Logging
--- a/Dockerfile_dependencies
+++ b/Dockerfile_dependencies
@@ -16,4 +16,4 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
         mv conanprofile.docker conanprofile; \
     fi
 
-RUN conan install . --build=missing --profile ./conanprofile --profile:build ./conanprofile --output-folder=build
+RUN conan install . --build=missing --profile ./conanprofile --profile:build ./conanprofile --output-folder=build/Release/generators
--- a/Dockerfile_dependencies
+++ b/Dockerfile_dependencies
@@ -6,7 +6,7 @@ RUN apt update && apt dist-upgrade -y \
     && apt install -y \
     cmake python3-pip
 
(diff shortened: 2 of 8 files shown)


Dockerfile_dependencies
```diff
@@ -16,4 +16,4 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
         mv conanprofile.docker conanprofile; \
     fi
 
-RUN conan install . --build=missing --profile ./conanprofile --profile:build ./conanprofile --output-folder=build
+RUN conan install . --build=missing --profile ./conanprofile --profile:build ./conanprofile --output-folder=cmake
```
fengelniederhammer: (Not necessarily for this PR): I wonder why we don't simply use the `build_with_conan` script here? Actually there is no need to replicate this command in several places. And installing Python in the build stage of the Dockerfile would be no big deal.

build_with_conan.py
```diff
@@ -18,34 +18,36 @@ def run_cmd(context: str, cmd: str):
...
+        conan_options.append("--output-folder=build/Release/generators")
     else:
+        build_folder = "build/Debug"
         cmake_options.append("-D CMAKE_BUILD_TYPE=Debug")
         conan_options.append("-s build_type=Debug")
+        conan_options.append("--output-folder=build/Debug/generators")
+
+    if args.clean:
+        print("----------------------------------")
+        print("cleaning build directory...")
```
fengelniederhammer, line 38:
```suggestion
        print(f"cleaning build directory {build_folder}")
```

CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 endif ()
 
 set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/build/")
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_BINARY_DIR}/generators/")
```
fengelniederhammer, line 22: Out of interest: Is there a reason to use `CMAKE_BINARY_DIR` instead of `CMAKE_SOURCE_DIR`?

Taepper: First, we save the `/build/` that was necessary before. (`${CMAKE_SOURCE_DIR}/build/` was always equivalent to `${CMAKE_BINARY_DIR}/build`)

But more importantly,  `CMAKE_BINARY_DIR` is now a lot less _magical_ in the discovery of the dependency files. For example before, CLion used `cmake-build-debug/` as the build folder. Still, `build/` was used to find the dependencies. Now, `CMAKE_BINARY_DIR` will already point at `cmake-build-debug/` and looking for dependencies is relative to this directory

PR #622, opened 2024-10-24
build: fix broken build for clang by removing erroneous generator for fmt library, which we do not explicitly depend on ourselves

Taepper:
### Summary

We depend on fmt transitively through spdlog. Unfortunately, this makes us unable to explicitly depend on a different version of fmt. (There is the newer version 11.0.2 available on conan center, but spdlog still depends on version 10.2.1). With that fmt version there is a bug appearing when using a new version of clang-20. This bug does not appear when not calling the generator for fmt as well.

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,11 @@ endif ()
 
 set(CMAKE_CXX_STANDARD 20)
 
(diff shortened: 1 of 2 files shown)


Taepper:
Now ready for review. Sorry for premature marking :)

fengelniederhammer (approved):

PR #624, opened 2024-10-24
Prevent VARCHAR round-trip auto cast, check for duplicate primary keys and add preprocessing tests

Taepper:
### Summary

We now prevent VARCHAR round-trip auto cast which lead to a VARCHAR column being first cast into a json and back by duckdb automatically. We also added a test for this.

Additionally, I added a check and test for duplicate primary keys

## PR Checklist


fengelniederhammer (approved):
LGTM

anna-parker:
Since this change the dummyOrganism is failing to start because 
```
[2024-10-25 10:29:27.109] [logger] [error] [api.cpp:289] Unknown metadata type: pango_lineage
```
This isn't very important as we only use it in the dummyOrganism - but would be good to figure out why

PR #639, opened 2024-11-06
panic macro prefix

pflanze:
### Summary

gtest has an `ASSERT_EQ` macro just like panic.h, and maybe there are more clashes. They weren't a problem at first, but with the command line parsing changes some tests now include panic.h and fail. Solve it once and for all by prefixing all the macro names.

## PR Checklist

- [x] The implemented feature is covered by an appropriate test. (Existing tests.)


Taepper:
Note that the linter error is a false positive caused by the downgrade of clang-tidy

fengelniederhammer:
The tests don't build yet:
`/src/src/silo/common/panic.test.cpp:113:4: error: ‘DEBUG_ASSERT_GE’ was not declared in this scope; did you mean ‘SILO_DEBUG_ASSERT_GE’?`

Is this intentionally a draft PR? Or is it actually ready for review?

pflanze:
> The tests don't build yet: `/src/src/silo/common/panic.test.cpp:113:4: error: ‘DEBUG_ASSERT_GE’ was not declared in this scope; did you mean ‘SILO_DEBUG_ASSERT_GE’?`
> 
> Is this intentionally a draft PR? Or is it actually ready for review?

As mentioned in the standup, that's an oversight (and then I was side tracked by the second unrelated issue first). I'll submit a fixed PR later, Alex will first submit a PR to fix the linter update issues.

pflanze:
Now should be fixed, except it's not based on a fixed main yet so the linter will still trigger.

BTW I forgot to mention the first of the two commits, which introduces TODO (now SILO_TODO), which is the panic macro I actually most in Rust. Meant to be used temporarily while working on a feature.

pflanze:
Ah and yes, draft was accidental, because Github reused the mode from the last PR.

Taepper (approved):

PR #640, opened 2024-11-07
Deprecate tsv fasta

### LINKED ISSUE #562: Remove tsv/fasta input format

Taepper:
resolves #562 

## PR Checklist


include/silo/config/preprocessing_config.h
```diff
@@ -51,13 +41,7 @@ class PreprocessingConfig {
    std::optional<uint32_t> duckdb_memory_limit_in_g;
    std::optional<std::filesystem::path> lineage_definitions_file;
    std::optional<std::filesystem::path> ndjson_input_filename;
```
fengelniederhammer, line 43: Can we make this non-optional now?

CMakeLists.txt
```diff
@@ -51,9 +51,8 @@ find_package(re2 REQUIRED)
 # Includes
 # ---------------------------------------------------------------------------
 
-include_directories(SYSTEM
+include_directories(
         ${CMAKE_SOURCE_DIR}/include
-        ${roaring_INCLUDE_DIRS}
```
fengelniederhammer, line 56: Why is this change?

src/silo/preprocessing/preprocessor.cpp
fengelniederhammer, line 1: According to my IDE, the includes of `"silo/database_info.h"`, `"silo/preprocessing/sequence_info.h"`and `"silo/preprocessing/sql_function.h"` are unused.

src/silo/preprocessing/preprocessor.test.cpp
fengelniederhammer, line 1: `TSV_FILE_WITH_SQL_KEYWORD_AS_FIELD`, `TSV_FILE_WITH_QUOTE_IN_FIELD_NAME` and `TSV_FILE_WITH_QUOTE_IN_PARTITION_BY` are unused

testBaseData/test_preprocessing_config_with_overridden_defaults.yaml
fengelniederhammer, line 1: Let's also set `preprocessing_database_location` and `duckdb_memory_limit_in_g`?

include/silo/preprocessing/preprocessor.h
fengelniederhammer, line 1: `createInsertionLambda` is unused

include/silo/sequence_file_reader/sequence_file_reader.h
```diff
@@ -1,29 +0,0 @@
...
-
-namespace silo::sequence_file_reader {
-class SequenceFileReader {
-  protected:
-   explicit SequenceFileReader(const std::filesystem::path& in_file_name)
-       : in_file(in_file_name){};
-   explicit SequenceFileReader(const std::string& file_content)
-       : in_file(file_content){};
-
-   silo::InputStreamWrapper in_file;
```
fengelniederhammer, line 16: `InputStreamWrapper` is now unused it seems.

fengelniederhammer (changes requested):
Actually looks good :+1: 

I would however rephrase the commit message: Instead of "deprecate" I would write something like "remove" (when reading deprecated as a user, I would think it still works, but I should slowly migrate away)

include/silo/config/preprocessing_config.h
Taepper: If this were non-optional, it would need a default value. Anyways, I would wait for the config refactor for this. Maybe it will resolve this already

CMakeLists.txt
Taepper: `${roaring_INCLUDE_DIRS}`

was unneeded because the `find_package` scripts and cmake did this already.

`SYSTEM` hid warnings from the specified include directories, as the flag is supposed to be used for system libraries which you cannot edit

fengelniederhammer (approved):

PR #642, opened 2024-11-07
PR #639 linter error diagnostic - OperatorResult refactor

Taepper:
### Summary

While diagnosing the linter error  #639 it appears it was not a false positive but really something we should consider. The error raised was about a shared pointer that was created for capturing a moveonly variable in a lambda. Using a shared_ptr for this purpose is highly discouraged as discussed here:
https://floating.io/2017/07/lambda-shared_ptr-memory-leak/
and
https://stackoverflow.com/questions/6458612/proper-way-to-receive-a-lambda-as-parameter-by-reference

Now, I instead made `OperatorResult` copyable and cleaned up some code, requiring less manual memory management. Furthermore, I renamed it to `CopyOnWriteBitmap` to be more descriptive.

## PR Checklist

### DIFF
--- a/include/silo/query_engine/actions/action.h
+++ b/include/silo/query_engine/actions/action.h
@@ -9,7 +9,7 @@
 #include <nlohmann/json_fwd.hpp>
 
 #include "silo/database.h"
-#include "silo/query_engine/operator_result.h"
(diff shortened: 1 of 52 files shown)


fengelniederhammer (approved):
LGTM

PR #652, opened 2024-11-27
doc: add documentation for the lineage_definitions format

Taepper:
### Summary

Add documentation for the lineage definitions file, which is required for the specification of monophyletic sublineage queries.

## PR Checklist


fengelniederhammer:
Linking https://github.com/GenSpectrum/LAPIS/issues/978 for reference.

Taepper:
I wanted something that is at least one step above "ah, I found an example in the codebase" which is what happened now for Loculus ^^

fengelniederhammer (approved):
I agree, this is definitely better than nothing. I don't want to block here, we can improve it later.

Taepper:
Ok, fixed a type and will merge

PR #666, opened 2024-12-18
Upgrade base image

### LINKED ISSUE #663: Update base-image version to ubuntu:24.04
We are currently on `ubuntu:22.04`. For example this limits our gcc features, as that image version uses `gcc 11.4.0` whereas the new long-term-support ubuntu image (`ubuntu:24.04`) provides `gcc 13.3.0`

When this is done, our uses of `std::find_if` should be changed to `std::ranges::find_if` to match our code style

Taepper:
resolves #663

### Summary

The base image was still at ubuntu:22.04. Upgrading this gives us access to a newer compiler version

## PR Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN  \
     && cp build/Release/silo .
 
 
-FROM ubuntu:22.04 AS server
+FROM ubuntu:24.04 AS server
 
 WORKDIR /app
 COPY docker_default_preprocessing_config.yaml ./default_preprocessing_config.yaml
--- a/Dockerfile_dependencies
+++ b/Dockerfile_dependencies
@@ -1,17 +1,22 @@
-FROM ubuntu:22.04
+FROM ubuntu:24.04
 
 ARG TARGETPLATFORM
 
 RUN apt update && apt dist-upgrade -y \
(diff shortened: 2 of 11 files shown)


Taepper:
Installing conan with `pipx` works seamlessly, however it is installed under `~/. local/bin`, which is not in `$PATH` by default. Running the recommended `pipx ensurepath` would add it to `$PATH` but a new shell would need to be started, which we would need to do from the `Dockerfile` itself? For now I specify the full path to the `conan` binary instead.

fengelniederhammer (changes requested):
LGTM :+1: 

Let's just rework the commit messages: 
* "fix: correctly specify and copy the conanprofile for arm architectures" doesn't need to be a `fix`. This doesn't affect any released artifact.
* The other "fix: install conan with pipx instead of pip" can maybe simply be squashed into the `feat` commit? We don't need this separately in the changelog.

fengelniederhammer:
I'm wondering why it shows "There is no change in the changelog. This PR will not produce a new releasable version."

Maybe we need to generate the missed release again by hand?

Taepper:
Ah, I also caught that, but was confused. Your point might be the right approach

PR #675, opened 2025-01-14
build: correctly specify our conan configuration for Debug builds

Taepper:
resolves current CI issue in #666 

### Summary

The comment in `build_with_conan.py` says all:
We still want to have our dependencies in Release, but the & tells conan we will build the consumer as Debug
Therefore, we set the flag `--settings '&:build_type=Debug'` (in both `build_with_conan.py` and `Dockerfile_dependencies`). 

It was not mentioned in the [docs](https://docs.conan.io/2/reference/commands/install.html), but I found this way of specifying packages only in: 
https://github.com/conan-io/conan/issues/14459

## PR Checklist

### DIFF
--- a/Dockerfile_dependencies
+++ b/Dockerfile_dependencies
@@ -22,4 +22,4 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
     fi
 
 RUN conan install . --build=missing --profile ./conanprofile --profile:build ./conanprofile --output-folder=build/Release/generators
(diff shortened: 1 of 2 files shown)


fengelniederhammer (approved):
:+1: 

Out of curiosity: Why do we actually need `--settings '&:build_type=Debug'` at all? Can't we simply use libraries that were built in Release mode when we build SILO in Debug mode?

Taepper:
That was my thinking as well. On Mac that also does work. But conan does some internal magic there that breaks if not using it as specified I suppose..

PR #676, opened 2025-01-14
Build arm images on demand

Taepper:
### Summary

When testing builds, it can be beneficial, to also have the arm images ready to test in PRs. This is not possible currently, because we only build these images on merges main, as github's arm emulation is quite time-consuming. This PR adds the possibility to request an ARM image by adding the label `arm-image` to the PR

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,11 +17,11 @@ env:
 jobs:
   dockerImage:
     name: Build docker images
-    runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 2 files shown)


fengelniederhammer (approved):
Looks good, let's try it :+1:

Taepper:
Uhhh, the runner is now repeatedly running out of memory

Taepper:
That was also the problem on the "alternate PR".. I think it is easier to fix here

PR #679, opened 2025-01-22
feat: add the version to the output of the info endpoint

### LINKED ISSUE #604: Add version information to /sample/info endpoint
I was hoping I could get the version of LAPIS that's serving a response from the `/sample/info` endpoint but it turns out I can't.

Might be nice to add this:
Google Chrome 2024-10-07 11 41 38

This came up as part of testing new versions of LAPIS and SILO on staging.pathoplexus.org and wondering whether I was already getting responses from the new versions (without having to log in to argocd, which of course works but takes more clicks)

Taepper:
resolves #604 

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,10 @@ jobs:
             echo "HEAD_SHA=${{ github.sha }}" >> $GITHUB_ENV
           fi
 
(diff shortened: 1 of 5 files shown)


fengelniederhammer:
Done by #680

PR #680, opened 2025-01-22
 feat: add the version to the output of the info endpoint

### LINKED ISSUE #604: Add version information to /sample/info endpoint

fengelniederhammer:
resolves #604

## PR Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,10 +1,9 @@
 name: Release SILO
 
 on:
-  workflow_run:
-    workflows: [LAPIS-SILO]
-    types: [completed]
-    branches: [main]
+  push:
+    branches:
+      - main
 
 env:
   DOCKER_IMAGE_NAME: ghcr.io/genspectrum/lapis-silo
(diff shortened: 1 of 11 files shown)


.github/workflows/release.yml
```diff
@@ -47,15 +47,26 @@ jobs:
...
-          for TAG in "${TAGS[@]}"; do
-            docker buildx imagetools create --tag $DOCKER_IMAGE_NAME:$TAG $DOCKER_IMAGE_NAME:latest
-          done
+        id: dockerMetadata
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.LAPIS_DOCKER_IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }}
+            type=raw,value=${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }}.${{ steps.release.outputs.patch }}
```
Taepper, line 58: The regression here is that this makes the `commit-sha` images and `0.5.0` images distinct, which might lead to additional downloads or more required debugging. (Note that the image that is built here will never have tests run against it)

Taepper (approved):
LGTM! Let's hope CI on main does not have surprises prepared for us :)

PR #683, opened 2025-01-28
build: simplify ci by always building the arm image

### LINKED ISSUE #682: Always build and run arm image in CI
With #681 I changed the runner for ARM images to an arm based one. Due to no qemu requirements, this sped up the building of that image by a lot. As it is now also being built parallel to the x86 image, we could remove quite some logic from the CI no longer conditioning when the image should be built

Taepper:
resolves #682 

### Summary

Straight-forward removal of the condition checking code

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,36 +50,7 @@ jobs:
           DIR_HASH=$(echo -n ${{ hashFiles('conanfile.py', 'conanprofile.docker', './Dockerfile_dependencies') }})
           echo "DIR_HASH=$DIR_HASH" >> $GITHUB_ENV
 
-      - name: Determine platform build necessity
-        run: |
-          # Default PLATFORM_BUILD to false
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -99,14 +70,12 @@ jobs:
           platforms: "linux/${{ matrix.arch }}"
```
fengelniederhammer, line 70: My guess is that we actually don't need this anymore, but it also doesn't hurt to leave it there.

fengelniederhammer (approved):
Nice :)

.github/workflows/ci.yml
Taepper: Hmm, I am actually unsure, which is the cleaner way

fengelniederhammer: Me too. Let's leave it as it is.

PR #687, opened 2025-01-28
chore: use clang-format 20

fengelniederhammer:
### Summary

Clang-format 17 is not in the standard apt repository for Ubuntu.

## PR Checklist


fengelniederhammer:
I'm acutally not sure whether clang 20 is officially released? I find their website confusing.

jidicula/clang-format-action doesn't support clang 20 yet. Ubuntu doesn't have clang-format 20 in its apt repo. I don't know whether this change is actually worth the trouble? Maybe we should just stay on 19?

PR #688, opened 2025-01-28
refactor: move preprocessing test cases from directories to unit tests

### LINKED ISSUE #623: Move preprocessor directory tests to unit tests
Currently, all preprocessor tests have a separate directory. We want to circumvent this by calling generators to generate the data as is now done with the invalid tests

Taepper:
resolves #623

### Summary

These directory tests are now all moved to unit tests. I tried to not change too of the data that is actually tested in these.

## PR Checklist


src/silo/preprocessing/preprocessor.test.cpp
```diff
@@ -114,14 +285,19 @@ const Scenario EMPTY_INPUT_NDJSON = {
...
 
-const Scenario EMPTY_INPUT_NDJSON_UNPARTITIONED = {
-   .input_directory = "testBaseData/emptyInputNdjsonUnpartitioned/",
-   .expected_sequence_count = 0,
-   .query = R"(
+const Scenario<Success> EMPTY_INPUT_NDJSON_UNPARTITIONED = {
+   .test_name = "EMPTY_INPUT_NDJSON_UNPARTITIONED",
+   .input_data = []() { return std::vector<NdjsonInputLine>{}; },
+   .database_config = NDJSON_WITH_NUMERIC_NAMES.database_config,
+   .reference_genomes = NDJSON_WITH_NUMERIC_NAMES.reference_genomes,
```
fengelniederhammer, line 297: IMO this is a little too sneaky. I wouldn't expect that someone reuses some random test data somewhere else. If someone adds `partition_by` to the `NDJSON_WITH_NUMERIC_NAMES.database_config`, then this test is also affected. I would rather split them:
* I'm not sure about the database config. Does it make sense to have a `DEFAULT_DATABASE_CONFIG_WITHOUT_PARTITIONING` here? Would other tests reuse it?
* It probably makes sense to have a standard `DEFAULT_REFERENCE_GENOMES` here that several tests can reuse (mostly tests than don't really care about which sequences are available).

src/silo/preprocessing/preprocessor.test.cpp
```diff
@@ -214,67 +552,134 @@ const Scenario PREVENT_LATE_AUTO_CASTING = {
...
+   std::ofstream file(config_with_input_dir.getNdjsonInputFilename().value());
+
+   if (!file.is_open()) {
+      std::cerr << "Could not open file for writing" << std::endl;
+      return;
+   }
+   for (const auto& line : scenario.input_data()) {
+      file << line.toJson().dump() << std::endl;
+   }
+   file.close();
```
fengelniederhammer, line 670: That is a lot of duplicated code, we could extract that into a function.

src/silo/preprocessing/preprocessor.test.cpp
```diff
@@ -214,67 +552,134 @@ const Scenario PREVENT_LATE_AUTO_CASTING = {
...
+   config_with_input_dir.input_directory = input_directory;
+   config_with_input_dir.intermediate_results_directory = input_directory;
+   config_with_input_dir.ndjson_input_filename = "input.json";
+   config_with_input_dir.validate();
+
+   std::ofstream file(config_with_input_dir.getNdjsonInputFilename().value());
+
+   if (!file.is_open()) {
+      std::cerr << "Could not open file for writing" << std::endl;
+      return;
```
fengelniederhammer, line 665: Does this return actually make the test silently fail? It prints something to stderr, but the test will probably still be reported as successful?

fengelniederhammer (commented):
Nice, I really like the new structure! The tests are quite easy to read..

There is still `testBaseData/ndjsonWithNullSequences` around. do we need that? Is it used somewhere?

I have a couple of questions, otherwise this looks good :+1:

src/silo/preprocessing/preprocessor.test.cpp
```diff
@@ -93,18 +250,32 @@ const Scenario NDJSON_WITH_NUMERIC_NAMES = {
...
+   .database_config =
+      {.schema =
+          {.instance_name = "Test",
+           .metadata =
+              {
+                 {.name = "accessionVersion", .type = ValueType::STRING},
+                 {.name = "2", .type = ValueType::STRING},
+              },
+           .primary_key = "accessionVersion",
+           .partition_by = "2"}},
```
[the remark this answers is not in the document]
Taepper: Yes, it is normally verified, but directly constructing the DatabaseConfig circumvents the validation

Taepper:
I adressed all your points.

For this point:
> I thought that we disallowed partitioning by columns with generate_lineage_index == false? What did we decide on in the end?

I first refactored a bit of the codebase to be totally happy with solution. I tried to leave all relevant parts in separate commits. The fixup commits will be squashed

fengelniederhammer (commented):
Some tests are failing, otherwise it looks good. The separate commits made it easy to review. :+1:

src/silo/preprocessing/preprocessor.test.cpp
```diff
@@ -672,7 +672,8 @@ INSTANTIATE_TEST_SUITE_P(PreprocessorTest, InvalidPreprocessorTestFixture, ::tes
 TEST_P(InvalidPreprocessorTestFixture, shouldNotProcessData) {
    const auto scenario = GetParam();
 
-   auto [input_directory, preprocessor] = prepareInputDirAndPreprocessorForScenario(scenario);
+   auto [input_directory, _preprocessor] = prepareInputDirAndPreprocessorForScenario(scenario);
+   auto preprocessor = std::move(_preprocessor);
```
fengelniederhammer, line 676: Out of curiosity: Why do we need this?

fengelniederhammer (approved):
Nice!

src/silo/preprocessing/preprocessor.test.cpp
Taepper: False positive of the linter... it detects the usage of the variable in the lambda, where it is used _as a reference_. This is dangerous, if the lambda lives longer than the variable it refers to. It is obvious that this cannot happen, but the linter lost its ability to detect this with the change of initialization.

Should I add this as a comment? I shortly thought about it, but didn't find a good way of succinctly explaining it, without leading to more confusion.

Maybe it would be the better solution do disable the warning at this position, because in that case, at least the fix is self-documenting

fengelniederhammer: IMO it also sounds good to disable the lint for this line and add a comment about the false positive.

Taepper: Lint disabling did not work.. I now added a struct, which I now like the most

PR #689, opened 2025-01-29
672 expose lineage definitions

### LINKED ISSUE #672: Expose the lineage definition file through an endpoint
LAPIS can already expose the reference genomes and the database config. However, LAPIS currently does not have knowledge of the lineage definition file. Therefore, we decided that SILO should have an endpoint that exposes this file.

SILO endpoint: `GET /linageDefinition/{column}`

It should check that the column exists and has generateLineageIndex: true 
→ SILO should support multiple lineage trees in the future

Taepper:
resolves #672

### Summary

LAPIS can already expose the reference genomes and the database config. However, LAPIS currently does not have knowledge of the lineage definition file. Therefore, we decided that SILO should have an endpoint that exposes this file.

SILO endpoint: GET /linageDefinition/{column}

It should check that the column exists and has generateLineageIndex: true
→ SILO should support multiple lineage trees in the future

This was rather straight-forward to implement. The `lineage_definition_file.h` can already deserialize YAML. Serialization to YAML is also added. (The alternative would have been to never discard the YAML string that we get when parsing, but this seems rather ugly to me).

## PR Checklist


src/silo/api/lineage_definition_handler.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+LineageDefinitionHandler::LineageDefinitionHandler(DatabaseMutex& database, std::string column_name)
+    : database(database),
+      column_name(std::move(column_name)) {}
+
+void LineageDefinitionHandler::get(
+   Poco::Net::HTTPServerRequest& request,
+   Poco::Net::HTTPServerResponse& response
+) {
+   const auto request_parameter = getQueryParameter(request);
```
fengelniederhammer, line 37: This looks unused?

src/silo/preprocessing/lineage_definition_file.test.cpp
```diff
@@ -24,6 +24,44 @@ TEST(LineageDefinitionFile, correctlyParsesFromYAML) {
...
+  aliases:
+    - SOMETHING_ELSE
+BOTH_PARENTS_AND_ALIASES:
+  parents:
+    - BASE
+    - SOMETHING
+  aliases:
+    - ALIAS1
+    - ALIAS2)");
+}
```
fengelniederhammer, line 59: The file for Covid has a bunch of entries like `A: {}`. Let's make sure that this is also preserved (from the e2e test I'm guessing that `{}` is converted to `~`, which mean `null`.

fengelniederhammer:
I also noticed while testing it that the call is very slow. It takes ~3 seconds to download the covid file

```
$ time curl 'http://localhost:8081//lineageDefinition/pango_lineage' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  251k    0  251k    0     0  71315      0 --:--:--  0:00:03 --:--:-- 71313

________________________________________________________
Executed in    3.62 secs      fish           external
   usr time    2.41 millis  209.00 micros    2.20 millis
   sys time    3.34 millis   38.00 micros    3.31 millis
``` 

Is that due to the size of the file? I guess we need something better so that it becomes properly usable for Loculus and our components.

src/silo/api/lineage_definition_handler.cpp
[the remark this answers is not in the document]
Taepper: these unit testing these api functions is a little cumbersome, so I hoped that the unit tests on the silo core code is sufficient

src/silo/preprocessing/lineage_definition_file.test.cpp
Taepper: Hmm, how strong are you on that requirement? This would require another approach

Taepper:
> I also noticed while testing it that the call is very slow. It takes ~3 seconds to download the covid file
> 
> ```
> $ time curl 'http://localhost:8081//lineageDefinition/pango_lineage' > /dev/null
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100  251k    0  251k    0     0  71315      0 --:--:--  0:00:03 --:--:-- 71313
> 
> ________________________________________________________
> Executed in    3.62 secs      fish           external
>    usr time    2.41 millis  209.00 micros    2.20 millis
>    sys time    3.34 millis   38.00 micros    3.31 millis
> ```
> 
> Is that due to the size of the file? I guess we need something better so that it becomes properly usable for Loculus and our components.

Did you test this as a debug or release build?

src/silo/preprocessing/lineage_definition_file.test.cpp
fengelniederhammer: What happens if we feed the downloaded file back into SILO? Will it yield the same results?

The strongest requirement is that you should be able to use the downloaded file to reconstruct the SILO setup.

fengelniederhammer:
> Did you test this as a debug or release build?

Yes, that was my local debug build.

With the built Docker images it's a lot faster:

```
$ time curl 'http://localhost:8080/lineageDefinition/pango_lineage' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  251k    0  251k    0     0   586k      0 --:--:-- --:--:-- --:--:--  587k

________________________________________________________
Executed in  433.50 millis    fish           external
   usr time    1.93 millis  296.00 micros    1.63 millis
   sys time    3.32 millis   59.00 micros    3.26 millis
```

Taepper:
I just pushed the alternative approach as discussed

src/silo/api/lineage_definition_handler.cpp
```diff
@@ -0,0 +1,72 @@
...
+   std::map<std::string, std::string> map;
+   const Poco::URI uri(request.getURI());
+   const auto query_parameters = uri.getQueryParameters();
+
+   for (const auto& parameter : query_parameters) {
+      map.insert(parameter);
+   }
+   return map;
+}
+}  // namespace
```
fengelniederhammer, line 25: Let's delete this function?

fengelniederhammer (approved):

PR #698, opened 2025-02-11
Refactor query_engine folder structure

Taepper:
### Summary

I moved the folder `silo/query_engine/operators` to `silo/query_engine/filter/operators` and the folder `silo/query_engine/filter_expressions` to `silo/query_engine/filter/expressions`

In streaming query engine PR I want to modularize the current "actions" to use common operators. This resturcturing should prevent possible future confusion.

## PR Checklist


src/silo/query_engine/operators/bitmap_producer.cpp
fengelniederhammer, line 1: The `.h` files of those are in `include/silo/query_engine/filter/operators/` -> Shouldn't the `.cpp` files then be in `src/silo/query_engine/filter/operators/`? (i.e. moving the `operators/` folder into `query_engine/`?)

include/silo/query_engine/filter/expressions/string_search.h
```diff
@@ -6,12 +6,12 @@
 #include <re2/re2.h>
 #include <nlohmann/json_fwd.hpp>
 
+#include "expression.h"
```
fengelniederhammer, line 9: Out of curiosity: Why does this work without any path before `expression.h`? Because they are in the same directory?

Taepper: Yes! We should not use this way of including though..

src/silo/query_engine/operators/bitmap_producer.cpp
Taepper: Yes, I missed that!

fengelniederhammer (approved):
Thanks for making this a separate PR :)

Issue #697, opened 2025-02-11
SILO memory usage seems to grow by factor of 3-10 when downloads happen, no/little deallocation thereafter


corneliusroemer:
Fresh SILO instance with mpox in Pathoplexus takes about 2.7GB RAM.

That increases over time and with usage, e.g. while playing around with Genspectrum using that Pathoplexus LAPIS instance for a few minutes, RAM goes to 3.1GB.

The memory requirement goes up even till 10GB at which point Kubernetes kills the container.

This looks like a memory leak (or too much caching going on).

See OOM killing after 3 days of running:

```
  silo:                                                                                                                                                                                   
│     Image:         ghcr.io/genspectrum/lapis-silo:0.5.2                                                                                                                                                                                                                                                                                                   │
│     State:          Running                                                                                                                                                               │
│       Started:      Mon, 10 Feb 2025 23:12:03 +0100                                                                                                                                       │
│     Last State:     Terminated                                                                                                                                                            │
│       Reason:       OOMKilled                                                                                                                                                             │
│       Exit Code:    137                                                                                                                                                                   │
│       Started:      Fri, 07 Feb 2025 20:37:04 +0100                                                                                                                                       │
│       Finished:     Mon, 10 Feb 2025 23:12:01 +0100

```

See also https://loculus.slack.com/archives/C05G172HL6L/p1738956717450729

corneliusroemer:
One can see the resident set size go up slowly over time:

```
root@loculus-lapis-silo-mpox-5c5dcf697b-lj582:/app# ps -o pid,rss,vsz 1
    PID   RSS    VSZ
      1 3081468 4959368
... clicking around a bit on https://genspectrum.org/mpox/single-variant?sampleCollectionDateRangeLower=Since+2021&clade=IIb&
root@loculus-lapis-silo-mpox-5c5dcf697b-lj582:/app# ps -o pid,rss,vsz 1
    PID   RSS    VSZ
      1 3399148 5221512
```

corneliusroemer:
Some further investigations here: https://github.com/GenSpectrum/LAPIS-SILO/pull/700#pullrequestreview-2618405186

It seems the leak is _not_ unbounded, and limited to around 3-4x increase in memory requirement compared to before requests come in.

corneliusroemer:
[...]

PR #700, opened 2025-02-14
build(deps): update poco dependency version to 1.13.3

Taepper:
### Summary

There is a [fix](https://github.com/pocoproject/poco/pull/3992) in POCO for a leak in [v1.12.5](https://github.com/pocoproject/poco/blob/poco-1.12.5-release/CHANGELOG). It lead to overallocation of threads during heavy load which are then never deallocated subsequently. This upgrade fixes this leak. It is uncertain, whether it fully resolves the behavior in #697. Therefore, I also introduced an additional change that allocates all threads Poco will ever use already at startup. This makes it easier to profile SILO in the future.

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,8 +53,8 @@ jobs:
       - name: Check if dependency image exists
         run: |
           EXISTS=$(docker manifest inspect \
-            ${{ env.DOCKER_DEPENDENCY_IMAGE_NAME }}:filehash-${{ env.DIR_HASH }}-${{ matrix.arch }} \ 
-            > /dev/null 2>&1 && echo "true" || echo "false")
+            ${{ env.DOCKER_DEPENDENCY_IMAGE_NAME }}:filehash-${{ env.DIR_HASH }}-${{ matrix.arch }} > /dev/null 2>&1 && 
+            echo "true" || echo "false")
           echo "CACHE_HIT=$EXISTS" >> $GITHUB_ENV
 
       - name: Build and push dependency image if input files changed
--- a/conanfile.py
+++ b/conanfile.py
@@ -8,7 +8,7 @@ class SiloRecipe(ConanFile):
     requires = [
         "boost/1.82.0",
         "duckdb/1.0.0",
-        "poco/1.12.4",
+        "poco/1.13.3",
         "hwloc/2.9.3",
         "onetbb/2021.10.0",
         "nlohmann_json/3.11.2",
--- a/src/silo/api/api.cpp
+++ b/src/silo/api/api.cpp
@@ -30,8 +30,16 @@ int Api::runApi(const silo::config::RuntimeConfig& runtime_config) {
    );
    poco_parameter->setMaxThreads(runtime_config.api_options.parallel_threads);
 
+   // For better profiling, we do not want requests to allocate new threads in the thread pool.
+   // Instead, just allocate all of them directly on start-up (by setting minCapacity)
+   Poco::ThreadPool thread_pool(
+      /* minCapacity = */ runtime_config.api_options.parallel_threads,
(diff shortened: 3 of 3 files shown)


Taepper:
Here are the changes in performance / memory usage:

```
alexander@dev-1:~$ bash benchmark.bash --parallel 100 # silo 0.5.3
Using 100 parallel jobs for requests. Change with --parallel
Monitoring PID: 3667770
2025-02-14 12:14:21, 59356 KB
2025-02-14 12:15:05, 115988 KB
alexander@dev-1:~$ bash benchmark.bash --parallel 100 # silo just upgrade poco
Using 100 parallel jobs for requests. Change with --parallel
Monitoring PID: 3769624
2025-02-14 12:15:33, 59564 KB
2025-02-14 12:16:16, 88304 KB
alexander@dev-1:~$ bash benchmark.bash --parallel 100 # silo also specify ThreadPool
Using 100 parallel jobs for requests. Change with --parallel
Monitoring PID: 3876724
2025-02-14 13:10:14, 59532 KB
2025-02-14 13:10:58, 88000 KB
```

The "benchmark" here uses a sample of 1000 queries. 100 times, the 1000 queries are sent to SILO in parallel:
```
for i in {1..100}; do
  seq 1000 | xargs -I {} -P$PARALLEL curl -X POST http://localhost:8081/query \
    -H "Content-Type: application/json" \
    -d @temp_json/query_{}.json > /dev/null 2>&1
done
```
where the files `temp_json/query_1.json` to `temp_json/query_1000.json` are the query samples.

corneliusroemer (commented):
Thanks for looking into the memory leak!

I've tested this branch's image with Loculus and unfortunately it doesn't look like it fixes the leak - or at least not entirely.

Here's how to reproduce:

1. Make a PR in Loculus pointing at SILO image: https://github.com/loculus-project/loculus/pull/3690
2. Make a preview by adding preview label
3. Run a lot of downloads against LAPIS/SILO, e.g. by something like: `while True; do curl "https://lapis-try-leak-fix.loculus.org/west-nile/sample/details?downloadAsFile=true&downloadFileBasename=west-nile_metadata_2025-02-14T1710&dataFormat=tsv&compression=zstd" -o /dev/null; done`
4. Then study memory consumption with something like: `kubectl top pod --containers -A | perl -pe 's/ +/\t/g' | sort -h -k 5 -t $'\t' | tsv-pretty -m80  | rg 'silo ' | rg 'west'`

In my tests, SILO memory goes up from 270MB to 870MB after running the above download comman for something like a minute:

iTerm2 2025-02-14 18 21 39

There could be a slight improvement, however - running against main Loculus, the memory there increases to 1025Mi vs 868Mi for new SILO. Yeah, I think I can reproduce around 10-20% max RAM improvements with this PR 👍 though there's still lots to be saved this is a step in the right direction! Although now I see that main (old SILO) actually freed some memory (red), so now that's better than this PR's SILO

image

Interestingly, the leak seems bounded, memory consumption saturates (albeit at a high level around 3-4x higher than before the request flood).

corneliusroemer:
Nice benchmark, could you share the `benchmarks.sh` file - maybe just add it somewhere to the code base - and also files `temp_json/query_1.json` to `temp_json/query_1000.json`

fengelniederhammer (approved):
Seems like things are not getting worse by upgrading POCO, so it's good to get it done anyway :+1:

corneliusroemer:
Thanks for looking into the excess memory usage!

By the way there's a more recent version of POCO as well: 2.14.0, and last week 2.14.1 https://github.com/pocoproject/poco/releases

Taepper:
But unfortunately they are not yet available as recipes on our dependency manager conan:
conan.io/center/recipes/poco

corneliusroemer:
You can contribute to Conan 😄 I made an issue anyways: https://github.com/conan-io/conan-center-index/issues/26639

PR #707, opened 2025-02-19
Hotfix sequence download

### LINKED ISSUE #697: SILO memory usage seems to grow by factor of 3-10 when downloads happen, no/little deallocation thereafter

Taepper:
resolves #697 

### Summary

While #705 already fixes a memory contention directly, there is another more pressing concern with the download of Fasta sequences. 

Due to the same `malloc` behavior, this leads to an increase of ~1.5 GB every time all mpox sequences are downloaded on Pathoplexus. This is bounded at _some_ point, when malloc decides to finally give back some of its free memory (>90% of the total memory consumption!). This fix tells malloc to trim its memory on every Fasta request.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/actions/fasta.cpp
+++ b/src/silo/query_engine/actions/fasta.cpp
@@ -1,5 +1,9 @@
 #include "silo/query_engine/actions/fasta.h"
 
+#if defined(__linux__)
+#include <malloc.h>
+#endif
+
 #include <fmt/format.h>
 #include <spdlog/spdlog.h>
 #include <boost/numeric/conversion/cast.hpp>
@@ -296,6 +300,13 @@ QueryResult Fasta::execute(const Database& database, std::vector<CopyOnWriteBitm
                primary_key_column,
                result_row_indices.size()
             );
+#if defined(__linux__)
+            SPDLOG_INFO(
+               "Fasta sequences generated for partition. Manually invoking malloc_trim() to give "
+               "back memory to OS."
+            );
+            malloc_trim(0);
+#endif
 
             SILO_ASSERT(results.size() == result_row_indices.size());
 


corneliusroemer (approved):
Can't speak to the code changes - but I've tested this image in Loculus/Pathoplexus previews and it appears to fix the memory leak #697!

fengelniederhammer (approved):
What would actually happen if you restrict how much memory SILO can get? Would malloc still allocate more and more memory and then crash? Or would it hit the limit and then deal with how much it can get?

Taepper:
> What would actually happen if you restrict how much memory SILO can get? Would malloc still allocate more and more memory and then crash? Or would it hit the limit and then deal with how much it can get?

My feeling is that this alone will not help

corneliusroemer:
> What would actually happen if you restrict how much memory SILO can get?

What are you thinking of here? How would you restrict how much memory it gets? Are there JVM-like args for max-memory for malloc?

In Loculus we provide memory limits and these should be visible through environment variables. Malloc doesn't care and go over the limit until the kernel kills the process with OOM.

fengelniederhammer:
> Malloc doesn't care and go over the limit until the kernel kills the process with OOM.

I guess that pretty much answers it. I was just curious how it would behave in a restricted environment.

PR #709, opened 2025-02-20
686 filter position out of range

### LINKED ISSUE #686: Improve error when mutation filter position is out of range
For Covid, the S gene has a length shorter than 500.

https://lapis.cov-spectrum.org/open/v2/sample/aggregated?aminoAcidMutations=S:5000

returns

```
{
    "error": {
        "type": "about:blank",
        "status": 500,
        "detail": "Error from SILO: vector::_M_range_check: __n (which is 4999) >= this->size() (which is 1274)"
    },
    "info": {
        "dataVersion": null,
        "requestId": "8a05e16c-c5f3-458f-b441-6e7b6e133c5e",
        "requestInfo": "sars_cov-2_nextstrain_open on lapis.cov-spectrum.org at 2025-01-28T12:01:19.766176398",
        "lapisVersion": "0.3.10"
    }
}
``` 

It's good that we get an error, but I think:
* the status code should be 400 instead of 500
* we can return a better message than a generic C++ vector out of bounds error.

Taepper:
resolves #686

## PR Checklist


include/silo/query_engine/bad_request.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <stdexcept>
+#include <string>
+
+#define CHECK_SILO_QUERY(condition, message) \
+   if (!(condition)) {                       \
+      throw silo::BadRequest(message);       \
+   }
```
fengelniederhammer, line 9: Actually I like the change. We just need to be conscious about the decision that this leaks the concept of having a HTTP interface into the query engine.

Initially we thought that SILO might also have other interfaces (e.g. CLI), so we kept HTTP specifics out of the query engine/database code.

I'm not saying that we shouldn't do that. AFAIK there are not plans to add additional interfaces anytime soon. I just wanted to make this consequence explicit.

src/silo/test/has_mutation.test.cpp
```diff
@@ -0,0 +1,112 @@
...
+const QueryTestScenario HAS_AMINO_ACID_MUTATION = {
+   .name = "HAS_AMINO_ACID_MUTATION",
+   .query = createHasAminoAcidMutationQuery(1),
+   .expected_query_result = nlohmann::json::parse(R"([{"count": 1}])")
+};
+
+const QueryTestScenario HAS_NUCLEOTIDE_MUTATION_OUT_OF_RANGE = {
+   .name = "HAS_NUCLEOTIDE_MUTATION_OUT_OF_RANGE",
+   .query = createHasNucleotideMutationQuery(2000),
+   .expected_error_message = "HasNucleotideMutation position is out of bounds 2000 > 5"
```
fengelniederhammer, line 92: I think we should also add the edges as tests: `position: 5` and `position: 6`. This has potential for an off-by-one error.

fengelniederhammer (approved):
LGTM :+1:

PR #718, opened 2025-02-26
711 silo mutations returns many floats

### LINKED ISSUE #711: SILO `mutations` returns many floats
@anna-parker mentioned that the payload of `mutations` contains a lot of float values `proportion` with seemingly high precisions, but are just fractions of two integers. SILO could return these integers itself. Note that `count` by itself is not sufficient, as the number of _valid_ symbols by position (I guess this is ≈coverage) is also required.

We could add the possibility to the action to specify that the returned fields can either additionaly contain the `coverage`. Or only `count` and `coverage` is returned

Taepper:
resolves #711 

### Summary

`coverage` is now also returned by the mutations action.

A straight-forward implementation of the feature. `mutations` now take an additional field `fields`, which can contain any number of the normally returned fields. If empty, all are returned.

## PR Checklist


fengelniederhammer (approved):
LGTM

Issue #654, opened 2024-11-27
Change the query engine's RowGroup abstraction to arrow


taepper:
For the new query engine #527 we require a new abstraction for RowGroups that represent a set of ~10'000 rows.

These can be either column or row-oriented. They will be used both between steps of the query engine and before serializing to json.

A row-oriented approach would require a custom binary storage format as other abstractions will result in too much per-tuple overhead

This might be merged directly into main instead of the feature branch of #527

Taepper:
We will use the arrow format as the row group abstraction in our query engine

PR #720, opened 2025-03-03
feat: add `append` as a third silo execution mode for incremental preprocessing

### LINKED ISSUE #368: ENH: Support incremental preprocessing
For the use case of processing large amounts of sequence data that are mostly changed only by addition of new sequences, it might be nice if preprocessing supported an incremental mode, that uses the previous processing result to speed up preprocessing.

I don't know the internals of what preprocessing does - but this could be very useful for use cases that want to do frequent updates with minimal delay.

Taepper:
resolves #368

### Summary

This PR adds `append` as a third execution mode to SILO

TODOs
- [x] be able to specify the partition to which SILO should append to (in separate issue #738)
- [x] add a check for unique primary keys! Currently, it is done during the preprocessing. The database itself should have mechanisms for it

## PR Checklist


src/silo/append/database_inserter.cpp
```diff
@@ -0,0 +1,3 @@
+#include "silo/append/database_inserter.h"
+
+namespace silo::append {}
```
fengelniederhammer, line 3: Should we put some implementation here? ;)

src/main.cpp
```diff
@@ -83,11 +87,16 @@ int mainWhichMayThrowExceptions(int argc, char** argv) {
...
+      mode = ExecutionMode::APPEND;
    } else if (mode_argument == "api") {
       mode = ExecutionMode::API;
+   } else if (mode_argument == "initialize") {
+      mode = ExecutionMode::INITIALIZE;
    } else {
       std::cerr << program_name
-                << ": need either 'preprocessing' or 'api' as the first program argument, got '"
+                << ": need 'preprocessing', 'initialize', 'append' or 'api' as the first program "
+                   "argument, got '"
```
fengelniederhammer, line 99:
```
$ build/Debug/silo
silo: need either 'preprocessing' or 'api' as the first program argument
``` 
Let's do the same in lin 82 (or extract that message into a commonly used variable/method)

fengelniederhammer: `config_specification.cpp`, line 66 has the same issue (because it only says "use 'silo api|preprocessing --help'")

src/silo/config/append_config.cpp
```diff
@@ -0,0 +1,113 @@
...
+            siloDirectoryOptionKey(),
+            ConfigValue::fromPath("."),
+            "The path to a silo-directory, a directory that contains silo outputs. This may be "
+            "used for input (see ..) and will be used for the output of the new silo state"
+         ),
+         ConfigAttributeSpecification::createWithoutDefault(
+            dataVersionOptionKey(),
+            ConfigValueType::STRING,
+            "The data version of the new database. If not given, a new data version will be minted "
+            "instead."
```
fengelniederhammer, line 56: I think it would be good to give some advice on when to use which. What's the consequence of specifying / not specifying it?

src/silo/config/append_config.cpp
```diff
@@ -0,0 +1,113 @@
...
+         ),
+         ConfigAttributeSpecification::createWithoutDefault(
+            appendFileOptionKey(),
+            ConfigValueType::PATH,
+            "The path to one or multiple files that contains the data that should be appended to "
+            "the database. If "
+            "no file is given, the data is expected on stdin instead."
+         ),
+         ConfigAttributeSpecification::createWithoutDefault(
+            partitionIdOptionKey(), ConfigValueType::PATH, "1, new, ~auto"
```
fengelniederhammer, line 66: Is `ConfigValueType::PATH` correct here?

fengelniederhammer, line 66: Should we also explain this option a little further?

src/silo/config/initialize_config.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+// Specification of the fields in inputs to the PreprocessingConfig struct
+ConfigSpecification InitializeConfig::getConfigSpecification() {
+   return ConfigSpecification{
+      .program_name = "silo initialize",
+      .attribute_specifications{
+         ConfigAttributeSpecification::createWithoutDefault(
+            initializeConfigOptionKey(),
+            ConfigValueType::PATH,
+            "The path to a initialize config that should be read before overwriting\n"
```
fengelniederhammer, line 50:
```suggestion
            "The path to an initialize config that should be read before overwriting\n"
```

fengelniederhammer (commented):
Review - part 1 ;)

src/silo/config/append_config.cpp
Taepper: I accidentally still included some options, whose behavior I removed afterall

include/silo/append/database_inserter.h
```diff
@@ -0,0 +1,88 @@
...
+
+template <std::ranges::range Data>
+   requires std::same_as<std::ranges::range_value_t<Data>, nlohmann::json>
+silo::append::TablePartitionInserter::Commit appendDataToTablePartition(
+   silo::append::TablePartitionInserter partition_inserter,
+   Data input_data
+) {
+   size_t line_count = 0;
+   for (const auto& json_obj : input_data) {
+      partition_inserter.insert(json_obj);
```
fengelniederhammer, line 55: `json_obj` is a `nlohmann::json`, isn't it? Why don't we directly ask it to convert itself to a proper struct that reflects the structure that we expect?
```suggestion
   for (const auto& json_obj : input_data) {
      partition_inserter.insert(json_obj.get<DataEntry>());
```

where 
```c++
struct DataEntry {
   map<...> metadata;
   map<...> aligned_nucleotide_sequences;
   // ...
}
``` 

Then we could also distinguish better between error (whether the line had a wrong format or whether the data in a line was wrong)

src/silo/storage/column_group.cpp
```diff
@@ -19,73 +19,281 @@
...
+      ColumnPartitionGroup& columns,
+      const schema::ColumnIdentifier& column,
+      const nlohmann::json& value
+   ) {
+      auto column_value = value["metadata"][column.name];
+      if (column_value.is_null()) {
+         columns.getColumns<ColumnType>().at(column.name).insertNull();
+      } else {
+         columns.getColumns<ColumnType>().at(column.name).insert(column_value.get<std::string>());
+      }
```
fengelniederhammer, line 202: What happens if the value is not present in the JSON? Does it throw the `nlohmann::json::type_error` that the upstream method catches?

src/silo/storage/column/sequence_column.cpp
```diff
@@ -66,27 +72,39 @@ InsertionEntry parseInsertion(const std::string& value) {
...
 ) {
-   auto [position, insertion] = parseInsertion(insertion_and_position);
-   insertion_index.addLazily(position, insertion, row_id);
+   try {
+      auto [position, insertion] = parseInsertion(insertion_and_position);
+      insertion_index.addLazily(position, insertion, sequence_count - 1);
+   } catch (const std::runtime_error& error) {
+      SPDLOG_ERROR(error.what());
+      // TODO do something
+   }
```
fengelniederhammer, line 102: Let's do something? :) Probably throw?

src/silo/storage/column/sequence_column.cpp
```diff
@@ -66,27 +72,39 @@ InsertionEntry parseInsertion(const std::string& value) {
          return {.position_idx = position, .insertion = insertion};
       }
    } catch (const boost::bad_lexical_cast& error) {
-      const std::string message = "Failed to parse insertion due to invalid format: " + value;
-      throw silo::preprocessing::PreprocessingException(message + ". Error: " + error.what());
+      // TODO change error type
```
fengelniederhammer, line 75: Let's do it?

src/silo/storage/column/sequence_column.cpp
```diff
@@ -66,27 +72,39 @@ InsertionEntry parseInsertion(const std::string& value) {
...
+         "Failed to parse insertion due to invalid format. Expected position that is parsable as "
+         "an integer, instead got: '{}'",
+         value
+      );
+      throw std::runtime_error(message + ". Error: " + error.what());
    }
-
-   const std::string message = "Failed to parse insertion due to invalid format: " + value;
-   throw silo::preprocessing::PreprocessingException(message);
+   // TODO change error type
```
fengelniederhammer, line 83: Let's do it?

include/silo/storage/column_group.h
```diff
@@ -53,77 +51,57 @@ class ColumnPartitionGroup {
...
+   std::map<std::string, column::ZstdCompressedStringColumnPartition>
+      zstd_compressed_string_columns;
+
+   void addJsonValueToColumn(
+      const schema::ColumnIdentifier& column_identifier,
+      const nlohmann::json& value
    );
 
-   void addNullToColumn(const std::string& column_name, config::ColumnType column_type);
+   void addNullToColumn(const schema::ColumnIdentifier& column_identifier);
```
fengelniederhammer, line 85: This method seems unused?

include/silo/storage/column_group.h
```diff
@@ -53,77 +51,57 @@ class ColumnPartitionGroup {
...
 
    void reserveSpaceInColumn(
```
fengelniederhammer, line 87: This method seems unused?

fengelniederhammer (commented):
Round 2

src/silo/query_engine/actions/details.cpp
```diff
@@ -20,18 +20,30 @@
...
-      CHECK_SILO_QUERY(metadata.has_value(), "Metadata field " + field + " not found.");
-      field_metadata.push_back({metadata->name, metadata->getColumnType()});
+   field_metadata.reserve(fields.size());
+   for (const auto& field : fields) {
+      auto column = schema.getColumn(field);
+      CHECK_SILO_QUERY(
+         column.has_value() &&
+            column.value().type != silo::schema::ColumnType::NUCLEOTIDE_SEQUENCE &&
+            column.value().type != silo::schema::ColumnType::AMINO_ACID_SEQUENCE &&
+            column.value().type != silo::schema::ColumnType::ZSTD_COMPRESSED_STRING,
```
fengelniederhammer, line 43: Maybe a function `isSequenceColumn(column_type)` wouldn't hurt.

src/silo/api/info_handler.cpp
```diff
@@ -96,15 +21,9 @@ void InfoHandler::get(
...
-   const auto request_parameter = getQueryParameter(request);
-
    response.set("data-version", database->getDataVersionTimestamp().value);
 
-   const bool return_detailed_info = request_parameter.find("details") != request_parameter.end() &&
-                                     request_parameter.at("details") == "true";
-   const nlohmann::json database_info = return_detailed_info
-                                           ? nlohmann::json(database->detailedDatabaseInfo())
-                                           : nlohmann::json(database->getDatabaseInfo());
+   const nlohmann::json database_info = nlohmann::json(database->getDatabaseInfo());
```
fengelniederhammer, line 26: Isn't this actually already a breaking change? (Noone really uses the details info, but still)

fengelniederhammer (commented):
Round 3 - I'm mostly done.

src/silo/config/preprocessing_config.test.cpp
fengelniederhammer, line 1: Why did you delete this test?

include/silo/storage/column/bool_column.h
```diff
@@ -8,26 +8,30 @@
...
    [[maybe_unused]] void serialize(Archive& archive, const uint32_t /* version */) {
       // clang-format off
-      archive & column_name;
       archive & values;
       // clang-format on
    }
 
-   std::string column_name;
    std::vector<silo::common::OptionalBool> values;
+   Metadata* metadata;
```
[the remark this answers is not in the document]
fengelniederhammer: (Same for the other columns)

src/silo/storage/column/date_column.cpp
```diff
@@ -4,16 +4,20 @@
...
    return is_sorted;
 }
 
-void DateColumnPartition::insert(const silo::common::Date& value) {
-   values.push_back(value);
+void DateColumnPartition::insert(const std::string& value) {
+   const auto date_value = silo::common::stringToDate(value);
+   if (!values.empty() && date_value < values.back()) {
+      is_sorted = false;
+   }
```
fengelniederhammer, line 19: I.e. this essentially puts sorting a date column on the user? Did you check how much performance we would lose on our Covid instances by not sorting the dates anymore?

Should we maybe at least provide a tool that does the sorting for the users?

src/silo/storage/column/date_column.test.cpp
```diff
@@ -3,12 +3,13 @@
...
-   under_test.insert(silo::common::stringToDate("2025-01-01"));
-   under_test.insert(silo::common::stringToDate("2021-03-21"));
-   under_test.insert(silo::common::stringToDate(""));
+   under_test.insert("2020-01-01");
+   under_test.insert("2023-1-05");
+   under_test.insert("2021-12-03");
+   under_test.insert("2025-01-01");
+   under_test.insert("2021-03-21");
+   under_test.insert("");
 }
```
fengelniederhammer, line 15: This test doesn't really test anything, does it?

src/silo/storage/column/float_column.cpp
fengelniederhammer, line 1: Most includes seem unused now.

src/silo/storage/column/string_column.cpp
```diff
@@ -10,20 +10,40 @@ using silo::common::STRING_SIZE;
...
 
-StringColumnPartition::StringColumnPartition(
+std::optional<String<STRING_SIZE>> StringColumnMetadata::embedString(const std::string& string
+) const {
+   return String<STRING_SIZE>::embedString(string, dictionary);
+}
+
+YAML::Node StringColumnMetadata::toYAML() const {
+   YAML::Node yaml_node;
+   // TODO rename from dictionary to make distinct from indexedString column
```
fengelniederhammer, line 20: What about this TODO? Let's do it?

src/silo/storage/column/zstd_compressed_string_column.cpp
fengelniederhammer, line 1: Let's also add a unit test?

src/silo/query_engine/actions/fasta.cpp
```diff
@@ -63,259 +43,110 @@ void Fasta::validateOrderByFields(const Database& database) const {
    }
 }
 
-namespace {
+const size_t PARTITION_CHUNK_SIZE = 100;
```
fengelniederhammer, line 46:
```suggestion
constexpr size_t PARTITION_CHUNK_SIZE = 100;
```

src/silo/query_engine/actions/fasta.cpp
```diff
@@ -63,259 +43,110 @@ void Fasta::validateOrderByFields(const Database& database) const {
...
-      const auto create_table_query =
-         fmt::format("CREATE TABLE {} AS ({})", result_table_name, table_query);
-      SPDLOG_TRACE(
-         "Create table query for unaligned in-memory sequence tables: {}", create_table_query
-      );
-
-      query(connection, create_table_query);
+   for (const auto& sequence_name : sequence_names) {
+      const auto& column =
+         table_partition.columns.zstd_compressed_string_columns.at("_" + sequence_name);
```
fengelniederhammer, line 60: Why is that `"_"` here?

src/silo/preprocessing/preprocessor.test.cpp
fengelniederhammer, line 1: Do we have a similar test that replaces this one somewhere?

fengelniederhammer (changes requested):
Looks quite good :+1: Let's clean it up a bit, then it should be good to go.

In general, there are many variables and arguments that could be declared `const`. I'm not quite sure why the linter doesn't catch it?

Also, conceptually, after calling append, the database should be saved under a new data version, shoudn't it? After all, it's a new set of data.

src/silo/config/preprocessing_config.cpp
```diff
@@ -119,6 +99,21 @@ ConfigSpecification PreprocessingConfig::getConfigSpecification() {
             ConfigValue::fromPath("reference_genomes.json"),
             "File name of the file holding the reference genome. Relative from inputDirectory."
          ),
+         ConfigAttributeSpecification::createWithDefault(
+            ndjsonInputFilenameOptionKey(),
+            ConfigValue::fromPath("reference_genomes.json"),
+            "File name of the file holding the reference genome. Relative from inputDirectory."
```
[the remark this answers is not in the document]
Taepper: Ouchie

src/silo/schema/database_schema.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+TableSchema TableSchema::fromYAML(const YAML::Node& yaml_node) {
+   std::map<ColumnIdentifier, std::shared_ptr<storage::column::ColumnMetadata>> column_metadata;
+   for (const auto entry : yaml_node["columns"]) {
+      std::string column_name = entry["name"].as<std::string>();
+      ColumnType column_type = columnTypeFromString(entry["type"].as<std::string>());
+      column_metadata.emplace(
+         ColumnIdentifier{column_name, column_type},
+         storage::column::visit(
+            column_type, ColumnMetadataLoaderByType{}, column_name, entry["metadata"]
```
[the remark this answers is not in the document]
Taepper: Yes, exactly! It was a little bit of engineering to get it working, but you can find other examples (where it was more fitting), where I pass a lambda instead

src/silo/append/append.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+silo::SiloDataSource getMostRecentOrSpecifiedDatabaseState(
+   const silo::SiloDirectory& silo_directory,
+   const std::optional<std::filesystem::path>& specified_directory
+) {
+   if (specified_directory.has_value()) {
+      auto specified_data_source =
+         silo::SiloDataSource::checkValidDataSource(specified_directory.value());
+      if (specified_data_source == std::nullopt) {
+         throw AppendError{"The specified siloDataSource directory is not valid data-source."};
```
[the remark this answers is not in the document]
Taepper: Some `Result` type might be a good thing, for now, I would opt for the exception for consistency in the code-base

include/silo/append/database_inserter.h
```diff
@@ -0,0 +1,88 @@
...
+silo::append::TableInserter::Commit appendDataToTable(
+   silo::storage::Table& table,
+   Data input_data
+) {
+   TableInserter table_inserter(&table);
+
+   // TODO(#738) make partition configurable
+   auto table_partition = table_inserter.openNewPartition();
+
+   appendDataToTablePartition(table_partition, input_data);
```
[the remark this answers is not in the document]
Taepper: I think this is safer, from a type-system point-of-view. I implemented the Commit mechanism to make sure, that one cannot forget to call `.commit` in all code paths. But after commit is called, one should not do anything with the object anymore

src/silo/storage/column_group.cpp
Taepper: Ah, actually no, it throws an `nlohmann::json::out_of_range`. I added a catch clause upstream

include/silo/append/database_inserter.h
Taepper: While true, this would mean that we copy all sequence data one additional time into that struct, and also adds more boiler-plate which makes this code more rigid. I am not a big fan of the current ndjson format, and maybe we want to change it in the future?

src/silo/database.cpp
```diff
@@ -9,368 +9,63 @@
...
-      partition.validate();
-   }
-}
-
-using RoaringStatistics = roaring::api::roaring_statistics_t;
+Database::Database(silo::schema::DatabaseSchema database_schema)
+    : schema(database_schema),
+      table(schema.tables.at(schema::TableName::getDefault())) {}
 
 DatabaseInfo Database::getDatabaseInfo() const {
```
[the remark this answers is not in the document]
Taepper: Good point, I added #740 to make rebasing the streaming PR easier

include/silo/database.h
```diff
@@ -11,108 +11,48 @@
...
    void saveDatabaseState(const std::filesystem::path& save_directory);
 
-   static Database loadDatabaseState(const std::filesystem::path& save_directory);
+   static Database loadDatabaseState(const silo::SiloDataSource& silo_data_source);
 
    [[nodiscard]] virtual DatabaseInfo getDatabaseInfo() const;
 
-   [[nodiscard]] virtual DetailedDatabaseInfo detailedDatabaseInfo() const;
-
    void setDataVersion(const DataVersion& data_version);
```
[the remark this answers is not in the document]
Taepper: added `updateDataVersion()` instead, which has to be public

src/silo/query_engine/actions/mutations.cpp
```diff
@@ -274,33 +277,37 @@ QueryResult Mutations<SymbolType>::execute(
...
    std::unordered_map<std::string, Mutations<SymbolType>::PrefilteredBitmaps> bitmaps_to_evaluate =
-      preFilterBitmaps(database, bitmap_filter);
+      preFilterBitmaps(table, bitmap_filter);
 
    std::vector<QueryResultEntry> mutation_proportions;
    for (const auto& sequence_name : sequence_names_to_evaluate) {
-      const SequenceStore<SymbolType>& sequence_store =
-         database.getSequenceStores<SymbolType>().at(sequence_name);
+      const storage::column::SequenceColumnMetadata<SymbolType>* sequence_column_metadata =
+         table.schema.getColumnMetadata<typename SymbolType::Column>(sequence_name).value();
```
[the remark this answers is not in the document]
Taepper: Might be viable, it is not a downgrade from what we had before (the references in the `..Partition` columns to the shared non `..Partition` one were effectively these raw pointers)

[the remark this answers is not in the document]
Taepper: So I would keep it as is for now..

src/silo/api/info_handler.cpp
Taepper: I think this can be argued either way. I would say it was an internal API (albeit not documented as such), which we do not need to mark breaking for users

src/silo/database.cpp
```diff
@@ -9,368 +9,63 @@
...
-         size_t local_nucleotide_symbol_n_bitmaps_size = 0;
-         for (const auto& [_, seq_store] : database_partition.nuc_sequences) {
-            local_total_size += seq_store.computeSize();
-            for (const auto& bitmap : seq_store.missing_symbol_bitmaps) {
-               local_nucleotide_symbol_n_bitmaps_size += bitmap.getSizeInBytes(false);
-            }
+   for (size_t partition_idx = 0; partition_idx < table.getNumberOfPartitions(); ++partition_idx) {
+      const storage::TablePartition& table_partition = table.getPartition(partition_idx);
+      for (const auto& [_, seq_column] : table_partition.columns.nuc_columns) {
+         total_size += seq_column.computeSize();
```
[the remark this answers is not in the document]
Taepper: It is only meant for rough guidance I guess. Yes, it only shows size for nucleotide sequence columns. I guess we can add an issue to change its behavior?

src/silo/config/preprocessing_config.test.cpp
Taepper: Ah, I removed preprocessing entirely at one point, only readded it then later. Will readd

src/silo/query_engine/actions/fasta.cpp
Taepper: I added the following extensive comment in `column_group.cpp`:
   // TODO(#741) we prepend the unalignedSequence columns (which are using the type ZstdCompressedStringColumnPartition) with 'unaligned_'. This should be cleaned up with a refactor and breaking change of the current input format.

src/silo/query_engine/actions/fasta.cpp
Taepper: This will be obsolete after #654

src/silo/storage/column/date_column.test.cpp
Taepper: I think it tests as much as was tested before

src/silo/storage/column/date_column.cpp
Taepper: Maybe @pflanze's benchmark script can test this in detail? staging uses this image for some time now, and I could not see a measurable difference

Taepper: I also just tested a bunch of curl requests against both staging and prod and could not see a regression for filter times

src/silo/storage/column/date_column.test.cpp
Taepper: I.e. no errors are thrown

include/silo/append/database_inserter.h
fengelniederhammer: Copying is a valid point, we probably want to avoid that. The data is potentially gigantic.

I think "rigidity" wouldn't hurt here. We have that rigidity already anyway since it's the same JSON structure that we're dealing with (no matter whether as a struct or as plain `nlohmann::json`). Having it in a struct makes it more transparent though.
Changing the NDJSON format later would require changes in any case, but having a struct the compiler could help.

src/silo/query_engine/actions/mutations.cpp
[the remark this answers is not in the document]
fengelniederhammer: Should we add an issue to make it a smart pointer later?

src/silo/api/info_handler.cpp
fengelniederhammer: But in the end it was a public endpoint ;) 

I don't mind so much. But making it breaking also doesn't hurt because it's just a different number that we're increasing.

src/silo/database.cpp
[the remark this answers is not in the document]
fengelniederhammer: Yes, that would be good. Noone cares currently, but it would be good to get it consistent (or at least documented) at some point.

src/silo/storage/column/date_column.test.cpp
fengelniederhammer: But the previous code under test threw errors, the current code doesn't.

I think it would be good to check that the values have actually been added (also the previous version of the test would have profited from that).

src/silo/storage/column/date_column.cpp
fengelniederhammer: Ok, let's try it. 

But doesn't that essentially make the `dateToSortBy` in the database config obsolete?

Taepper: Yes, that is true. I will add a deprecated warning

src/silo/api/info_handler.cpp
Taepper: But quite some work went into making this fully backwards-compatible :(

But the change to `dateToSortBy` probably warrants some user care anyways (even thought I highly expect actual impact to be minimal)

include/silo/storage/column/bool_column.h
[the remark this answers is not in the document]
Taepper: Also note here that I added #744 :)

src/silo/preprocessing/preprocessor.test.cpp
Taepper: I readded it!

src/silo/database.test.cpp
```diff
@@ -1,136 +1,76 @@
...
-   auto database = preprocessor.preprocess();
-
-   const auto simple_database_info = database.getDatabaseInfo();
-
-   EXPECT_GT(simple_database_info.total_size, 0);
-   EXPECT_EQ(simple_database_info.sequence_count, 100);
    EXPECT_EQ(simple_database_info.number_of_partitions, 1);
 }
 
 // NOLINTNEXTLINE(readability-function-cognitive-complexity)
```
fengelniederhammer, line 67: We probably don't need this one anymore, do we?

src/silo/database.test.cpp
```diff
@@ -1,136 +1,76 @@
...
 TEST(DatabaseTest, shouldReturnCorrectDatabaseInfo) {
```
fengelniederhammer, line 68: This test is pretty much a duplicate of the one above. The only difference is `EXPECT_GT(simple_database_info.total_size, 0)` vs `EXPECT_EQ(simple_info.total_size, 996)`.

src/silo/append/database_inserter.cpp
```diff
@@ -0,0 +1,53 @@
...
+         table_partition->columns.addJsonValueToColumn(column_metadata, ndjson_line);
+      } catch (const nlohmann::json::type_error& error) {
+         throw silo::append::AppendException(
+            "The following line does not conform to SILO's json specification when adding "
+            "to database column {}: '{}'\n"
+            "We got a json type_error: {}",
+            column_metadata.name,
+            ndjson_line.dump(),
+            error.what()
+         );
```
fengelniederhammer, line 18: I don't see a red unit test when I comment out this catch block -> Let's add a test (or maybe even several ones?)

src/silo/storage/column/sequence_column.cpp
```diff
@@ -53,40 +60,44 @@ struct InsertionEntry {
...
          const auto position = boost::lexical_cast<uint32_t>(position_and_insertion[0]);
          const auto& insertion = position_and_insertion[1];
          return {.position_idx = position, .insertion = insertion};
+      } catch (const boost::bad_lexical_cast& error) {
+         throw silo::storage::InsertionFormatException(
+            "Failed to parse insertion due to invalid format. Expected position that is parsable "
+            "as "
+            "an integer, instead got: '{}'",
+            value
+         );
```
fengelniederhammer, line 80: When I delete this catch block, then I don't get a red test -> let's add one?

src/silo/storage/column/sequence_column.cpp
```diff
@@ -53,40 +60,44 @@ struct InsertionEntry {
...
-      throw silo::preprocessing::PreprocessingException(message + ". Error: " + error.what());
    }
-
-   const std::string message = "Failed to parse insertion due to invalid format: " + value;
-   throw silo::preprocessing::PreprocessingException(message);
+   throw silo::storage::InsertionFormatException(
+      "Failed to parse insertion due to invalid format. Expected two parts (position and insertion "
+      "value), instead got: '{}'",
+      value
+   );
```
fengelniederhammer, line 87: I also don't get a red test if I delete this statement.

src/silo/storage/column/sequence_column.cpp
```diff
@@ -104,62 +115,67 @@ void silo::SequenceStorePartition<SymbolType>::finalize() {
...
+                 ++sequence_offset) {
+               const auto& [is_valid, sequence, offset] = lazy_buffer[sequence_offset];
                if (!is_valid || position_idx < offset || position_idx - offset >= sequence.size()) {
                   continue;
                }
                const char character = sequence[position_idx - offset];
                const auto symbol = SymbolType::charToSymbol(character);
                if (!symbol.has_value()) {
-                  throw silo::preprocessing::PreprocessingException(
+                  throw preprocessing::PreprocessingException(
```
fengelniederhammer, line 172: We could also add a unit test for this case.

fengelniederhammer (commented):
Looks good so far, I need to have another look tomorrow.

src/silo/database.cpp
```diff
@@ -9,368 +9,64 @@
...
+   }
 
    return DatabaseInfo{
       .version = silo::RELEASE_VERSION,
       .sequence_count = sequence_count,
       .total_size = total_size,
       .n_bitmaps_size = nucleotide_symbol_n_bitmaps_size,
-      .number_of_partitions = partitions.size()
+      .number_of_partitions = table.getNumberOfPartitions()
    };
```
fengelniederhammer, line 67: Not for this PR, I was just wondering: Do we want to include the data version in the info?

fengelniederhammer (commented):
Looks good :+1: 

I was thinking maybe we could add another unit test?
* initialize (via `Initializer::createSchemaFromConfigFiles`)
* check db info: check you have a data version and 0 sequences
* append a few sequences via `silo::append::appendDataToDatabase`
* check db info: check that the data version increased and you have x sequences
* append another few sequences
* check db info again.

src/silo/database.cpp
Taepper: True, this might be good. While it is anyways queryable from the http metadata, it might still make sense to have it all in one place

fengelniederhammer (approved):
Please add a reference to the issue to the changelog messages (e.g. by adding `resolves #...` to one of the commit messages).

corneliusroemer:
Thanks @Taepper - is there anything else we should know that's not contained in those help blurbs?

PR #726, opened 2025-03-12
fix: correctly expect an escaped stop codon in requests

### LINKED ISSUE #725: Insertions that start with stop codon don't work
The variant query `ins_S:486:*N` doesn't work ([link](https://lapis.cov-spectrum.org/open/v2/sample/aggregated?host=Homo+sapiens&variantQuery=ins_S%3A486%3A*N&fields=date&orderBy=date)) and throws:

```
{"error":{"type":"about:blank","title":"Internal Server Error","status":500,"detail":"Error from SILO: Mismatched '(' and ')' in regular expression"},"info":{"dataVersion":null,"requestId":"cebc726e-0b8e-4536-a37e-7eff731a2bd1","requestInfo":"sars_cov-2_nextstrain_open on lapis.cov-spectrum.org at 2025-03-10T17:20:21.674197789","reportTo":"Please report to https://github.com/GenSpectrum/LAPIS/issues in case you encounter any unexpected issues. Please include the request ID and the requestInfo in your report.","lapisVersion":"0.3.10"}}
```

but `ins_S:486:R*N` works ([link](https://lapis.cov-spectrum.org/open/v2/sample/aggregated?host=Homo+sapiens&variantQuery=ins_S%3A486%3A*N&fields=date&orderBy=date
))

(@anna-parker, since you're already looking into ANTLR and the variant queries for GenSpectrum/LAPIS#1041, would you mind also taking a look at this bug?)

Taepper:
resolves #725

### Summary

We now correctly expect a `\*` when searching for a stop codon. We validate the insertion search regex to only contain valid symbols, `.*` or `\*`.

Also, clean up of some code paths in the `insertion_index`

## PR Checklist

### DIFF
--- a/src/silo/query_engine/filter/expressions/insertion_contains.cpp
+++ b/src/silo/query_engine/filter/expressions/insertion_contains.cpp
@@ -67,8 +67,12 @@ std::unique_ptr<silo::query_engine::filter::operators::Operator> InsertionContai
       sequence_stores.at(valid_sequence_name);
    return std::make_unique<operators::BitmapProducer>(
       [&]() {
-         auto search_result = sequence_store.insertion_index.search(position_idx, value);
-         return CopyOnWriteBitmap(std::move(*search_result));
+         try {
+            auto search_result = sequence_store.insertion_index.search(position_idx, value);
+            return CopyOnWriteBitmap(std::move(*search_result));
+         } catch (const silo::storage::insertion::InsertionException& insertionException) {
+            throw silo::BadRequest(insertionException.what());
+         }
       },
       database_partition.sequence_count
    );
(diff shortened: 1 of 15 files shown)


src/silo/query_engine/filter/expressions/insertion_contains.cpp
```diff
@@ -67,8 +67,12 @@ std::unique_ptr<silo::query_engine::filter::operators::Operator> InsertionContai
...
    return std::make_unique<operators::BitmapProducer>(
       [&]() {
-         auto search_result = sequence_store.insertion_index.search(position_idx, value);
-         return CopyOnWriteBitmap(std::move(*search_result));
+         try {
+            auto search_result = sequence_store.insertion_index.search(position_idx, value);
+            return CopyOnWriteBitmap(std::move(*search_result));
+         } catch (const silo::storage::insertion::InsertionException& insertionException) {
+            throw silo::BadRequest(insertionException.what());
+         }
```
fengelniederhammer, line 75: Why do we need that catch here? Isn't the `silo::storage::insertion::InsertionException` only thrown in the preprocessing code?

fengelniederhammer (approved):
Looks good :+1: 

It would be great if you could mention the issue in the commit message with "resolves #..." so that we get the reference in the changelog.

src/silo/query_engine/filter/expressions/insertion_contains.cpp
Taepper: Actually, no. That was quite unclear in the code before.

But looking at the lines following https://github.com/GenSpectrum/LAPIS-SILO/blob/8c81c06300b33b5834aa22d1f499d9bf3681c05d/src/silo/storage/insertion_index.cpp#L29

This check is now removed, but we no longer throw runtime_errors and removed a lot of code duplications

PR #727, opened 2025-03-13
query-doc

Taepper:
### Summary

Some description on the filters and actions that are possible in SILO. I started this some time ago and finished it now, to help with the implementation of "generalized advanced queries"

## PR Checklist

### DIFF
--- a/documentation/maybe_documentation.md
+++ b/documentation/maybe_documentation.md
@@ -1,0 +1,16 @@
+
+### `Maybe`
+
+This filter is true iff its child expression is true, where,
+for all `NucleotideEquals` or `AminoAcidEquals` expressions
+contained in this filters subtree, _there exists at least
+one_ possible non-ambiguous sequence symbols at the sequence
+position, where the expression would evaluate to true.
+
+### `Exact`
+
+This filter is true iff its child expression is true, where,
+for all `NucleotideEquals` or `AminoAcidEquals` expressions
+contained in this filters subtree, the expression would
+evaluate to true _for all_ possible non-ambiguous sequence
+symbols at the sequence position.
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -1,0 +1,235 @@
+# Silo Query Documentation
+
+WARNING: This document is work-in-progress and offers some preliminary documentation
+
+All queries are accepted by the same `/query` endpoint.
+
+Each query consists of a `filter` and an `action`, described separately below.
+
+Example files can be seen in the [repository](https://github.com/GenSpectrum/LAPIS-SILO/tree/main/endToEndTests/test/queries).
+
+## Filter
+
+This is similar to the SQL `WHERE` clause. We directly get an abstract syntax tree of a boolean expression in JSON format.
+
+Any `Filter` is a json object that contains a `type` and a number of additional attribute, depending on the type as listed below. The types can be grouped as follows:
+
+### n-ary 
+
+always have attribute `children: Filter[]`
+
+####  `And`
+
+  This filter is true iff all its children are true.
+ 
+#### `Or`
+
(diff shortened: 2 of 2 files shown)


fengelniederhammer (commented):
Thanks for writing this up! 

I just wrote down my thoughts. I don't want to block on this too hard, since having this is already a huge improvement over having nothing ;) Let's discuss how far we want to improve this now before merging it.

Some general improvements would be good:
* We should define the response schema of each action.
* IMO every filter should at least have one correct example
* IMO every action should have at least one correct example for a request and for a response (they can be quite minimalistic).

We could also think about marking this as work in progress / draft to let readers know that it might be incomplete.

documentation/query_documentation.md
```diff
@@ -0,0 +1,200 @@
...
+Any `Filter` is a json object that contains a `Type` and a number of additional attribute, depending on the type as listed below. The types can be grouped as follows:
+
+- n-ary (always have attribute `children: Filter[]`)
+    - `And`
+
+      This filter is true iff. all its children are true.
+ 
+    - `Or`
+
+    This filter is true iff at least one of its children is true.
```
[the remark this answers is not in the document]
Taepper: Good point, it was a little painful as a single wrong line break messes up the current formatting completely

documentation/query_documentation.md
```diff
@@ -0,0 +1,200 @@
...
+
+      This filter is true iff its child is false.
+
+    - `Maybe`
+
+      This filter is true iff its child expression is true, where, for all `NucleotideEquals` or `AminoAcidEquals` expressions contained in this filters subtree, _there exists at least one_ possible non-ambiguous sequence symbols at the sequence position, where the expression would evaluate to true.
+
+    - `Exact`
+
+      This filter is true iff its child expression is true, where, for all `NucleotideEquals` or `AminoAcidEquals` expressions contained in this filters subtree, _for all_ possible non-ambiguous sequence symbols at the sequence position, where the expression would evaluate to true.
```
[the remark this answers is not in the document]
Taepper: The additional sentence you wrote is actually incorrect

Taepper:
I added issues #730 #731

documentation/query_documentation.md
```diff
@@ -0,0 +1,235 @@
...
+
+Any `Filter` is a json object that contains a `type` and a number of additional attribute, depending on the type as listed below. The types can be grouped as follows:
+
+### n-ary 
+
+always have attribute `children: Filter[]`
+
+####  `And`
+
+  This filter is true iff. all its children are true.
```
fengelniederhammer, line 23:
```suggestion
  This filter is true iff all its children are true.
```

documentation/query_documentation.md
```diff
@@ -0,0 +1,235 @@
...
+
+This filter is true iff its child is false.
+
+#### `Maybe`
+
+See `maybe_documentation.md`.
+
+#### `Exact`
+
+See `maybe_documentation.md`.
```
fengelniederhammer, line 46: Usually links work in IDEs and the browser:
```suggestion
#### `Maybe`

See [`maybe_documentation.md`](maybe_documentation.md).

#### `Exact`

See [`maybe_documentation.md`](maybe_documentation.md).
```

documentation/maybe_documentation.md
fengelniederhammer, line 1: I think an example would to understand the difference. (In a follow-up issue?)

fengelniederhammer (approved):
Looks good for now :+1:

documentation/maybe_documentation.md
Taepper: Yes, very valid. Was thinking the same, with some (ASCII) graphic, that explains the example step-by-step. Maybe in #730 ?

PR #735, opened 2025-03-29
make InsertionContains match on the full string

Taepper:
## PR Checklist


Taepper:
I will fix the commit message casing after review

fengelniederhammer (approved):
Let's maybe also add a short note on that in the query docs?

PR #742, opened 2025-04-15
fix: return 503 response when no database has been loaded yet

### LINKED ISSUE #723: SILO does not respond when no database has been loaded
* Start SILO. Don't load any data, make sure that the `--data-directory` is empty.
* Then `/info` and `/query` don't respond properly:

```
14:37:14 ~ $ curl localhost:8081/query
curl: (52) Empty reply from server
14:37:20 ~ $ curl localhost:8081/info
curl: (52) Empty reply from server
``` 

I'm quite sure that this is due to the `UninitializedDatabaseException` being thrown too early (when creating the request handlers instead of when evaluating the request handlers), thus the error handler doesn't get a chance to catch the exception.
https://github.com/GenSpectrum/LAPIS-SILO/blob/c6c8fd231a3d2304988707a9c095955c60ceb3a3/src/silo/api/active_database.cpp#L66-L72

fengelniederhammer:
resolves #723

### Summary

We need to delay getting the database into the handlers, otherwise the exception is not caught by the `ErrorRequestHandler`. The `SiloRequestHandlerFactory` must not throw exceptions, because then POCO simply returns an empty response.

## PR Checklist

### DIFF
--- a/include/silo/api/info_handler.h
+++ b/include/silo/api/info_handler.h
@@ -10,10 +10,10 @@ namespace silo::api {
 
 class InfoHandler : public RestResource {
   private:
-   std::shared_ptr<Database> database;
+   std::shared_ptr<ActiveDatabase> database_handle;
 
   public:
-   explicit InfoHandler(std::shared_ptr<Database> database);
+   explicit InfoHandler(std::shared_ptr<ActiveDatabase> database_handle);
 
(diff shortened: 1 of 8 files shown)


Taepper (approved):
LGTM, I was shortly debating whether putting the catch clause into `request_handler_factory.cpp` might lead to a cleaner solution, but am not sure afterall.

I think rebasing effort for #720 will be minimal, so this can already be merged before

Issue #777, opened 2025-05-27
SILO crashes on FastaAligned with InsertionContains filter


fengelniederhammer:
Currently the LAPIS e2e tests are failing, because SILO crashes. They use the dataset that is defined here: https://github.com/GenSpectrum/LAPIS/tree/73e22505fd1dfece8b1bffdb950c399c635f3ca7/lapis-e2e/testData/singleSegmented

One of the tests sends the SILO query
```
{
  "action": {
    "randomize": false,
    "type": "FastaAligned",
    "sequenceName": "main"
  },
  "filterExpression": {
    "children": [
      {
        "position": 25701,
        "value": "CC.*",
        "type": "InsertionContains"
      },
      {
        "position": 5959,
        "value": ".*AT",
        "type": "InsertionContains"
      }
    ],
    "type": "And"
  }
}
```
upon which SILO simply quits and doesn't respond.

Some datapoints: 
```
{
  "action": {
    "randomize": false,
    "type": "FastaAligned",
    "sequenceName": "main"
  },
  "filterExpression": {
    "position": 25701,
    "value": "CC.*",
    "type": "InsertionContains"
  }
}
```
also crashes SILO.

The same filter but with a `Details` action works. The same `FastaAligned` action with a simple `True` filter also works.

It happens both with the current `latest` image and with the `pr-775` image of #775.

Taepper:
Thanks!

PR #780, opened 2025-05-28
feat: update onetbb, nlohmann_json, gtest, roaring, yaml-cpp and zstd to most recent versions

taepper:
### Summary

We were quite a few versions behind on some dependencies. In the fix of #777 I want to use a newer roaring feature.

## PR Checklist


fengelniederhammer (approved):
I builds and the tests are green -> should be good.

Issue #786, opened 2025-06-02
SILO crashes on FastaAligned query


Taepper:
This query crashed the staging instance:

```
curl -X POST -H "Content-Type: application/json" -d '{
  "action": {
    "type": "FastaAligned",
    "sequenceName": "main", "orderByFields": ["date"], "additionalFields": ["date"]
  },
  "filterExpression": {
    "type": "DateBetween",
    "column": "date",
    "from": "2024-08-25",
    "to": null
  }
}' http://127.0.0.1:8086/query
```

The interval only contains 70564 sequences, so it should not pose any challenge to SILO whatsoever

Issue #785, opened 2025-06-02
`FastaAligned` doesn't return any data when used with `limit`


fengelniederhammer:
This test fails when added to the https://github.com/GenSpectrum/LAPIS-SILO/blob/3abd2d98acdcd0c50087a0b443fb8a6fd91cf6a8/src/silo/query_engine/actions/fasta_aligned.test.cpp

```
const QueryTestScenario FASTA_ALIGNED_WITH_LIMIT = {
   .name = "FASTA_ALIGNED_WITH_LIMIT",
   .query = nlohmann::json::parse(
      R"(
         {
            "action": {
               "type": "FastaAligned",
               "sequenceName": ["segment1"],
               "limit": 1
            },
            "filterExpression": {
               "type": "StringEquals",
               "column": "primaryKey",
               "value": "id_0"
           }
         }
      )"
   ),
   .expected_query_result = nlohmann::json::parse(
      R"([{"primaryKey":"id_0","segment1":"ATGCN"}])"
   )
};
```

It doesn't return any data, although it should.

(FWIW: The test works when you add `"orderByFields": ["primaryKey"]`.)

PR #792, opened 2025-06-04
fix: make better use of the arrow::acero API, no longer rely on passing the stream inside an extra operator

### LINKED ISSUE #782: Validate that arrow batches are emitted sequentially
Arrow might produce batches out-of-order. We possibly have to set appropriate options for batches to be guaranteed to be emitted in the correct order. We need to ensure this ordering if `orderByFields` are set

taepper:
resolves #782 #786 #785

### Summary

This PR refactors a bit of the new arrow code, which had a few bugs. A few more tests were added

## PR Checklist

### DIFF
--- a/src/silo/query_engine/actions/fasta_aligned.test.cpp
+++ b/src/silo/query_engine/actions/fasta_aligned.test.cpp
@@ -83,16 +83,19 @@ const QueryTestScenario FASTA_ALIGNED = {
    .query = nlohmann::json::parse(
       R"(
 {
-   "action": {
-     "type": "FastaAligned",
-     "sequenceName": ["segment1"],
-     "orderByFields": ["primaryKey"]
-   },
+  "action": {
+    "type": "FastaAligned",
+    "sequenceName": [
+      "segment1"
+    ],
+    "orderByFields": [
+      "primaryKey"
+    ]
+  },
   "filterExpression": {
     "type": "True"
   }
-}
-)"
+})"
    ),
    .expected_query_result = nlohmann::json::parse(
       R"(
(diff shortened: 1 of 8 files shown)


src/silo/query_engine/actions/fasta_aligned.test.cpp
```diff
@@ -245,6 +245,109 @@ const QueryTestScenario FASTA_ALIGNED_SUBSET = {
...
+{
+   "action": {
+     "type": "FastaAligned",
+     "sequenceName": ["segment1"],
+     "additionalFields": ["country"],
+     "orderByFields": ["country"]
+   },
+  "filterExpression": {
+    "type": "True"
+  }
```
fengelniederhammer, line 261: Nitpick: the indent is inconsistent.

src/silo/query_engine/actions/fasta_aligned.test.cpp
fengelniederhammer, line 1: I think it might still be good to add a test with limit but without order by.

src/silo/query_engine/exec_node/zstd_decompress_expression.cpp
```diff
@@ -29,15 +29,39 @@ struct BinaryDecompressKernel {
...
+      const arrow::ArraySpan& input_span = input.values[0].array;
+      std::string buffer_debug_info{fmt::format(
+         "buffers (data/size): [{}/{}, {}/{}, {}/{}]",
+         fmt::ptr(input_span.buffers[0].data),
+         input_span.buffers[0].size,
+         fmt::ptr(input_span.buffers[1].data),
+         input_span.buffers[1].size,
+         fmt::ptr(input_span.buffers[2].data),
+         input_span.buffers[2].size
+      )};
```
fengelniederhammer, line 49: Do we still need this? If yes, it might be worth putting in a guard so that we only execute it when trace logs are active. (Otherwise we construct a lot of useless strings that will never get logged anyway).

fengelniederhammer (commented):
Could you maybe explain what fixes the bug?

Also when I send this query to SILO, then it doesn't respond. The request simply times out:
```
{
    "action": {
      "sequenceName": ["main"],
      "additionalFields": ["division"],
      "limit": 5,
      "type": "FastaAligned"
    },
    "filterExpression": {
      "type": "True"
    }
}
```

Taepper:
#782 is fixed by not implementing a `SinkNode` inheriting from `ExecNode` ourselves, but just using the `SinkNode` implementation by arrow acero. This implementation includes a sequencer for batches.

#786 is also fixed by the above sequencer, as before, multiple batches of data could try sending data over the outgoing stream concurrently. But we did not have anything in place to allow concurrent access to the outgoing stream.

#785 is fixed by declaring implicit ordering on the `TableScan` and `LegacyResultProducer`


Unfortunately having the legacy and new implementation in place at the same time lowered our test-coverage immensely, as many code paths are now only tested for `FastaAligned` queries. Migrating many of the other actions should now be rather straight-forward and should be done before releasing the new version

PR #799, opened 2025-06-10
766 migrate details action to new framework

### LINKED ISSUE #766: Migrate `Details` action to new framework
See #758

Taepper:
resolves #766 

### Summary

This also migrates the optimized select-k implementation, when sorting is used in conjunction with limit

## PR Checklist


fengelniederhammer (changes requested):
Looks good, but I didn't find a unit test. Could you please add one? It's tested implicitly in many tests, but I think it would be good to have a dedicated test that checks various combinations of fields, limit, offset, etc, including error cases.

PR #803, opened 2025-06-12
refactor: merge /include directories into /src for easier code navigation

### LINKED ISSUE #771: Merge `/include` and `/src`
It can have advantages when they are separate. These advantages are mostly limited to the cases in which the project is used as a dependency. SILO is not designed as such, therefore the induced friction should be removed

Taepper:
resolves #771

### Summary

Both approaches (separate `/include` and `/src`, or together) have their pros and cons.

Together allows easier navigability, separate can be a bit easier for tooling, but this is mostly good for packages that are consumed by other packages, which is not the case for us.

## PR Checklist


fengelniederhammer (approved):
Nice :)

PR #806, opened 2025-06-16
feat(silo): read in phylogenetic tree files and create tree structure

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS-SILO/issues/807

### Summary

1. Adds new files `src/silo/preprocessing/phylo_tree_file` this contains:
- functions to read a phylogenetic tree from a simple [newick](https://en.wikipedia.org/wiki/Newick_format) or [auspice.json](https://docs.nextstrain.org/projects/auspice/en/stable/releases/v2.html#new-dataset-json-format) file (also handles edge cases if the file format is incorrect), adds tests for correct and incorrect case
- a `PhyloTreeFile` class, this class contains a map from treeNodeIds to TreeNode pointers `std::unordered_map<TreeNodeId, std::shared_ptr<TreeNode>> nodes`, 
- `TreeNode` class with 
```
class TreeNode {
  public:
   TreeNodeId node_id;
   std::vector<std::shared_ptr<TreeNode>> children;
   std::optional<std::shared_ptr<TreeNode>> parent;
   int depth;
};
```
2. Adds `src/silo/common/tree_node_id.h` with the simple `TreeNodeId` class to define a treeNodeId "type" (just a string of the node name)
3. Allow the phylogenetic tree file path to be passed in the `preprocessing_config.yaml` and allow the metadata column corresponding to the nodeId to be annotated with `generatePhyloTreeIndex` in the `database_config.yaml` - this column must be a type STRING but must not have a bitmap index generated (searches will be performed using the tree structure instead), also adds tests for this
4. Update the initializeDatabase code to generate a phyloTree from the file specified in the config and then create a schema using the phyloTree. For metadata columns of type StringColumnPartition allow them to be linked to a phyloTree if `generatePhyloTreeIndex` is set to true.

## PR Checklist


src/silo/preprocessing/phylo_tree_file.h
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+
+#include <nlohmann/json.hpp>
+
+namespace silo::preprocessing {
+
+class TreeNode {
+  public:
+   std::string node_id;
+   std::vector<std::string> children;
```
anna-parker, line 13: turn this into a vector of TreeNode pointers

src/silo/preprocessing/phylo_tree_file.h
```diff
@@ -0,0 +1,31 @@
...
+#include <filesystem>
+#include <vector>
+
+#include <nlohmann/json.hpp>
+
+namespace silo::preprocessing {
+
+class TreeNode {
+  public:
+   std::string node_id;
```
anna-parker, line 12: turn this into a type similar to LineageName

src/silo/preprocessing/phylo_tree_file.h
anna-parker: done

src/silo/preprocessing/phylo_tree_file.h
anna-parker: done

Taepper (approved):
Some error handling could still be stream-lined (the `invalid_argument` exception is never caught and could therefore also be the user facing `initialize_exception` already), but there were many other such cases in the code base. I think this will be cleaned up at a later point, where error reporting, specifically during preprocessing should be improved, as per user reports.

Therefore I would say this is good to go!

PR #809, opened 2025-06-17
769 migrate mutations action to new framework

### LINKED ISSUE #769: Migrate `Mutations` action to new framework
See #758

Taepper:
resolves #769

### Summary

This was a rather straight-forward implementation, moving the custom implementation to a `arrow::acero::SourceNode`.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/actions/mutations.cpp
+++ b/src/silo/query_engine/actions/mutations.cpp
@@ -8,6 +8,8 @@
 #include <variant>
 #include <vector>
 
+#include <arrow/acero/options.h>
+#include <arrow/compute/exec.h>
 #include <fmt/format.h>
 #include <fmt/ranges.h>
 #include <oneapi/tbb/blocked_range.h>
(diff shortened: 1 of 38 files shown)


src/silo/query_engine/actions/mutations.cpp
```diff
@@ -295,54 +307,115 @@ QueryResult Mutations<SymbolType>::execute(
...
+            ARROW_RETURN_NOT_OK(addMutationsToOutput(
+               sequence_name,
+               *sequence_column_metadata,
+               given_min_proportion,
+               bitmaps_to_evaluate.at(sequence_name),
+               output_builder
+            ));
+         }
+      }
+      // Order of result_columns is relevant
```
fengelniederhammer, line 356: Maybe we should also say why it's relevant?

fengelniederhammer (approved):
Looks good :) 

What do you think about adding more unit tests? I didn't find a test that sets a `sequenceName`. There are probably some more edge cases that are not covered yet.

Taepper:
@fengelniederhammer just pinging because I think you will like the second refactor commit :)

PR #815, opened 2025-06-19
feat(silo): add function to get all clades that are descendents of an internal node

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS-SILO/issues/814

1. Add new field `row_index` to PhyloTree class, this corresponds to the row of sample corresponding to a node (if it exists) and is inserted by `StringColumnPartition::insert` (serialization must be updated for this)
1. Add a new `PhyloDescendantOf` filter option, this uses new functions in the `phyloTree` class to get a bitmap of all samples that are descendants of an internal node in the tree.

Query features:
- throws an error if internal node does not exist in the tree
- returns all descendants (leaves) of an internal node that correspond to a sample (if the node does not exist in the samples then it will be ignored, additionally if a sample does not exist in the tree it will also of course be ignored)

## PR Checklist


Taepper (approved):
LGTM, just a few minor comments

PR #816, opened 2025-06-19
fix: do not write out column metadata as yaml, which might mess up utf-8

Taepper:
### Summary

Resolves current genspectrum bug, where silo errors on requests where it has to return a string, that contains multi-byte utf-8 characters at specific positions.

The bug was introduced in 0.6.0, where the `database_schema` had been refactored. Early in the PR, it was envisioned that the data structure might also used as the input format of the user (i.e. replacing the `database_config.yaml` / `reference_genomes.json` .. format). Later, also the metadata of columns was added to the data structure and with this, the deduplicated string suffixes.

`StringColumns` store short strings (<=16 bytes) in-place, while the suffixes of bigger strings (>16 bytes) are stored in a custom map instead of the heap for additional deduplication. The suffix is taken after 12 *bytes*, not characters. Therefore, if a multi-byte utf-8 character was at that boundary, the suffix might not be valid utf-8 anymore. This causes problems after the suffix then round-tripped to a yaml utf-8 string and back.

## PR Checklist

~- [ ] The implemented feature is covered by an appropriate test.~


anna-parker:
Oh Im curious about this! Could you maybe write a short summary of what caused the bug and how you fixed this? Will also make it easier to search for issues in the future :-)

Taepper:
> Oh Im curious about this! Could you maybe write a short summary of what caused the bug and how you fixed this? Will also make it easier to search for issues in the future :-)

Good call, I updated the PR with more details!

src/silo/common/lineage_tree.h
```diff
@@ -70,6 +70,16 @@ class LineageTreeAndIdMap {
       BidirectionalMap<std::string>&& lineage_id_lookup_map,
       std::string&& file
    );
+
+   friend class boost::serialization::access;
+   template <class Archive>
+   [[maybe_unused]] void serialize(Archive& archive, const uint32_t /* version */) {
```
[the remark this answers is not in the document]
anna-parker: ah I see it is archived in binary format using boost::archive::binary_oarchive - so I assume the format is all managed by boost. (As the functions were called FromYaml toYaml before I was wondering if this still had to be defined in order for boost to be able to write the object to disk)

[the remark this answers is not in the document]
Taepper: You got it. Applying the operator `archive >> object;` or `archive << object;` saves and loads the object to and from disk respectively. Then boost "looks" for the method serialize (and some other alternatives) at compile-time. If it does not find a serialization method, it will throw a compilation error

src/silo/database.cpp
```diff
@@ -105,20 +108,15 @@ void Database::saveDatabaseState(const std::filesystem::path& save_directory) {
...
    SPDLOG_INFO("Saving database schema");
 
-   auto yaml_string =
-      YAML::Dump(schema::DatabaseSchema{{{schema::TableName::getDefault(), table->schema}}}.toYAML()
-      );
-   const auto database_schema_filename = versioned_save_directory / "database_schema.yaml";
-   std::ofstream database_schema_file{database_schema_filename};
-   database_schema_file << yaml_string;
-   database_schema_file.close();
+   const auto database_schema_path = versioned_save_directory / DATABASE_SCHEMA_FILENAME;
```
anna-parker, line 111: so previously we were storing a yamlstring of the table->schema and now we are storing a binary_oarchive of tables?

src/silo/storage/column/column_metadata.h
```diff
@@ -14,9 +15,31 @@ class ColumnMetadata {
...
+
+BOOST_SERIALIZATION_SPLIT_FREE(silo::storage::column::ColumnMetadata);
+namespace boost::serialization {
+template <class Archive>
+[[maybe_unused]] void save(
+   Archive& ar,
+   const silo::storage::column::ColumnMetadata& object,
+   [[maybe_unused]] const uint32_t version
+) {
+   ar & object.column_name;
```
anna-parker, line 30: Does this mean that we only write the column_name to the archive? where do we write the data in that column?

src/silo/storage/column/string_column.h
```diff
@@ -96,3 +81,34 @@ class StringColumnPartition {
...
+[[maybe_unused]] void load(
+   Archive& ar,
+   std::shared_ptr<silo::storage::column::StringColumnMetadata>& object,
+   [[maybe_unused]] const uint32_t version
+) {
+   std::string column_name;
+   silo::common::BidirectionalMap<std::string> dictionary;
+   ar & column_name;
+   ar & dictionary;
+   object = std::make_shared<silo::storage::column::StringColumnMetadata>(
```
anna-parker, line 110: Don't we also have to `std::move(column_name)`? (You did that for the sequence column)

Taepper: It saves a copy, right!

src/silo/storage/column/zstd_compressed_string_column.h
```diff
@@ -69,3 +64,35 @@ class ZstdCompressedStringColumnPartition {
...
+[[maybe_unused]] void load(
+   Archive& ar,
+   std::shared_ptr<silo::storage::column::ZstdCompressedStringColumnMetadata>& object,
+   [[maybe_unused]] const uint32_t version
+) {
+   std::string column_name;
+   std::string dictionary_string;
+   ar & column_name;
+   ar & dictionary_string;
+   object = std::make_shared<silo::storage::column::ZstdCompressedStringColumnMetadata>(
```
anna-parker, line 94: Don't both the dictionary and column name need to be moved?

src/silo/storage/column/column_metadata.h
Taepper: We only write the column_name to the archive _that contains the database schema_. (For columns that do not have any additional metadata attached to the columns, otherwise the metadata is serialized by the respective subclasses of `ColumnMetadata` as well).

The data inside columns (which might be spread across multiple partitions) is not serialized in the same file as the schema, but in corresponding data files instead. You can take a look at `Table::loadData` and `Table::saveData` for the entry point of those serializations

anna-parker (approved):
Lgtm!

PR #826, opened 2025-06-25
Sort order fix

Taepper:
### Summary

The breaking change to sorting is partly being removed with this PR. Only multi-field sorting is still breaking.

## PR Checklist


fengelniederhammer (approved):
LGTM

PR #837, opened 2025-06-27
feat: add randomize to streaming query engine

### LINKED ISSUE #794: Add randomize to arrow query engine
The new arrow query engine does not support randomize. This feature should be introduced again. We might be breaking in the API as there are some considerations which make it unpleasant to implement.

This complication arises from the combination of interactions with:
randomize <-> offsetAndLimit
randomize <-> orderByFields

Ideally, we do not want to shuffle all input data if in the result we anyways only return a few records. Also, if orderByFields and randomize is given, it should first shuffle and then orderBy. Implementing both of these features together is a little bad to do in a general fashion, without introducing a check for the randomize flag at many different unrelated sections of the query engine -> complexity, harder to test, bugprone

Maybe we need to give more control to the user, where he must specify his correct semantics.

Some example choices that makes the implementation a lot easier (it is mostly a choose one of the following options):
- the user has freedom whether the limit is supposed to apply before or after the randomizing
- the user can decide whether the limit is supposed to apply before or after the orderBy
- randomize and orderBy together is not supported -> we return an error

Taepper:
resolves #794 

### Summary

This adds back randomization to the arrow query engine. Afterall, I found a solution that did still allow fully featured sorting

## PR Checklist

### DIFF
--- a/endToEndTests/test/queries/nOf_2of3_details.json
+++ b/endToEndTests/test/queries/nOf_2of3_details.json
@@ -3,7 +3,9 @@
   "query": {
     "action": {
       "type": "Details",
-      "orderByFields": ["date"]
+      "randomize": {
+        "seed": 321
+      }
     },
     "filterExpression": {
       "type": "N-Of",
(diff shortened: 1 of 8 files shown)


src/silo/query_engine/query_plan.cpp
```diff
@@ -19,6 +19,8 @@ arrow::Result<QueryPlan> QueryPlan::makeQueryPlan(
       exec_node::createGenerator(arrow_plan.get(), root, &query_plan.results_generator)
    );
    query_plan.results_schema = root->output_schema();
+   // ARROW_RETURN_NOT_OK(query_plan.results_schema->AddField(0,
+   // std::make_shared<arrow::Field>("__SILO_RANDOMIZE_HASH", arrow::uint64()))); // TODO remove
```
fengelniederhammer, line 23: Let's remove it? :)

fengelniederhammer (approved):
LGTM

src/silo/query_engine/query_plan.cpp
Taepper: Ahh, missed a stupid `TODO` something again! Will do!

PR #843, opened 2025-06-30
fix: evaluate the bitmaps in the preparation of the QueryPlan to not send wrong http headers

Taepper:
### Summary

This surfaced when Anya was working on a new filter. In the initial implementation, the `Expression::evaluate()` threw exceptions. While a refactor made it nicer to throw the exceptions already earlier. We do not want to send wrong http headers when the evaluate method failed. After all, it is a very cheap operation to evaluate the filer.

This is now also tested by an improvement to the class `InsertionContains`. Here we wrote code for the same error twice. Once, validating the input using a regex, then again when parsing the actual characters. It suffices to only throw when encountering bad characters. Now that evaluate method throws and is covered by two end-to-end tests.

Headers are sent after the query plan is constructed without errors. Afterwards the query plan is streamed to the output. Therefore, sending the headers even though an exception gets thrown right afterwards is bad, because we already sent the http code 200

## PR Checklist


src/silo/query_engine/filter/expressions/insertion_contains.cpp
```diff
@@ -72,53 +72,21 @@ std::unique_ptr<silo::query_engine::filter::operators::Operator> InsertionContai
...
-            throw silo::BadRequest(exception.what());
+            throw silo::BadRequest(
+               "The field 'value' in the InsertionContains expression does not contain a valid "
+               "regex "
+               "pattern: \"{}\". It must only consist of {} symbols and the regex symbol '.*'. "
+               "Also note "
+               "that the stop codon * must be escaped correctly with a \\ in amino acid queries.",
+               value,
+               SymbolType::SYMBOL_NAME_LOWER_CASE
+            );
```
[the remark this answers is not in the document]
Taepper: Yes,
`endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json`
and 
`endToEndTests/test/invalidQueries/insertionContains_invalidPattern2.json`

src/silo/query_engine/actions/action.h
```diff
@@ -61,6 +61,13 @@ class Action {
...
    ) const;
 
+  protected:
+   arrow::Result<arrow::acero::ExecNode*> addSortNode(
+      arrow::acero::ExecPlan* arrow_plan,
+      arrow::acero::ExecNode* node,
+      const silo::schema::TableSchema& table_schema
+   ) const;
+
   protected:
```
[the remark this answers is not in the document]
Taepper: Why?

src/silo/query_engine/actions/action.h
```diff
@@ -61,6 +61,13 @@ class Action {
       const silo::schema::TableSchema& table_schema
    ) const;
 
+  protected:
+   arrow::Result<arrow::acero::ExecNode*> addSortNode(
```
anna-parker, line 65: I can't see where this function is used?

anna-parker (dismissed):
Hi! In your PR title and description you say we are nopt setting the headers correctly when we error - I don't see how this is changed in this PR. Could you point to the code that changes this? And maybe a test that checks it?

src/silo/query_engine/exec_node/table_scan.cpp
```diff
@@ -184,8 +184,9 @@ arrow::Status TableScan::appendEntries(
 arrow::Status TableScan::produce() {
    SPDLOG_TRACE("TableScan::produce");
    for (size_t partition_idx = 0; partition_idx < table->getNumberOfPartitions(); ++partition_idx) {
-      auto filter_for_partition = partition_filter_operators->at(partition_idx)->evaluate();
-      silo::query_engine::BatchedBitmapReader reader{filter_for_partition, batch_size_cutoff - 1};
+      silo::query_engine::BatchedBitmapReader reader{
```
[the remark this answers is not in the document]
Taepper: That is part of the error handling

Taepper:
Yes. The http status code is not an error, but then the stream interrupts, but it is too late.

By calling evaluate() earlier, the error is correctly caught and headers are sent correctly.

Taepper:
These two tests get the correct error code now, before it was only due to manual validation:

> Yes,
> `endToEndTests/test/invalidQueries/insertionContains_invalidPattern.json`
> and
> `endToEndTests/test/invalidQueries/insertionContains_invalidPattern2.json`

src/silo/query_engine/actions/action.h
Taepper: Ah, that was indeed an error during rebase

anna-parker: Could you then remove the code if it is not used here?

src/silo/query_engine/query.cpp
```diff
@@ -38,23 +38,22 @@ QueryPlan Query::toQueryPlan(
...
          filter::expressions::Expression::AmbiguityMode::NONE
-      ));
+      );
       SPDLOG_DEBUG(
-         "Simplified query for partition {}: {}",
-         partition_index,
-         partition_filter_operators->back()->toString()
+         "Simplified query for partition {}: {}", partition_index, filter_operator->toString()
       );
+      partition_filters.emplace_back(filter_operator->evaluate());
```
anna-parker, line 53: this is the main change in the code, evaluate is now called during the `toQueryPlan` function and not afterwards. This means that errors will be seen before creating the header (header is created here: https://github.com/GenSpectrum/LAPIS-SILO/blob/main/src/silo/api/query_handler.cpp#L45)

anna-parker (approved):

PR #844, opened 2025-06-30
fix: add proper error messages for invalid groupByFields

### LINKED ISSUE #827: Invalid `Aggregated.groupByFields` throws non-descriptive error
```
{
    "action": {
      "groupByFields": [
        {
          "field": "test_boolean_column",
          "order": "ascending"
        }
      ],
      "type": "Aggregated"
    },
    "filterExpression": {
      "type": "True"
    }
}
```
returns 
```
{
  "error": "Bad request",
  "message": "The query was not a valid JSON: [json.exception.type_error.302] type must be string, but is object"
}
```

It would be nice if it told the user what's wrong.

Taepper:
also improve error of orderByFields

resolves #827 

## PR Checklist


src/silo/query_engine/actions/aggregated.cpp
```diff
@@ -25,16 +25,19 @@
...
+   const std::vector<GroupByField>& group_by_fields
 ) {
    std::vector<silo::schema::ColumnIdentifier> group_by_metadata;
-   for (const std::string& group_by_field : group_by_fields) {
-      auto column = schema.getColumn(group_by_field);
+   for (const GroupByField& group_by_field : group_by_fields) {
+      auto column = schema.getColumn(group_by_field.name);
       CHECK_SILO_QUERY(
-         column.has_value(), "Metadata field '{}' to group by not found", group_by_field
+         column.has_value(), "Metadata field '{}' to group by not found", group_by_field.name
```
fengelniederhammer, line 40: (Possibly in another PR): We could actually tell which fields would be available (currently it's quite impossible in SILO to find that out without looking into the config).

fengelniederhammer (approved):

PR #855, opened 2025-07-01
react to StopProducing calls

Taepper:
resolves #

## PR Checklist


Taepper:
I can confirm that downloading 10,100,1000 sequences with this PR completes in 1-5 seconds whereas before it took a few minutes

src/silo/query_engine/exec_node/table_scan.h
```diff
@@ -81,6 +83,7 @@ class TableScan : public arrow::acero::ExecNode {
 
    arrow::Status StopProducingImpl() override {
```
fengelniederhammer, line 84: Why is there actually `StopProducingImpl` and `StopProducing` (in the parent class)?

fengelniederhammer (approved):
LGTM (I didn't try it at large scale though)

PR #857, opened 2025-07-02
828 unprecise date between error

### LINKED ISSUE #828: Unprecise error when asking for `DateBetween` that is not a date column
```
{
  "action": {
    "type": "Aggregated"
  },
  "filterExpression": {
    "column": "country",
    "from": "2020-05-12",
    "to": "2025-06-12",
    "type": "DateBetween"
  }
}
```

returns 

```
{
  "error": "Bad request",
  "message": "The database does not contain the column 'country'"
}
```

which is not really correct. The column "country" exists. It should say that it's not a date column (or something in that spirit).

Taepper:
resolves #828

### Summary

Very small change. Also contains a slight refactor

## PR Checklist


fengelniederhammer (approved):
Probably other filters could also use the same fix (at least the StringEquals). But I don't want to block this PR.

PR #872, opened 2025-07-08
feat: enable filtering for recombinant lineages

### LINKED ISSUE #589: Recombinant wildcard queries
@corneliusroemer mentioned in #458 that 
> one interesting possible extension is to include recombinants not as their own trees but give them two parents. One could then offer an alternative wildcard query mode that treats a recombinant as a parent of both of their parents (but only counting them once if both parents are included in a query).
> The recombinant multi-parent mode would be cool, but is by far not as important as supporting general schemes.

While working on a PR for #458, I did not implement this feature as it comes with some complexity. Therefore this issue tracks the additional feature and can be prioritized accordingly

Taepper:
resolves #589

### Summary

This includes a specifiable mode for including all recombinant sublineages or only the ones that are fully contained in the clade of the filtered node

## PR Checklist


src/silo/storage/column/lineage_index.test.cpp
```diff
@@ -0,0 +1,141 @@
...
+      );
+      auto expected = lineages_with_sublineages_all_recombinants.at(lineage_id);
+      assertEqualHelper(actual, expected);
+   }
+}
+
+TEST(LineageIndex, allTreeValuesCorrectBehavior) {
+   auto lineage_tree = LineageTree::fromEdgeList(
+      6, {{1, 2}, {2, 3}, {1, 0}, {0, 3}, {3, 4}, {0, 5}, {5, 4}}, {}, {}
+   );
```
fengelniederhammer, line 86: Can we extract the edge list (or the lineage tree) so that the tests that test the same topology also use the same variable?

src/silo/common/lineage_tree.cpp
```diff
@@ -140,6 +169,112 @@ Idx LineageTree::resolveAlias(Idx value_id) const {
...
+   std::vector<std::vector<Idx>> child_relation(parent_relation.size());
+   for (Idx child = 0; child < parent_relation.size(); ++child) {
+      for (Idx parent : parent_relation.at(child)) {
+         child_relation.at(parent).emplace_back(child);
+      }
+   }
+
+   // We compute the topological rank to get a measure for the distance to the root
+   // The topological rank of a child is guaranteed to be greater than the topological rank of all
+   // its parents
```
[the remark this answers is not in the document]
Taepper: I had the thought: topological rank does describe itself, so the method does not need a comment. Here I comment to tell the reader _why_ we need it _at this point_

Taepper:
I added the test case with multiple roots and rebased.

src/silo/common/lineage_tree.test.cpp
```diff
@@ -317,4 +343,281 @@ TEST(containsCycle, correctDirectedAcyclicGraphs) {
...
+      correct_lca
+   );
+}
+
+TEST(LineageTree, correctlyHasNoAncestors) {
+   auto lineage_tree = createDiamondLineageTreeWithTwoRoots();
+   std::set<Idx> ancestors_of_3{3};
+   ASSERT_EQ(
+      lineage_tree.getAllParents(
+         3, silo::common::RecombinantEdgeFollowingMode::FOLLOW_IF_FULLY_CONTAINED_IN_CLADE
```
fengelniederhammer, line 619: Wouldn't it be more interesting to ask for the parents of `4`?

fengelniederhammer (approved):
LGTM

PR #875, opened 2025-07-10
fix(benchmarking): only depend on WisePulse for running, not the data

pflanze:
The resulting data doesn't depend on the data in
WisePulse/target/release/split_into_sorted_chunks--it should not be recreated just because the modification time on that executable changed!

This led to unnecessary CPU investment / latency for jobs.

## PR Checklist

### DIFF
--- a/benchmarking/Makefile
+++ b/benchmarking/Makefile
@@ -86,7 +86,8 @@ dependencies: $(API_QUERY) $(WISEPULSE)
 
 # ---- Sorting input -----------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


pflanze:
> I'm currently running commit [1bfdb35](https://github.com/GenSpectrum/LAPIS-SILO/commit/1bfdb35740f91d116bf51891b44c878043ec2363) on gs-staging-1 to verify that it works.

I can confirm that it works.

pflanze:
Force-pushed to include the PR number

PR #896, opened 2025-07-21
feat(benchmarking): update api-query, add REPEAT, CONCURRENCY vars

pflanze:
Also check early for COMMIT_ID.

### Summary

This will allow to use small (e.g. manually generated) silo_queries.ndjson files and still have enough queries to get reliable results.

Requires a change in etc/evobench-run.ron to add the REPEAT custom variable (with value 50 for compatibility).

Not sure where to document this, if at all.

## PR Checklist


Taepper (approved):
LGTM

Issue #840, opened 2025-06-29
Add action that returns a phylogenetic tree of all sequences returned by a filter


anna-parker:

Taepper:
I would kind of dislike returning a tree in a single row. Are there no table-like-representations?

anna-parker:
> Are there no table-like-representations?

Not really, fundamentally a tree is a graph structure - the closest to a table would be an edge list. But a nwk or json would be a more compact and standard representation

chaoran-chen:
This feature would be very useful!

chaoran-chen:
Apart from #909, all the initial tickets have been resolved, so I'm closing this epic now.

PR #902, opened 2025-07-21
feat(silo): add an endpoint to export a newick of the nodes in the filter

### LINKED ISSUE #908: Return a simple subtree nwk

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS-SILO/issues/908, https://github.com/GenSpectrum/LAPIS-SILO/issues/910

1. Updates the serialization by adding the root_id to the phylo tree class
2. Adds functions to create a newick from a phylo tree where only nodes in the filter should be seen in the newick
3. Adds tests

## PR Checklist


anna-parker:
@Taepper do you know why the build is failing?
I see a warning about 
```
#9 102.2 /src/src/silo/query_engine/actions/action.h:40:29: note:     ‘virtual std::string_view silo::query_engine::actions::Action::getType() const’
```
but I don't see this function in that file...

Taepper:
> Update: it seems the CI runs all code on top of main - so I rebased and fixed this now

Whoa, this would be surprising, or I did misunderstand?

src/silo/query_engine/actions/subtree.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+std::vector<schema::ColumnIdentifier> Subtree::getOutputSchema(
+   const schema::TableSchema& table_schema
+) const {
+   std::vector<schema::ColumnIdentifier> fields;
+   fields.emplace_back("subtreeNewick", schema::ColumnType::STRING);
+   fields.emplace_back("missingNodeCount", schema::ColumnType::INT);
+   if (print_nodes_not_in_tree) {
+      fields.emplace_back("missingFromTree", schema::ColumnType::STRING);
```
Taepper, line 173: Most of this code is shared with the mrca action. Even the calculation of the values for the fields `missingNodeCount`, `missingFromTree`. I think this can be refactored to a shared parent class `TreeAction` or something similar, where most of the validation takes place. The subclasses mrca and subtree then add one additional field to the output-schema, and override an abstract method from the class `TreeAction` that calculates this extra field. What do you think?

taepper (changes requested):

src/silo/query_engine/actions/subtree.cpp
anna-parker: Done :-) (got chatGPT to help me with the templates!)

src/silo/common/phylo_tree.test.cpp
```diff
@@ -163,3 +167,19 @@ TEST(PhyloTree, correctlyReturnsMRCA) {
       mrca_response.not_in_tree[1] == "NOT_IN_TREE2"
    );
 }
+
+TEST(PhyloTree, correctlyReturnsSubTreeNewick) {
+   auto phylo_tree =
+      PhyloTree::fromNewickString("(((A1.1, A1.2)A1,(A2.1)A2)A,(B1,(B2.1,B2.2)B2)B)R;");
+   auto subtree_left_side = phylo_tree.toNewickString({"A1.1", "A1.2", "A2.1"}).newick_string;
+   ASSERT_EQ(subtree_left_side, "((A1.1,A1.2)A1,(A2.1)A2)A;");
```
Taepper, line 175: Hmm, it could be argued that we should do branch compression (i.e. the test should omit A2, as it does not provide any topological information):

`"((A1.1,A1.2)A1,A2.1)A;"`

endToEndTests/test/invalidQueries/Subtree_invalidColumn.json
```diff
@@ -15,6 +15,6 @@
   },
   "expectedError": {
     "error": "Bad request",
-    "message": "Subtree action cannot be called on Column 'country' as it does not have a phylogenetic tree associated with it"
+    "message": "TreeAction cannot be called on Column 'country' as it does not have a phylogenetic tree associated with it"
```
Taepper, line 18: I think we should not leak `TreeAction` which is an implementation detail to the user (LAPIS).

Is it a lot of effort to provide an overloaded `getActionName()` method to the TreeActions?

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,127 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
+      if (i > 0)
+         result += ",";
+      result += child_newick_strings[child_newick_strings.size() - i - 1];
+   }
+   result += self_id.has_value() ? ")" + self_id.value() : "";
+   return result;
+}
+
+PartialNewickResponse PhyloTree::partialNewickString(
+   const std::vector<std::string>& filter,
```
Taepper, line 452: Would the logic in the function simplify a lot, if we first check that all nodes are contained in the tree? (i.e. `TreeNodeId`s are passed to this function instead of strings?)

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,127 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
+   const TreeNodeId& ancestor
+) const {
+   PartialNewickResponse response;
+
+   auto node_it = nodes.find(ancestor);
+   if (node_it == nodes.end()) {
+      throw std::runtime_error(fmt::format("Ancestor '{}' not found in tree.", ancestor.string));
```
Taepper, line 459: Can you add an e2e test that triggers one of the exceptions in the class? I am currently not sure, whether they will return the correct http status code?

Generally, I would prefer typed exceptions in this class (which also remove the `..exception(fmt::format .. )` boilderplate by providing a corresponding constructor)

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,127 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
+   if (responses.empty()) {
+      response.newick_string_with_self = std::nullopt;
+      response.newick_string_without_self = std::nullopt;
+      return response;
+   }
+   if (responses.size() == 1) {
+      if (responses[0].newick_string_without_self.has_value()) {
+         response.newick_string_without_self = responses[0].newick_string_without_self.value();
+      } else {
+         response.newick_string_without_self = responses[0].newick_string_with_self.value_or("");
```
Taepper, line 484: Why can't we return std::nullopt?

src/silo/common/phylo_tree.test.cpp
anna-parker: I asked about this in a meeting a couple weeks ago but it didn't seem like there were too strong opinions (unless I forgot) - as I think @chaoran-chen wants to show this on Monday is it ok to not compress the branches for the moment? 

I can then update to add 2 options when downloading one with compressed and one with internal nodes

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,127 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
+   });
+   if (responses.empty()) {
+      throw std::runtime_error(
+         fmt::format("Newick should contain nodes '{}' but returned <empty>.", in_tree)
+      );
+   }
+   if (responses.size() == 1) {
+      // Only one child has nodes in the filter, root is not needed
+      // The output nwk's root should be the MRCA of the nodes in the filter
+      if (not responses[0].newick_string_without_self.has_value()) {
```
Taepper, line 534: the `not` operator is very uncommon to use in cpp. I would prefer to stay consistent here

src/silo/query_engine/actions/subtree.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+using silo::query_engine::filter::operators::Operator;
+
+arrow::Status Subtree::addResponseToBuilder(
+   std::vector<std::string>& all_node_ids,
+   std::unordered_map<std::string_view, exec_node::JsonValueTypeArrayBuilder>& output_builder,
+   const storage::column::StringColumnMetadata* metadata,
+   bool print_nodes_not_in_tree
+) const {
+   NewickResponse response = metadata->toNewickString(all_node_ids);
```
Taepper, line 41: Do we need the member function in metadata, I think it pollutes the interface? Can we not just call it directly on `metadata->phylo_tree`. And then we can return a 500 directly, as would be appropriate if `phylo_tree` is `nullopt`

src/silo/common/phylo_tree.cpp
anna-parker: Do you mean check that all filter strings are in the tree? I already do that in `toNewickString`

src/silo/common/phylo_tree.cpp
anna-parker: I would need to rewrite `newickJoin` then - but I can also do that

src/silo/common/phylo_tree.test.cpp
anna-parker: Follow up issue created in https://github.com/GenSpectrum/LAPIS-SILO/issues/840?issue=GenSpectrum%7CLAPIS-SILO%7C910

src/silo/query_engine/actions/subtree.cpp
anna-parker: So you mean to remove the getMRCA and toNewickString definitions from the `string_column.h`?

src/silo/common/phylo_tree.cpp
anna-parker: I honestly wouldn't know how to trigger this exception in an e2e - it should never be raised as it would mean that the tree structure is invalid - I could raise it in a unit test?

anna-parker: I switched them to SILO_ASSERT? Is that ok?

src/silo/common/phylo_tree.cpp
Taepper: But we have the function `getTreeNodeId` for this purpose I thought? And then we should keep the strings as `TreeNodeId` to show with the type, that their containment has already been checked

src/silo/common/phylo_tree.cpp
Taepper: Ah, I see, it was not obvious to me that they were assertions, but that clears it up

src/silo/query_engine/actions/subtree.cpp
Taepper: moving `getMRCA` would be cleaner as well I think, but here specifically the `if contains` check is redundant because that is already checked in `subtree.cpp`. I think it would make it much clearer, if it only needed to be checked once, at the place where we anyways validate the user query and then immediately bind the phylo_tree to a variable, which can then be used subsequently

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,128 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
 
+bool isInFilter(const std::string& str, const std::vector<std::string>& filter) {
+   return std::find(filter.begin(), filter.end(), str) != filter.end();
+}
+
+std::optional<std::string> newickJoin(
+   const std::vector<std::optional<std::string>>& child_newick_strings,
+   std::optional<std::string> self_id = std::nullopt
+) {
+   std::string result = self_id.has_value() ? "(" : "";
```
Taepper, line 441: Wouldn't the body be much easier, if we first filter for `has_value`, and then make a `fmt::join`? And why do we need to reverse the child order?

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,128 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
+   std::erase_if(responses, [](const PartialNewickResponse& resp) {
+      return !resp.newick_string_with_self.has_value();
+   });
+   SILO_ASSERT(!responses.empty());
+   if (responses.size() == 1) {
+      // Only one child has nodes in the filter, root is not needed
+      // The output nwk's root should be the MRCA of the nodes in the filter
+      if (!responses[0].newick_string_without_self.has_value()) {
+         // The child is the MRCA of the nodes in the filter and should be the root of the nwk
+         response.newick_string = responses[0].newick_string_with_self.value_or("") + ";";
```
Taepper, line 537: Is this `value_or` here correct? What would an empty string here represent?

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,128 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
...
+         return response;
+      }
+   }
+   // Multiple children have nodes in the filter, the root is the MRCA of the nodes in the filter
+   std::vector<std::optional<std::string>> strings;
+   strings.reserve(responses.size());
+   for (const auto& resp : responses) {
+      strings.push_back(resp.newick_string_with_self);
+   }
+   response.newick_string = newickJoin(strings, root_id.string).value_or("") + ";";
```
Taepper, line 553: if `newickJoin` never returns `std::nullopt` and its signature is change accordingly, we can also just call `value`

src/silo/common/phylo_tree.cpp
anna-parker: Because when we read in the newick it is reversed order - I can add a commment. This doesn't change the tree structure but is useful for tests

src/silo/common/phylo_tree.cpp
```diff
@@ -428,4 +430,128 @@ MRCAResponse PhyloTree::getMRCA(const std::vector<std::string>& node_labels) con
    return response;
 }
 
+bool isInFilter(const std::string& str, const std::vector<std::string>& filter) {
+   return std::find(filter.begin(), filter.end(), str) != filter.end();
+}
+
+std::optional<std::string> newickJoin(
```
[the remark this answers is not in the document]
anna-parker: Ah true - I initially used the function in more places but now it indeed will always return a string - I refactored the code :-)

src/silo/common/phylo_tree.cpp
anna-parker: I will need to do this as the type is optional - but as I filtered out nullopts I now added an assert

src/silo/common/phylo_tree.test.cpp
anna-parker: Actually this was very straight forward to add - so I did this now. This now means that there is one more option in the action - adding one json field only for one action and not the other made the `makeTreeAction` template very complicated so I again just split it up - but I think we have still managed to minimize a lot of code duplication :-)

src/silo/query_engine/actions/subtree.cpp
anna-parker: Done :-)

PR #911, opened 2025-07-27
feat(silo): add an endpoint to export a newick of the nodes in the filter

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS-SILO/issues/910, https://github.com/GenSpectrum/LAPIS-SILO/issues/908

2. Adds functions to create a newick from a phylo tree where only nodes in the filter should be seen in the newick
3. Adds tests

## PR Checklist

### DIFF
--- a/documentation/phylogenetic_queries.md
+++ b/documentation/phylogenetic_queries.md
@@ -49,3 +49,28 @@ The result of such a query is a ndjson with a single row, where `missingFromTree
   "missingFromTree": "MISSING_NODE1,MISSING_NODE2"
 }
 ```
+
+### Subtree
+
+```json
+"action": {
+  "type": "Subtree",
+  "columnName": "COLUMN_NAME",
+  "printNodesNotInTree": true,
+  "contractUnaryNodes": true
+}
+```
+
+Returns a subtree of all sequences in the filter it is applied to in newick format (`.nwk`). The subtree is produced from the phylogenetic tree, leaf nodes are only shown in the newick if they were included in the filter. Internal nodes are included if they are part of the subtree created from the leaves in the filter, where the root is the most recent common ancestor that includes all nodes. To contract unary internal nodes in the output newick set `contractUnaryNodes` to true.
+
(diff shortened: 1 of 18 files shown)


Taepper (approved):
LGTM! I like this a lot better, and it is the interface I would have expected (esp. for the `partialNewickResponse`)! The only question I still have is whether it makes sense for a user that the tree is always rooted in the MRCA. If `contractUnaryNodes` is false, I would have expected the true tree root as the root of the exported newick. In the `contractUnaryNodes=true` case, it will become the MRCA anyways, just by the logic of contracting.

But this can be decided separately and I added #912

anna-parker:
> LGTM! I like this a lot better, and it is the interface I would have expected (esp. for the `partialNewickResponse`)! The only question I still have is whether it makes sense for a user that the tree is always rooted in the MRCA. If `contractUnaryNodes` is false, I would have expected the true tree root as the root of the exported newick. In the `contractUnaryNodes=true` case, it will become the MRCA anyways, just by the logic of contracting.
> 
> But this can be decided separately and I added #912

The root of a subtree will always be the MRCA of the subtree's leaves by definition. 

But yes it isn't very well defined if in the case where we do not contract the unary nodes we should add the nodes leading up to the original root after the MRCA. I think it might make sense to switch `contractUnaryNodes=true` to the default actually as it does seem increasingly like this is the case that users will use as the `contractUnaryNodes=false` case is not well defined. 

The main reason I kept the unary nodes for the subtree was to better preserve the structure of the subtree. But we could actually do this better by keeping the branch lengths of the imported tree and then adding them together for branches where unary nodes are contracted - this might actually be the better idea! I will make an issue for that!

documentation/phylogenetic_queries.md
```diff
@@ -49,3 +49,28 @@ The result of such a query is a ndjson with a single row, where `missingFromTree
   "missingFromTree": "MISSING_NODE1,MISSING_NODE2"
 }
 ```
+
+### Subtree
+
+```json
+"action": {
+  "type": "Subtree",
```
anna-parker, line 57: @Taepper and @chaoran-chen before I merge this is this a good name for this action or should I maybe rename it to sth more clear like `PhyloSubtree`?

anna-parker: We discussed and decided on PhyloSubtree! :-)

anna-parker:
> Achieving this would just mean removing these lines, right?:

As we don't store the root at the moment - we would need to calculate it first, but yes in theory if we calculate/know the root then we could just replace the MRCA with the root and apply the functions there for that case

PR #927, opened 2025-08-04
    (benchmarking) feat: add a separate target for the preprocessing

### LINKED ISSUE #919: Benchmarking: add a separate target for the preprocessing
Currently, we're capturing benchmarking data from the new probes we added to the preprocessing via a hack, https://github.com/GenSpectrum/evobench/commit/f21b8dc646d9e4ee23e9a7398e3bbca09d557b64

But, this..

- prevents getting multiple measurements to build a summary over as with the normal data
- is a hack and hurts future work
- the "full" dataset causes so much preprocessing logging that it fills up /dev/shm on staging

Evobench now has a feature (https://github.com/GenSpectrum/evobench/issues/20, https://github.com/GenSpectrum/evobench/issues/22) that allows configuring separate targets for different purposes. Let's actually use it. So:

- Add a new target, "make bench_preprocessing" (or perhaps a separate Makefile), that just runs the preprocessing (with or without benchmark logging, depending on whether evobench sets the EVOBENCH_LOG env var). Make it *always* run the preprocessing even if already cached, to allow for re-running for statistical significance (but store the output so that it *is* picked up by the normal "make bench").

- Change the [configuration](https://github.com/GenSpectrum/silo-benchmark-ci/blob/master/etc/evobench-run.ron) to have separate `JobTemplate`s for running the preprocessing; only do so for the `west_nile` dataset, not `full`, to avoid running into the /dev/shm issue. (Might want to only run it in the night, as it's long-running jobs; that would/will need an evobench feature to specify the queue pipeline per job, though.)

- Do *not* change that running the "make bench" target still runs the preprocessing *with* caching, but change that way of running the preprocessing to clear the EVOBENCH_LOG env var (so that it is compatible with the "full" dataset)

pflanze:
resolves #919

Allowing for separate benchmarking runs via "make preprocessing" and "make api". The latter still also runs the preprocessing if necessary, but it doesn't collect benchmarking data for it (the evobench probes are disabled by deleting the EVOBENCH_LOG env var). "make preprocessing" however will leave the files around so that a subsequent "make api" can use them, saving a preprocessing run.

Done by splitting off 2 new make files, one per target. That way, each target can do its own checking for required variables (the set of "custom variables" in evobench terminology differs for the two targets), also it keeps them less cluttered. The main "Makefile" is still to be used as entry point, as silo is still built here, and the logic for determining the input file paths is still determined here, too.

### DIFF
--- a/benchmarking/Makefile
+++ b/benchmarking/Makefile
@@ -1,3 +1,4 @@
+all: please_give_target_name__api_or_preprocessing
 
 # Required env vars:
 
@@ -9,7 +10,8 @@ ifeq ($(BENCH_OUTPUT_LOG),)
 $(error "BENCH_OUTPUT_LOG must be set")
 endif
 
-# Our custom parameters:
+# Our custom parameters (the shared ones, there can be more checks in
+# the `Makefile_*` files):
 
 # The name of a subfolder of `~/silo-benchmark-datasets/` where the
 # dataset is stored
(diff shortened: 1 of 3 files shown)


pflanze:
The matching changes in evobench are https://github.com/GenSpectrum/evobench/commit/89d0ecd9e0b7e8d44469652f3fd876d8f86f8997 and https://github.com/GenSpectrum/evobench/commit/57d4a6f991416047fb07f40ab3e7696d3c0da6d4

The matching changes in the evobench configuration are: https://github.com/GenSpectrum/silo-benchmark-ci/compare/868498a7b4892b9b91e93c9477fe3236800af593..edf5ef6bb66611ef0833d1a2fd842f38a7764112

Taepper (approved):
LGTM

PR #934, opened 2025-08-06
Bugfix – limit without order

Taepper:
### Summary

SILO silently crashed requests when a limit was provided without orderByFields in an aggregated query. This fixes that problem and also adds a arrow plan validation before the plan is executed

### DIFF
--- a/src/silo/query_engine/actions/action.cpp
+++ b/src/silo/query_engine/actions/action.cpp
@@ -377,6 +377,12 @@ arrow::Result<arrow::acero::ExecNode*> Action::addLimitAndOffsetNode(
    arrow::acero::ExecNode* node
 ) const {
    if (limit.has_value() || offset.has_value()) {
+      CHECK_SILO_QUERY(
+         !node->ordering().is_unordered(),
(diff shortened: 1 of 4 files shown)


src/silo/query_engine/actions/action.cpp
```diff
@@ -377,6 +377,12 @@ arrow::Result<arrow::acero::ExecNode*> Action::addLimitAndOffsetNode(
    arrow::acero::ExecNode* node
 ) const {
    if (limit.has_value() || offset.has_value()) {
+      CHECK_SILO_QUERY(
+         !node->ordering().is_unordered(),
+         "Offset and limit can only be applied if the output of the operation has some ordering. "
+         "Implicit ordering such as in the case of TableScans is also allowed. E.g. HashAggregates "
```
anna-parker, line 383: I wonder if this error message is clear enough for users - how can a TableScan or hashaggregate be called with a silo query?

anna-parker (approved):

src/silo/query_engine/actions/action.cpp
Taepper: True, I will reword to use more user-facing constructs

Issue #926, opened 2025-08-04
Phylo missingNodeCount should include nulls


chaoran-chen:
Currently, `missingNodeCount` in the MRCA response counts the number of sequences that have an value in the phylo column that cannot be matched to the tree but not those for which the values in the phylo column is null. I think that the nulls should also be counted, i.e., that the `missingNodeCount` should be `"the number of sequences matching the filter" - "the number of sequences found on the tree"`. This value is needed for calculating the "cladeness".

PR #936, opened 2025-08-07
feat(silo): include null values in phyloTree column in missingNodeCount

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS-SILO/issues/926

### Summary
1. Adds docs on how the phylogenetic trees were created for the e2e tests.
2. Changes the e2e configuration by adding a new field `usherTree` which should be used as the `phyloTreeField` instead of the primaryKey - the tests had to be modified to now include this new field in the expected result.
3. @chaoran-chen discovered that in the case where there were NULL values in the phyloTree column these are not included in the missing Node count - I have now added these - however, we offer the option to print missing nodes - we print the name of the nodes in the `usherTree` field - thus for NULL values these are not added to the list. Is this intended behavior? Should we create a new field for NULL values instead of adding this to the missing node count?

## PR Checklist

### DIFF
--- a/src/silo/query_engine/actions/tree_action.cpp
+++ b/src/silo/query_engine/actions/tree_action.cpp
@@ -58,7 +58,7 @@ NodeValuesResponse TreeAction::getNodeValues(
       num_rows += filter->cardinality();
    }
    std::unordered_set<std::string> all_tree_node_ids;
-   int num_empty = 0;
+   uint32_t num_empty = 0;
    all_tree_node_ids.reserve(num_rows);
    for (size_t i = 0; i < table->getNumberOfPartitions(); ++i) {
       const storage::TablePartition& table_partition = table->getPartition(i);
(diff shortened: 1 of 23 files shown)


src/silo/query_engine/actions/tree_action.cpp
```diff
@@ -58,6 +58,7 @@ std::unordered_set<std::string> TreeAction::getNodeValues(
       num_rows += filter->cardinality();
    }
    std::unordered_set<std::string> all_tree_node_ids;
+   int num_empty = 0;
```
Taepper, line 61: For consistency, I would use the unsigned integer type `uint32_t`

src/silo/query_engine/actions/tree_action.h
```diff
@@ -19,6 +19,12 @@
 
 namespace silo::query_engine::actions {
 
+class NodeValuesResponse {
+  public:
+   std::unordered_set<std::string> node_values;
+   int32_t missing_node_count = 0;
```
Taepper, line 25: I would use `uint32_t`

Taepper (approved):
LGTM

PR #940, opened 2025-08-08
build: always export compile commands

Taepper:
Export compile commands for configuring language servers or debugging the build.

It does have non noticable over-head on project setup


pflanze:
Approved, although it would be good to know when it runs, only on the first running of cmake, or on every build? Maybe it would be useful to know and how much time it takes to run (it's much below 10 seconds AFAIR but if it runs on every build we better be aware)

Taepper:
It runs on every cmake configuration, so not on every build. Also it takes a lot less than a second. For the full configuration:

Before:
Min: 0m0.794s
Max: 0m0.938s

Now:
Min: 0m0.794s
Max: 0m0.925s


so no concern at all

PR #941, opened 2025-08-08
feat(silo): add a new phylo parents endpoint

### LINKED ISSUE #924: Get parent node in phylogenetic tree
It would be useful to be able to get the parent node ID of a node. A use case is to allow users to navigate up a tree in CoV-Spectrum.

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS-SILO/issues/924

### Summary
See documentation for details

## PR Checklist


Taepper:
I think this issue has mostly been resolved by #948. If we see that we have further needs, we can open another issue

PR #942, opened 2025-08-10
feat: add a default timeout of 2 minutes between sending response batches of responses

taepper:
### Summary

As #934 showed. A misconfiguration of an arrow plan lead to a hanging plan execution that never finished. The awaiting for new results of a plan allows for an offset in seconds. This PR adds that offset with a default value of 120 seconds. Note that this timeout does not apply to the total time of the query but the timeout for awaiting the next batch of the result. In particular this still allows for large downloads that may take longer than 2 minutes due to slow connections.

### DIFF
--- a/src/silo/query_engine/query_plan.cpp
+++ b/src/silo/query_engine/query_plan.cpp
@@ -24,7 +24,10 @@ arrow::Result<QueryPlan> QueryPlan::makeQueryPlan(
    return query_plan;
 }
 
-arrow::Status QueryPlan::executeAndWriteImpl(std::ostream* output_stream) {
+arrow::Status QueryPlan::executeAndWriteImpl(
+   std::ostream* output_stream,
+   uint64_t timeout_in_seconds
+) {
    EVOBENCH_SCOPE("QueryPlan", "execute");
    SPDLOG_TRACE("{}", arrow_plan->ToString());
    arrow_plan->StartProducing();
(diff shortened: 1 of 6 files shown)


corneliusroemer:
> > Do you log when the time out is hit? Would be good to be able to review logs for how often the timeout fires
>
> Oh, I am afraid that this is not logged but only send as a response to the user instead. It will contain a status code 400, which will be logged, but I do not believe that the response body is logged.

Hmm shouldn't this be a 500 Error?

Taepper:
>> Oh, I am afraid that this is not logged but only send as a response to the user instead. It will contain a status code 400, which will be logged, but I do not believe that the response body is logged.

> Hmm shouldn't this be a 500 Error?

Ah, yes, this is a typo here.

corneliusroemer:
OK I think it makes sense to limit the generator to 120s. The generator should return a chunk quickly and not time out when the client throttles because the timeout only applies to arrow not to the response in ndjson. Only a very slow (or neverending) query would time out, if it takes >120s to get the the next chunk of data.

src/silo/query_engine/query_plan.cpp
```diff
@@ -24,14 +24,24 @@ arrow::Result<QueryPlan> QueryPlan::makeQueryPlan(
...
    EVOBENCH_SCOPE("QueryPlan", "execute");
    SPDLOG_TRACE("{}", arrow_plan->ToString());
    arrow_plan->StartProducing();
    SPDLOG_TRACE("Plan started producing, will now read the resulting batches.");
    while (true) {
       arrow::Future<std::optional<arrow::ExecBatch>> future_batch = results_generator();
       SPDLOG_TRACE("await the next batch");
+      bool finished_batch_in_time = future_batch.Wait(timeout_in_seconds);
+      if (!finished_batch_in_time) {
+         return arrow::Status::ExecutionError(fmt::format(
```
corneliusroemer, line 40: Based on some docs reading, you might want to call `StopProducing()` before returning: https://arrow.apache.org/docs/cpp/api/acero.html#_CPPv4N5arrow5acero8ExecPlan13StopProducingEv

Issue #929, opened 2025-08-05
Add an `IsNull` and `IsNotNull` filter


taepper:
This filter should work on all columns

Issue #385, opened 2024-04-10
Make boolean columns indexed


fengelniederhammer:
https://github.com/GenSpectrum/LAPIS-SILO/issues/219 introduced boolean columns. Make them indexed (there should only be indexed boolean columns).

We know the values (`null`, `true`, `false`), so every columns would only need to store three bitmaps.

PR #953, opened 2025-08-25
refactor: organise string column data in pages

### LINKED ISSUE #930: Do not interpret empty string as `NULL`
Currently the empty string is interpreted as the value `NULL` (this is a remnant from the tsv input format). This should be changed

taepper:
This PR cleans up some of the code and is a necessary prerequisite for adding out-of-memory capabilities to SILO. (might be necessary for our Wastewater instance)

### Summary

First, this changes the string representation of SILO a bit. GermanStrings are now named and implemented closer to how the source introduces them. Also the storage of these strings is now no longer in a literal `std::vector`, but instead stored on byte buffers of size 16 KB (pages), which are designed such that they can be stored to and loaded from files. This is necessary for Out-Of-Memory capabilities.

Second, as this changed removed `"" for NULL` in string columns, additional null bitmaps are introduced to a string columns and other column types. This unifies the implementation and almost implements #929 already.

Also resolves #930, #385

## PR Checklist


src/silo/common/data_version.h
chaoran-chen, line 1: Would you like to briefly describe what the changes? I think that this is the first time that I heard about "pages" in SILO and would be interested to learn more!

chaoran-chen: thanks!

PR #954, opened 2025-08-25
Update evobench-probes library, change costly probe to a sparse ('statistical') one

### LINKED ISSUE #950: benchmarking: remove probe "QueryPlan|sendDataToOutputStream" again due to being too hot
We again have a probe that causes too much data collection (the last time it was in the preprocessing, which is OK for profiling the preprocessing with the west_nile dataset, so we left it in and made it possible to only benchmark the preprocessing with that dataset): this time it is a probe named "QueryPlan|sendDataToOutputStream" added in commit 4c97982b31de07123f8cdd2a210cf299ad83eb86; it allocates "only" about 20% of /dev/shm, so that's not a problem, but it caused high RAM usage while the evaluator was running (now worked around by not using parallelism in evobench-evaluator any more, which makes it slower, but caching is coming which will make it fast again), but the collected files are also ~800 MB each, compressed, thus will fill up our server disk quite soon.

Alex said this probe can be removed again, so I will do so.

If adding probes that produce too much data keeps being a recurring topic, then https://github.com/GenSpectrum/evobench/issues/24 could be done, but it will be a bit of work.

pflanze:
resolves #950

### Summary

There is a new feature in evobench, described at https://github.com/GenSpectrum/evobench/issues/24#issuecomment-3221833344 ; this updates the copy of the evobench-probes library in SILO to that, and changes the hot probe (mentioned in issue #950, and also described in the linked comment) to be such a new sparse one. This solves the evaluation cost (disk space, RAM use, and benchmarking cpu time), and appears to actually make this probe useful, thus leaving it in in this new form.

## PR Checklist

### DIFF
--- a/src/evobench/evobench.cpp
+++ b/src/evobench/evobench.cpp
@@ -4,6 +4,7 @@
 
 #include "evobench/evobench.hpp"
 
+#include <charconv>
 #include <cstdint>
 #include <cstring>
 #include <iomanip>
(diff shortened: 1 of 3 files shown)


pflanze:
PS. this tries to optimizes the log writing, too (commit 4bebe4bc6b3112ba4ee0e1021b832887248cecef), I haven't checked how much it helps though (can always do that later by reverting this separate commit).

PR #962, opened 2025-09-01
benchmarking: preprocessing sc2open

pflanze:
### Summary

Benchmarking of SCov2 was impossible with the existing probes because it generated too much data (so much so that it filled the /dev/shm tmpfs).

Change the hot probes to statistical ones so that benchmarking of preprocessing of SCov2 works.

## PR Checklist

### DIFF
--- a/src/silo/append/database_inserter.cpp
+++ b/src/silo/append/database_inserter.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 
+#include "evobench/evobench.hpp"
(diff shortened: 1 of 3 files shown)


pflanze:
Note though that I had to divide down so much that now west_nile has only around 100 measurements for these probes.

It's possible to feed a runtime value for `n` to the `EVOBENCH_PROBE_EVERY` macro--could take an environment variable to override (with some default) (where to initialize though?).

PR #963, opened 2025-09-01
Remove TBB

### LINKED ISSUE #873: Remove TBB
Because it is discouraged to link TBB statically, is our only dynamically linked library. (Conan even removed the option to specify static linkage for tbb at some point..)

We want to link all libraries statically to generate a binary that is easier to use.

Over time, our usage of tbb decreased a lot. There are only 3 files left, that use tbb:
- `src/silo/storage/column/sequence_column.cpp`
- `src/silo/query_engine/actions/mutations.cpp`
- `src/silo/storage/table.cpp`

These usages should be removed, benchmarked for performance degradation, and different parallelism constructs introduced accordingly.

pflanze:
resolves #873

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ endif ()
 
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
-set(CMAKE_CXX_FLAGS "-Wall")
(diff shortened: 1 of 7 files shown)


pflanze:
Obsoleted by #985

PR #975, opened 2025-09-09
[DO NOT MERGE] V0.7.10.1

### LINKED ISSUE #838: Add timeout for arrow queries to avoid getting stuck in a non-finishing query
In https://github.com/GenSpectrum/LAPIS-SILO/pull/815#issuecomment-3010155049 I had to restructure the query because if an error was raised during the execution of a filter query SILO was stuck.

From discussion with @Taepper this was due to a refactoring where the compile filter expression action was added as an arrow query node. It seems if an exception occurs while executing an arrow node it is undefined behavior and arrow will not complete. 

Because of this e2e tests were stuck up to 5hours. We never want this to happen on a running SILO instance - even though we have refactored the code if a query in an arrow node fails we could get stuck again and this will end up eating up SILO threads. We should add a timeout for arrow queries or figure out how to make them error. 

@Taepper please feel free to update this PR with more details about this issue.

taepper:
backported bug fix #838

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,6 +164,7 @@ jobs:
           images: ${{ env.DOCKER_IMAGE_NAME }}
           tags: |
             type=ref,event=pr
+            type=raw,value=${{ github.head_ref }}
             type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
             type=raw,value=commit-${{ env.HEAD_SHA }}
             type=semver,value=${{ env.SILO_VERSION }},enable=${{ env.SILO_VERSION != '' }},pattern={{version}}


src/silo/query_engine/query_plan.cpp
```diff
@@ -24,14 +24,24 @@ arrow::Result<QueryPlan> QueryPlan::makeQueryPlan(
...
    EVOBENCH_SCOPE("QueryPlan", "execute");
    SPDLOG_TRACE("{}", arrow_plan->ToString());
    arrow_plan->StartProducing();
    SPDLOG_TRACE("Plan started producing, will now read the resulting batches.");
    while (true) {
       arrow::Future<std::optional<arrow::ExecBatch>> future_batch = results_generator();
       SPDLOG_TRACE("await the next batch");
+      bool finished_batch_in_time = future_batch.Wait(timeout_in_seconds);
+      if (!finished_batch_in_time) {
+         return arrow::Status::ExecutionError(fmt::format(
```
anna-parker, line 40: would be good to log the message right here

PR #981, opened 2025-09-09
Suggestion for reduced chance of stall, better cleanup, better logging

corneliusroemer:
I don't know much Cpp nor Arrow but some reading of docs and interaction with ChatGPT makes me think that there is scope of improvement here.

E.g. it seems you still don't always have timeouts when you wait? The debug log mentions a race condition - what guarantees that this race condition won't cause a deadlock?

resolves #

## PR Checklist

### DIFF
--- a/src/silo/query_engine/query_plan.cpp
+++ b/src/silo/query_engine/query_plan.cpp
@@ -28,19 +28,28 @@ arrow::Status QueryPlan::executeAndWriteImpl(
    std::ostream* output_stream,
    uint64_t timeout_in_seconds
 ) {
+   constexpr double kGraceShutdownSeconds = 5.0;  // avoid hanging on teardown
    EVOBENCH_SCOPE("QueryPlan", "execute");
    SPDLOG_TRACE("{}", arrow_plan->ToString());
    arrow_plan->StartProducing();
    SPDLOG_TRACE("Plan started producing, will now read the resulting batches.");
+
+   // Ensure plan is stopped on any exit path (timeout/error/exception).
+   struct PlanStopGuard {
+      std::shared_ptr<arrow::acero::ExecPlan> plan;
+      ~PlanStopGuard() { if (plan) plan->StopProducing(); }
+   } guard{arrow_plan};
+
    while (true) {
       arrow::Future<std::optional<arrow::ExecBatch>> future_batch = results_generator();
       SPDLOG_TRACE("await the next batch");
-      bool finished_batch_in_time = future_batch.Wait(timeout_in_seconds);
+      bool finished_batch_in_time = future_batch.Wait(static_cast<double>(timeout_in_seconds));
       if (!finished_batch_in_time) {
-         return arrow::Status::ExecutionError(fmt::format(
-            "Request timed out, could not detect any progress within {} seconds.",
-            timeout_in_seconds
-         ));
+         SPDLOG_WARN("Batch wait timed out after {} s — stopping plan.", timeout_in_seconds);
+         arrow_plan->StopProducing();  // best-effort cancel
+         (void)arrow_plan->finished().Wait(kGraceShutdownSeconds);  // bounded drain
+         return arrow::Status::ExecutionError(
+            fmt::format("Request timed out, no batch within {} seconds.", timeout_in_seconds));
       }
       ARROW_ASSIGN_OR_RAISE(std::optional<arrow::ExecBatch> optional_batch, future_batch.result());
       SPDLOG_TRACE("Batch received");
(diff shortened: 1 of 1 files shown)


src/silo/query_engine/query_plan.cpp
```diff
@@ -28,27 +28,36 @@ arrow::Status QueryPlan::executeAndWriteImpl(
...
+            fmt::format("Request timed out, no batch within {} seconds.", timeout_in_seconds));
       }
       ARROW_ASSIGN_OR_RAISE(std::optional<arrow::ExecBatch> optional_batch, future_batch.result());
       SPDLOG_TRACE("Batch received");
 
       if (!optional_batch.has_value()) {
          break;  // end of input
       }
-      SPDLOG_TRACE("Batch contains data with {} values.", optional_batch.value().length);
+      SPDLOG_TRACE("Batch contains {} rows.", optional_batch->length);
```
corneliusroemer, line 60:
```suggestion
      SPDLOG_TRACE("Batch contains data with {} values.", optional_batch.value().length);
```

src/silo/query_engine/query_plan.cpp
```diff
@@ -28,27 +28,36 @@ arrow::Status QueryPlan::executeAndWriteImpl(
...
+   struct PlanStopGuard {
+      std::shared_ptr<arrow::acero::ExecPlan> plan;
+      ~PlanStopGuard() { if (plan) plan->StopProducing(); }
+   } guard{arrow_plan};
+
    while (true) {
       arrow::Future<std::optional<arrow::ExecBatch>> future_batch = results_generator();
       SPDLOG_TRACE("await the next batch");
-      bool finished_batch_in_time = future_batch.Wait(timeout_in_seconds);
+      bool finished_batch_in_time = future_batch.Wait(static_cast<double>(timeout_in_seconds));
```
corneliusroemer, line 46:
```suggestion
      bool finished_batch_in_time = future_batch.Wait(timeout_in_seconds);
```

src/silo/query_engine/query_plan.cpp
```diff
@@ -57,20 +66,17 @@ arrow::Status QueryPlan::executeAndWriteImpl(
...
    SPDLOG_TRACE("Finished reading all batches.");
    auto future = arrow_plan->finished();
-   if (future.state() == arrow::FutureState::PENDING) {
-      SPDLOG_DEBUG(
-         "Plan still pending, but no more results in the future? Not a problem in itself as it "
-         "could be a race condition so the next await will clear this, but be a hint in case we "
-         "detect some weird behavior."
-      );
-   }
-   future.Wait();
```
corneliusroemer, line 67: This wait could be a problem, if there's a deadlock it'll hang forever. An await doesn't fix a deadlock.

src/silo/query_engine/query_plan.cpp
```diff
@@ -32,15 +32,42 @@ arrow::Status QueryPlan::executeAndWriteImpl(
    SPDLOG_TRACE("{}", arrow_plan->ToString());
    arrow_plan->StartProducing();
    SPDLOG_TRACE("Plan started producing, will now read the resulting batches.");
+
+   // Ensure plan is stopped on any exit path (timeout/error/exception).
+   struct PlanStopGuard {
+      std::shared_ptr<arrow::acero::ExecPlan> plan;
+      ~PlanStopGuard() {
+         constexpr double kGraceShutdownSeconds = 5.0;  // avoid hanging on teardown
```
fengelniederhammer, line 40: Why is there a k in the name?
```suggestion
         constexpr double graceShutdownSeconds = 5.0;  // avoid hanging on teardown
```

fengelniederhammer (approved):
I can't really judge whether this the best option or whether this improves the behavior but it looks reasonable :+1:

src/silo/query_engine/query_plan.cpp
```diff
@@ -32,15 +32,42 @@ arrow::Status QueryPlan::executeAndWriteImpl(
...
-         return arrow::Status::ExecutionError(fmt::format(
-            "Request timed out, could not detect any progress within {} seconds.",
-            timeout_in_seconds
-         ));
+         SPDLOG_WARN("Batch wait timed out after {} s — stopping plan.", timeout_in_seconds);
+         return arrow::Status::ExecutionError(
+            fmt::format("Request timed out, no batch within {} seconds.", timeout_in_seconds)
+         );
       }
       ARROW_ASSIGN_OR_RAISE(std::optional<arrow::ExecBatch> optional_batch, future_batch.result());
```
anna-parker, line 72: @Taepper maybe I am missing it but is there a place where you would log the error message if this assignment failed?

src/silo/query_engine/query_plan.cpp
```diff
@@ -56,21 +83,6 @@ arrow::Status QueryPlan::executeAndWriteImpl(
       );
```
anna-parker, line 83: same question for errors thrown here: `ARROW_RETURN_NOT_OK`

src/silo/query_engine/query_plan.cpp
Taepper: As discussed in person: this is actually not an exception but an error return code. This will be logged by the caller on errors

src/silo/query_engine/query_plan.cpp
Taepper: As discussed in person: this is actually not an exception but an error return code. This will be logged by the caller on errors

Issue #986, opened 2025-09-16
Apache Arrow 21.0.0 cannot be built with cppstd 23


Taepper:
We need to wait until this is fixed upstream until removing the extra build step that fixes arrow's cpp std to 20 instead

pflanze:
(In other words: once the arrow project can be compiled as C++23, we don't need to instruct conan to build it as C++20 (unlike SILO's default of C++23) anymore.)

PR #987, opened 2025-09-16
feat: update arrow to version 21.0.0

Taepper:
### Summary

This updates arrow the most recent version. One include needed to be fixed and the cppstd for compiling conan fixed. See #986 for tracking the removal of that fixed version.

https://github.com/apache/arrow/issues/47576 tracks the error upstream.


pflanze:
To summarize, for (later) understanding:

- SILO was, and still is, compiled as C++23. But arrow won't compile as such. Thus pass a setting to conan to specify C++20 for the compilation of arrow. (What I'm surprised about is that conan packages don't set the appropriate C++ version for themselves? Are we overriding this by accident? But given that this here works,  I think it's fine to use.)

- The `#include <arrow/util/async_generator_fwd.h>` is just to update to the new arrow version.

PR #989, opened 2025-09-17
feat: flush outgoing buffers to avoid unbounded buffering

Taepper:
resolves #

## PR Checklist


fengelniederhammer (approved):
I'm not deep enough into Arrow to really judge whether this solves the problem (or whether it's the best solution). But as far as I can tell, this looks reasonable.

PR #993, opened 2025-09-22
Sparse positions

### LINKED ISSUE #965: Always enable "memory-efficient mode"
There used to be SILO versions that store vastly less symbols in the bitmaps (instead of _flipping_ the most numerous bitmap, it was instead deleted and recomputed on-the-fly, when required). This has since been removed from the main branch due to performance degradations.

I think it should be possible to mitigate the performance degradations and permanently enable the memory efficient mode instead

taepper:
resolves #965

### Summary

This optimizes the fundamental data structure SILO uses for storing alignment data. Instead of storing all bitmaps for all positions and symbols, we instead store a search tree (currently std::map, but should be replaced by a B-Tree at some point) of `SequenceDifference`s. 

These sequence differences are not an entire bitmap, but just one bitmap _container_ i.e. able to represent an interval `[0, 2^16)`. Therefore, we with each `SequenceDifference`, we store the position, the symbol, and the offset `x` (this will offset the IDs in the container by `x * 2^16`).

This allows SILO to adapt a lot better to short read-sequencing data, or data that has a lot of uncovered regions.

## PR Checklist

### DIFF
--- a/documentation/developer/sequence_storage.md
+++ b/documentation/developer/sequence_storage.md
@@ -1,0 +1,76 @@
+# Sequence Storage
+
+This document explains the fundamental concepts of how sequences are stored in SILO.
+
+## Overview
+
+SILO uses a hybrid storage approach that separates:
+- **Columnar storage** for mutations (differences from reference sequences)
+- **Row-wise storage** for coverage data (location of the N symbols in nucleotides or X symbols in amino acids)
+
+Both storage types use bitmaps, but we provide separate data structures holding the respective Roaring bitmap containers. This separation enables optimizations for both storage efficiency and query performance.
+
+These special structures are introduced to allow efficient operations on the sequence data with minimal storage requirements.
+In particular, we do not want to store entire bitmaps for all positions, because these can be mostly empty. It is still advantageous to keep bitmap containers for all variation in the sequence as this still allows very fast computations.
+The containers are organized in a tree-map for fast iteration. 
+
+## Roaring Bitmap Containers
+
+A Roaring bitmap container represents a bitmap with a domain of `[0, 2^16)` (0 to 65,535).
+
+The Roaring library provides `roaring::Roaring`, which is an array of Roaring containers that also stores the offset for each container. This allows it to represent the full 32-bit domain.
+
+However, individual containers can be used separately to enable more low-level optimizations and storage-aware implementations of data structures and algorithms.
+
+## Vertical Sequence Index
+
+We conceptually divide the sequence space into tiles with a side length of 2^16 (65,536):
+```
+        0      2^16   2*2^16 3*2^16
+        |      |      |      |
+      0-┌──────┬──────┬──────┬──────┬────>
+        │      │      │      │      │  Position (x-axis)
+        │ Tile │ Tile │ Tile │ Tile │
+        │ 0,0  │ 0,1  │ 0,2  │ 0,3  │
+   2^16-├──────┼──────┼──────┼──────┼
+        │      │      │      │      │
+        │ Tile │ Tile │ Tile │ Tile │
+        │ 1,0  │ 1,1  │ 1,2  │ 1,3  │
+ 2*2^16-├──────┼──────┼──────┼──────┼
+        │      │      │      │      │
+        │ Tile │ Tile │ Tile │ Tile │
+        │ 2,0  │ 2,1  │ 2,2  │ 2,3  │
+        ├──────┼──────┼──────┼──────┼
+        │
+        v
+    Sequence Number (y-axis)
+```
+
+### Tiling Structure
+
+**Horizontal tiling (x-axis)**: Identified by genome position
+- The conceptual tile number is `position_id / 2^16`
+- This is not stored explicitly
+
+**Vertical tiling (y-axis)**: Identified explicitly by `v_index`
+- The bitmap for a single position is split across tiles of size 2^16
+- Each tile's offset is identified by its `v_index`
+- For example, at position 10, we can have multiple bitmaps:
+    - One covering sequence range `[0, 2^16)`
+    - One for range `[2^16, 2*2^16)`
+    - One for range `[2*2^16, 3*2^16)`
+    - And so on...
+
+### Lookup Process
+
+To look up position `x` of sequence `y` and symbol `z`:
+
+1. Identify the vertical tile: `v_index = y / 2^16`
+2. Find the bitmap for position `x` with `v_index` and symbol `z`
+3. Look up index `y % 2^16` within that bitmap container
+
+### Storage Organization
+
+For every genome position, we store the differences (mutations) from the reference sequence in vertical bitmap containers, with one container per symbol (A, C, G, T, etc.).
+
+These bitmap containers are organized in a tree structure for efficient iteration and lookup. They are sorted by the key: `{position, v_index, symbol}`.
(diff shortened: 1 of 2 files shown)


src/silo/query_engine/exec_node/table_scan.cpp
```diff
@@ -8,68 +15,37 @@ namespace {
...
 ) {
+   size_t cardinality = row_ids.cardinality();
+
    std::string general_reference;
    std::ranges::transform(
-      sequence_store.metadata->reference_sequence,
+      sequence_column_partition.metadata->reference_sequence,
       std::back_inserter(general_reference),
       SymbolType::symbolToChar
    );
```
fengelniederhammer: Why don't we use this instead here?

```suggestion
   std::string general_reference = sequence_column_partition.reference_sequence_string;
```

src/silo/query_engine/filter/operators/is_in_covered_region.cpp
```diff
@@ -0,0 +1,78 @@
...
+   EVOBENCH_SCOPE("IsInCoveredRegion", "evaluate");
+   CopyOnWriteBitmap result_bitmap;
+   auto bitmap_iter = covered_region_bitmaps->begin();
+   for (size_t row_idx = 0; row_idx < row_count; ++row_idx) {
+      const auto& [start, end] = covered_region_ranges->at(row_idx);
+      std::optional<const roaring::Roaring*> row_bitmap = std::nullopt;
+      if (bitmap_iter != covered_region_bitmaps->end() && bitmap_iter->first == row_idx) {
+         row_bitmap = &bitmap_iter->second;
+         bitmap_iter++;
+      }
```
fengelniederhammer: We can't we do a lookup here? Something like `covered_region_bitmaps[row_idx]`?

src/silo/query_engine/filter/operators/is_in_covered_region.test.cpp
fengelniederhammer: I think this could use a test for when asking for a position outside of `start_ends`.

src/silo/roaring_util/subset_ranks.h
```diff
@@ -0,0 +1,17 @@
...
+
+std::vector<uint64_t> roaringSubsetRanks(
+   const roaring::internal::container_t* c_a,
+   uint8_t type_a,
+   const roaring::internal::container_t* c_b,
+   uint8_t type_b,
+   uint32_t base
+);
+
+}
```
fengelniederhammer:
```suggestion
}  // namespace silo::roaring_util
```

src/silo/roaring_util/subset_ranks.test.cpp
```diff
@@ -0,0 +1,99 @@
+#include "silo/roaring_util/subset_ranks.h"
+
+#include <gtest/gtest.h>
+#include <roaring/roaring.hh>
+
+using silo::roaring_util::roaringSubsetRanks;
+
+class RoaringSubsetRanksTest : public ::testing::Test {
```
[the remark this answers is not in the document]
Taepper: As these are tests I think it is fine, but happy to comply

src/silo/storage/column/sequence_column.h
```diff
@@ -76,35 +77,46 @@ class SequenceColumnPartition {
...
-   SequenceColumnMetadata<SymbolType>* metadata;
-   SequenceColumnInfo sequence_column_info;
-   std::vector<std::pair<size_t, typename SymbolType::Symbol>>
-      indexing_differences_to_reference_sequence;
-   std::vector<SequencePosition<SymbolType>> positions;
-   std::vector<roaring::Roaring> missing_symbol_bitmaps;
+   const SequenceColumnMetadata<SymbolType>* metadata;
+   const size_t genome_length;
+
+   std::map<size_t, typename SymbolType::Symbol> indexing_differences_to_reference_sequence;
```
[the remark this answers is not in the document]
Taepper: This will only come in the follow-up PR

src/silo/storage/column/vertical_sequence_index.cpp
```diff
@@ -0,0 +1,261 @@
...
+            sequence_diff.typecode,
+            &result_typecode
+         );
+         if (result_container != current_container) {
+            roaring::internal::container_free(current_container, current_typecode);
+            current_container = result_container;
+            current_typecode = result_typecode;
+         }
+      }
+   }
```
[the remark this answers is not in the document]
fengelniederhammer: Understood! I got confused there. It makes sense that we can't simply iterate over the v_index because we don't know the upper bound.

src/silo/storage/column/vertical_sequence_index.cpp
```diff
@@ -0,0 +1,261 @@
...
+   SymbolType::Symbol symbol
+) const {
+   // We compute the range of bitmap containers that are relevant for our position
+   auto start = vertical_bitmaps.lower_bound(
+      SequenceDiffKey{position_idx, 0, static_cast<SymbolType::Symbol>(0)}
+   );
+   auto end = vertical_bitmaps.lower_bound(
+      SequenceDiffKey{position_idx + 1, 0, static_cast<SymbolType::Symbol>(0)}
+   );
+   // We need to union all bitmap containers at this position and return the inverse
```
fengelniederhammer: I probably overlooked it - where does this take the inverse?

fengelniederhammer (commented):
Some final question. Otherwise I think this looks good (but I have to admit that it's quite hard to grasp every detail of this PR, maybe we should run some tests whether it still returns the same results as a previous version on a larger data set?)

documentation/developer/sequence_storage.md
```diff
@@ -0,0 +1,76 @@
...
+
+SILO uses a hybrid storage approach that separates:
+- **Columnar storage** for mutations (differences from reference sequences)
+- **Row-wise storage** for coverage data
+
+Both storage types use bitmaps, but we provide separate data structures holding the respective Roaring bitmap containers. This separation enables optimizations for both storage efficiency and query performance.
+
+These special structures are introduced to allow efficient operations on the sequence data with minimal storage requirements.
+In particular, we do not want to store entire bitmaps for all positions, because these can be mostly empty. It is still advantageous to keep bitmap containers for all variation in the sequence as this still allows very fast computations.
+The containers are organized in a tree-map for fast iteration.
```
fengelniederhammer: Just to make sure: This is not yet true but will be soon?

documentation/developer/sequence_storage.md
fengelniederhammer: 👍

documentation/developer/sequence_storage.md
Taepper: This is true already :)

documentation/developer/sequence_storage.md
```diff
@@ -0,0 +1,76 @@
+# Sequence Storage
+
+This document explains the fundamental concepts of how sequences are stored in SILO.
+
+## Overview
+
+SILO uses a hybrid storage approach that separates:
+- **Columnar storage** for mutations (differences from reference sequences)
+- **Row-wise storage** for coverage data
```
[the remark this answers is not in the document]
Taepper: Where the "N"s are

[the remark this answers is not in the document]
fengelniederhammer: Should we maybe also explain that here?

PR #998, opened 2025-09-26
feat(silo)!: support multiple lineage systems

### LINKED ISSUE #691: Support multiple lineage systems
Currently SILO supports only one lineage system per instance. This system is defined by the field `lineageDefinitionsFilename` in the preprocessingConfig.

This config field should be changed from a string to a map, where the keys will be used as keys of lineage systems which will be referenced in the databaseConfig. Therefore a column in the databaseConfig can reference this system's name in the key `generateLineageIndex`, which currently takes a boolean value.

Taepper:
BREAKING CHANGE: Multiple breaking changes to lineage configuration:

- PreprocessingConfig: `lineageDefinitionsFilename` renamed to `lineageDefinitionFilenames` and changed from string to array
- DatabaseConfig: `generateLineageIndex` changed from boolean to string (expects lineage definition filename)

resolves #691 

### Summary

This adds the possibility to define multiple lineage definition systems. 

In preprocessing, the lineage systems are given as a list of filenames in YAML. The filenames are expected as relative from inputDirectory.

The string columns with lineage indexes are still marked with `generateLineageIndex` in the database config. Instead of `true`, the value should be the name of the lineage definition file (`.yaml` may be omitted). 

## PR Checklist


src/silo/config/initialize_config.cpp
```diff
@@ -128,8 +141,8 @@ void InitializeConfig::overwriteFrom(const VerifiedConfigAttributes& config_sour
    if (auto var = config_source.getPath(inputDirectoryOptionKey())) {
       initialization_files.directory = var.value();
    }
-   if (auto var = config_source.getPath(lineageDefinitionsFilenameOptionKey())) {
-      initialization_files.lineage_definitions_file = var.value();
+   if (auto var = config_source.getString(lineageDefinitionFilenamesOptionKey())) {
+      initialization_files.lineage_definition_files = var.value();
```
fengelniederhammer, line 145: Do we maybe need config values that allow arrays as their value? From CLI those could be repeated arguments. In env variables we can still fall back to comma separated strings and parse them here. Anyway I think it would be good to do the parsing as early as possible (i.e. here and not only when reading `initialization_files.lineage_definition_files`).

src/silo/initialize/initializer.cpp
```diff
@@ -214,7 +231,7 @@ static const std::string UNALIGNED_NUCLEOTIDE_SEQUENCE_PREFIX = "unaligned_";
 silo::schema::DatabaseSchema Initializer::createSchemaFromConfigFiles(
    config::DatabaseConfig database_config,
    ReferenceGenomes reference_genomes,
-   common::LineageTreeAndIdMap lineage_tree,
+   std::map<std::filesystem::path, common::LineageTreeAndIdMap> lineage_tree,
```
fengelniederhammer, line 234:
```suggestion
   std::map<std::filesystem::path, common::LineageTreeAndIdMap> lineage_trees,
```
:)

src/config/source/yaml_file.cpp
```diff
@@ -17,6 +17,20 @@ using silo::config::ConfigKeyPath;
 
 namespace {
 
+bool isSequenceOfScalars(const YAML::Node& node) {
+   if (!node.IsSequence()) {
+      SPDLOG_TRACE("isSequenceOfScalars = false, node is not a sequence");
+      return false;
+   }
+   for (auto element : node) {
```
fengelniederhammer, line 25: Nitpick: My IDE says `Clang-Tidy: Replace loop by 'std::ranges::all_of()'`

Taepper:
> Would it make sense to change lineageDefinitionFilenames to an object? The values would be the lineage file and generateLineageIndex could then reference the keys there (instead of the filename with a autocomplete attempt that may be confusing and error prone).

Unfortunately, this is not possible, as the preprocessing-config can also be supplied by CLI arguments / env vars, hence, it can only be a string.


Mid-term I do not think this matters too much, as the current `preprocessing` should be replaced by a more idiomatic and dynamic API for schema management

fengelniederhammer:
> > Would it make sense to change lineageDefinitionFilenames to an object? The values would be the lineage file and generateLineageIndex could then reference the keys there (instead of the filename with a autocomplete attempt that may be confusing and error prone).
> 
> Unfortunately, this is not possible, as the preprocessing-config can also be supplied by CLI arguments / env vars, hence, it can only be a string.
> 
> Mid-term I do not think this matters too much, as the current `preprocessing` should be replaced by a more idiomatic and dynamic API for schema management

Ok that is unfortunate. I could imagine that we could still get keys from CLI arguments by implementing a syntax like `--option 'key:value'` but I'm not sure whether that's worth the trouble.

But I'd still like to get rid of the heuristics that try to autocomplete the filename. Do we need that at all? The DB config is written once and it doesn't really matter how long the config values are. We could simply require maintainers to put the whole filename there (maybe without the `.yaml` ending). That would be a lot more transparent, easier to explain and less error prone. IMO wouldn't lose anything.

PR #1008, opened 2025-10-02
feat(silo): address memory leak concerns

anna-parker:
resolves #

Address Claude AI concerns from https://loculus.slack.com/archives/C05G172HL6L/p1759339414089419?thread_ts=1759339296.128359&cid=C05G172HL6L marked as critical with the exception of 2 which seems to be a false positive as `poco_parameter` is owned by `Poco::Net::HTTPServer` once it is passed into the constructor. When server goes out of scope at the end of `runApi()` (after server.stop()), the HTTPServer destructor should delete the params object.

## PR Checklist

### DIFF
--- a/src/silo/storage/buffer/page.h
+++ b/src/silo/storage/buffer/page.h
@@ -13,6 +13,7 @@ class Page {
   public:
    uint8_t* buffer;
 
+   // On a failed allocation std::bad_alloc is thrown
    Page() { buffer = new uint8_t[PAGE_SIZE]; }
 
    Page(Page&& other) noexcept


src/silo/storage/buffer/page.h
```diff
@@ -13,7 +13,7 @@ class Page {
   public:
    uint8_t* buffer;
 
-   Page() { buffer = reinterpret_cast<uint8_t*>(malloc(PAGE_SIZE)); }
+   Page() { buffer = new uint8_t[PAGE_SIZE]; }
```
anna-parker, line 16:
```suggestion
   // On a failed allocation std::bad_alloc is thrown
   Page() { buffer = new uint8_t[PAGE_SIZE]; }
```

PR #1019, opened 2025-10-09
feat(silo): add template to run ppx instance

anna-parker:
resolves #

## PR Checklist


Taepper (approved):
Some comments, but otherwise works like a charm!

anna-parker:
Hey @Taepper I just wanted to use this again to redo some tests and I am agin getting the error:
```
siloPreprocessing-1  | [2025-10-23 15:32:21.652] [logger] [error] [main.cpp:177] Internal Error: Failed to read database config: Could not open file ./database_config.yaml
```
I don't understand how this is possible as it looks like the file is bound correctly? 
image

Its also really weird I also get this error when I check out old commits where I'm convinced it worked before... do you have any clue what is going on?

PR #1031, opened 2025-10-13
feat(benchmarking): support versioned datasets

### LINKED ISSUE #1018: Benchmarking: allow benchmarking across SILO versions when datasets must adapt
#1016 solves the immediate problem, but as discussed in the SILO meeting yesterday, re-running benchmarks for older SILO versions (when config details have changed) now breaks. Also there will be breaks when the config file is changed again in the future while there are both branches with the old and the new format. The breaking is also not limited to the config files, also the input formats can change and in fact has so in the last few months. Also I'd like to provide the possibility to calculate graphs of SILO performance numbers across versions, and this will need re-running older versions whenever we do changes to the benchmarking config like e.g. number of threads or similiar, to make runs comparable, and even just adding a new variable for new features would make it complicated to correlate older runs without that variable with newer runs, re-running benchmarks would be easier (although at some point may grow to be costly).

The conundrum is that the benchmarks/ directory in SILO, while it is part of SILO and thus can be maintained in sync with SILO changes, tries to be independent of datasets (we currently have two datasets, there will quite likely be more in the future), while the datasets try to be independent of SILO version, thus we have 2 dimensions of changes (a matrix).

The solution discussed yesterday was to version the dataset: the silo-benchmark-datasets/ folder currently has subfolders SC2open/ and west_nile/, which contain the files. The idea is to instead have e.g. subfolders `v0.1.0`, `v0.7.0`, `v0.9.0`, `b4bb617861473af1`, with files for the respective SILO version and forward until a folder for a newer SILO version is found. evobench already has code (and some changes unreleased from yesterday) to handle Git graphs, so it can resolve each subfolder name to a Git commit and reliably detect via the ancestor relationship which is newer--or more to the point, which is the latest ancestor (or itself) of the commit that is to be benchmarked. Symlinks can be used to avoid duplicating files.

An alternative solution, not discussed yesterday, would be to add code to the benchmarking/ folder in SILO that *derives* the dataset files from the single dataset folder. I.e. instead of manually editing the files or manually running the conversions on them (as I did for the input format change a couple months ago), the benchmarking code in SILO does that programmatically, and we would continue to have a single dataset folder, compatible with say SILO `v0.1.0` (that folder could still receive additional files over time, that older SILO versions would simply ignore; i.e. it would still be possible to store versioned config files in there if that is easier than converting them, e.g. `database_config.yaml` then `database_config-0.9.0.yaml` and then have code in benchmarking/ that is adapted in 0.9.0 to copy/symlink the latter path). The disadvantage for this solution is that one can forget to update SILO's benchmarking/ folder to do that (unless we make it part of CI), thus v0.10.0 might be released and not work for benchmarking; while changing the datasets as above is outside SILO's Git history.

pflanze:
Closes #1018

### Summary

Use the new `DATASET_DIR` env variable from evobench, to make use of the dataset versioning.

## PR Checklist

### DIFF
--- a/benchmarking/Makefile
+++ b/benchmarking/Makefile
@@ -13,10 +13,25 @@ endif
 # Our custom parameters (the shared ones, there can be more checks in
 # the `Makefile_*` files):
 
-# The name of a subfolder of `~/silo-benchmark-datasets/` where the
-# dataset is stored
+# The name of a subfolder of the setting in
+# `versioned_datasets_base_dir`, where the versioned datasets are
+# stored; in spite of `DATASET_DIR` below having the full path, we
+# still also use this dataset name, thus require it to be set
+# (guaranteed when running via evobench-run, but not when running
+# manually).
 ifeq ($(DATASET),)
 $(error "DATASET must be a subfolder name, got '$(DATASET)'")
+endif
+
+# Set by `evobench-run run`--chosen according to the optional
+# `versioned_datasets_base_dir` setting in `evobench-run.ron`, the
+# value of the `DATASET` custom variable, the current commit id, and
+# the names of the subdirectories of `DATASET_DIR` representing git
+# revision names. See documentation in the `evobench-run.ron` file,
+# and [Offer dataset versioning
+# #32](https://github.com/GenSpectrum/evobench/issues/32).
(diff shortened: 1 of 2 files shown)


Taepper (approved):
lgtm

PR #1062, opened 2025-11-25
refactor(silo): linter fixes and minor edits

taepper:
This PR only contains some minor linter fixes and general code quality improvements, to keep the change list of subsequent PRs more concise

### DIFF
--- a/src/silo/query_engine/filter/expressions/expression.h
+++ b/src/silo/query_engine/filter/expressions/expression.h
@@ -5,8 +5,8 @@
 
 #include <nlohmann/json_fwd.hpp>
 
-#include "silo/database.h"
 #include "silo/query_engine/filter/operators/operator.h"
+#include "silo/storage/table.h"
 #include "silo/storage/table_partition.h"
 
 namespace silo::query_engine::filter::expressions {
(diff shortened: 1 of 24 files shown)


src/silo/query_engine/filter/expressions/expression.h
```diff
@@ -22,9 +22,9 @@ class Expression {
    /// codes count as matches), LOWER_BOUND returns the lower bound of sequences matching this
    /// expression (i.e. ambiguous codes in negations count as matches)
    /// NONE does not specially consider ambiguous symbols
-   enum AmbiguityMode { UPPER_BOUND, LOWER_BOUND, NONE };
+   enum AmbiguityMode : uint8_t { UPPER_BOUND, LOWER_BOUND, NONE };
```
fengelniederhammer: This syntax is explicitly saying to use 8 bit values under the hood for the enum entries (instead of the default which is probably int32)?

fengelniederhammer (approved):

src/silo/query_engine/filter/expressions/expression.h
Taepper: Exactly!

PR #1063, opened 2025-11-25
`SymbolInSet` and add a separate rewrite step in query optimization

taepper:
This splits up the `Expression::compile` function into a `Expression::rewrite` and `Expression::compile`. This removes some duplicated logic within rewrite functions, removes anti-patterns in compile functions (`return OtherExpression{...}.compile(...);` is very hard to reason about), and allows better extensability.

Additionally a new expression `SymbolInSet` is introduced which is being mapped to by other expressions in the rewrite step. It behaves similar to `SymbolEquals`, but without ambiguity semantics and for a set of symbols. This allows `SymbolEquals` to resolve the ambiguity semantics in the rewrite phase already and the implementation of matching a set is now separate. And `HasMutation` is no longer almost a duplicate of the `SymbolEquals` implementation, but also maps to `SymbolEquals`.

Lastly, this change can lead to large performance improvements with the new sequence storage architecture, especially when combined with more improvements to the optimizer phase. As we can no longer address bitmaps of symbols@positions directly, we need to iterate over all bitmap containers of a given position already. Now we can check for all symbols in a single pass instead. This is relevant when matching for the reference symbol, as it previously required a separate scan through these positions for all other symbols, which has now been reduced to one scan again.

The optimizer improvement included in this PR allows many `SymbolEquals` filters under an `Or` expression to be combined into a single `SymbolInSet` expression, because `123A | 123C | 123G | 123T | 123-` is a common query pattern

### DIFF
--- a/src/silo/roaring_util/bitmap_builder.cpp
+++ b/src/silo/roaring_util/bitmap_builder.cpp
@@ -7,21 +7,31 @@ void BitmapBuilderByContainer::addContainer(
    roaring::internal::container_t* container,
    uint8_t typecode
 ) {
-   SILO_ASSERT(current_v_tile_index <= v_index);
-   if (current_v_tile_index != v_index) {
-      if (current_container != nullptr) {
-         roaring::internal::ra_append(
-            &result_bitmap.roaring.high_low_container,
-            current_v_tile_index,
-            current_container,
-            current_typecode
-         );
-      }
+   if (current_v_tile_index > v_index) {
+      throw std::runtime_error(
+         "containers must be inserted into BitmapBuilderByContainer by increasing v_index"
+      );
+   }
+   // first container received
+   if (current_container == nullptr) {
       current_v_tile_index = v_index;
       current_typecode = typecode;
       current_container = roaring::internal::container_clone(container, typecode);
-   } else { /* current_v_tile_index == sequence_diff_key.v_index */
-      SILO_ASSERT(current_container != nullptr);
+   }
+   // current_container is finished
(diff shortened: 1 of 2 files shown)


src/silo/query_engine/filter/expressions/expression.h
```diff
@@ -26,11 +26,17 @@ class Expression {
...
 
-   [[nodiscard]] virtual std::unique_ptr<silo::query_engine::filter::operators::Operator> compile(
+   [[nodiscard]] virtual std::unique_ptr<silo::query_engine::filter::expressions::Expression>
+   rewrite(
       const storage::Table& table,
       const storage::TablePartition& table_partition,
       AmbiguityMode mode
    ) const = 0;
+
+   [[nodiscard]] virtual std::unique_ptr<silo::query_engine::filter::operators::Operator> compile(
```
fengelniederhammer:
```suggestion
   [[nodiscard]] virtual std::unique_ptr<operators::Operator> compile(
```

src/silo/query_engine/filter/expressions/symbol_in_set.cpp
```diff
@@ -0,0 +1,260 @@
+#include "silo/query_engine/filter/expressions/symbol_in_set.h"
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include <fmt/ranges.h>
+#include <spdlog/spdlog.h>
+#include <nlohmann/json.hpp>
```
fengelniederhammer: We probably also don't need this import, do we?

PR #1066, opened 2025-11-26
Fix superfluous empty bitmaps

taepper:
With the sparse-positions refactor, we only store horizontal bitmaps when coverage information is actually required (i.e. there are Ns other than at the start and at the end).

This was actually implemented wrong and we added bitmaps when ANY Ns were found. Now we only add them when Ns are found that are not at the beginning or the end


pflanze:
Trying to understand, in my words:

The local variable `horizontal_bitmap` will now only be added if it turns out to be non-empty; this saves a slot in the `horizontal_bitmaps` map. The new logic is necessary because previously there could be removals after the emptiness check.

PR #1070, opened 2025-11-28
feat(silo): add efficient way of computing vertical coverage bitmaps and use them to adapt reference positions

taepper:
Second-to-last speed-up PR for the sparse-positions refactor.

This allows us to readapt the local sequence positions to using a different reference symbol. Can improve storage efficiency and also adds an efficient method of recomputing vertical coverage bitmaps, which will be used by the last speed-up PR


src/silo/storage/column/vertical_sequence_index.cpp
```diff
@@ -59,6 +60,79 @@ std::pair<const_iterator<SymbolType>, const_iterator<SymbolType>> VerticalSequen
...
+   uint32_t best_count = symbol_counts[global_reference_symbol];
+   for (const auto& symbol : SymbolType::SYMBOLS) {
+      if (symbol == global_reference_symbol) {
+         continue;
+      }
+      if (symbol_counts[symbol] > best_count) {
+         best_symbol = symbol;
+         best_count = symbol_counts[symbol];
+      }
+   }
```
fengelniederhammer: Nitpick: We could extract this into its own little method.

src/silo/storage/column/horizontal_coverage_index.test.cpp
```diff
@@ -0,0 +1,453 @@
...
+}
+
+TEST_F(HorizontalCoverageIndexTest, ConstructsWithZeroLength) {
+   EXPECT_NO_THROW({ HorizontalCoverageIndex index; });
+}
+
+// Null Sequence Tests
+TEST_F(HorizontalCoverageIndexTest, InsertSingleNullSequence) {
+   EXPECT_NO_THROW(index->insertNullSequence());
+}
```
fengelniederhammer: Is this test really useful? Can we extend the test to produce a meaningful result?

src/silo/storage/column/horizontal_coverage_index.cpp
```diff
@@ -23,13 +23,25 @@ void HorizontalCoverageIndex::insertCoverage(
...
+   uint16_t sequence_idx_lower_bits = sequence_idx & 0xFFFF;
+
+   if (sequence_idx_lower_bits == 0) {
+      batch_start_ends.emplace_back(start, end);
+   } else {
+      auto& [batch_start, batch_end] = batch_start_ends.back();
+      batch_start = std::min(batch_start, start);
+      batch_end = std::max(batch_end, end);
+   }
+   SILO_ASSERT_EQ(batch_start_ends.size(), sequence_idx_upper_bits + 1);
```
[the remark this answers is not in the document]
Taepper: Let's prefer safety. I do not think that this loop is very "hot"

PR #1072, opened 2025-12-01
Resolve memory leak and adapt CI to catch it in the future

taepper:
This resolves a memory leak introduced in #993. I adapted the CI to also run the tests in debug build (which is also using ASAN), so this would be caught in the future

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -267,3 +267,40 @@ jobs:
         run: |
           docker build -f ./Dockerfile_linter \
             --build-arg DEPENDENCY_IMAGE=ghcr.io/genspectrum/lapis-silo-dependencies:commit-${{ env.HEAD_SHA }}-amd64 . 
+    
+  unitTests:
+    runs-on: ubuntu-latest
+    
+    steps:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -64,4 +64,4 @@ full-clean: clean
 	rm -rf build
 
 .PHONY:
-	full-clean clean clean-api e2e format all test all-tests format ci
```
fengelniederhammer: Is this change intentional?

.github/workflows/ci.yml
```diff
@@ -267,3 +267,40 @@ jobs:
...
+          
+          echo "DOCKER_IMAGE=$DOCKER_IMAGE" >> $GITHUB_ENV
+      
+      - name: Run CI in Docker
+        run: |
+          docker run --rm \
+            -v ${{ github.workspace }}:/workspace \
+            -w /workspace \
+            ${{ env.DOCKER_IMAGE }} \
+            bash -login -c "mv /src/conanprofile . && make test"
```
fengelniederhammer: There is already a "Run unit tests" step somewhere above. Do we need both of them?

PR #1076, opened 2025-12-02
feat(build): add Makefile rule to generate conanprofile

taepper:
Developer QoL change, on new checkout `make test` is now working if conan, cmake, .. is installed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ DEPENDENCIES_FLAG=dependencies
 
 ci: format all-tests
 
+conanprofile:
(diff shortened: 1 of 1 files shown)


pflanze (approved):
Looks good, this still depends on conan to be present already, which is fine, but it reminds me that in the later commit you only changed the $(DEPENDENCIES_FLAG) target, not the conanprofile one!

Taepper:
I think the later commit could also move this into the make install script, still an improvement by itself I would say

PR #1077, opened 2025-12-02
Automatic conan setup

taepper:
WIP PR to automatically install dependencies on new `git clone`, without needing to install `conan` which is not a standard build tool on our machines


pflanze (commented):
Other than the "source" thing above:

I don't remember how we found that "uv" works: does it still use a current venv if one is set? If not, *does* it use a user-global directory (~/.uv something)? I would:

1. ensure that the installed conan (and other uv things) remain shared, to avoid reinstalling those for every benchmarking run.

2. document that this is the case.

Also, does that mean that benchmarking/ should be updated to use these scripts, I guess?

PR #1045, opened 2025-10-30
fix(silo): reduce peak memory load by changing default allocator and tuning mimalloc options

taepper:
reduces impact of #1034 

### Summary

We were already using `mimalloc` indirectly with Arrow (e.g. through Arrow `RecordBatch` creating). Now we are using `mimalloc` for all allocations in SILO.

I tested this by starting an instance with loculus data and stress testing the sequence download:

Before:
```
alexander@dev-1:~/LAPIS-SILO$ monitor_rss
Monitoring RSS of process 177715...
Timestamp, PID, RSS (KB)
2025-10-30 11:07:48, 177715, 61696
2025-10-30 11:07:49, 177715, 61696
2025-10-30 11:07:50, 177715, 437364
2025-10-30 11:07:51, 177715, 517664
2025-10-30 11:07:52, 177715, 467140
2025-10-30 11:07:53, 177715, 513908
2025-10-30 11:07:54, 177715, 468708
2025-10-30 11:07:55, 177715, 468708
2025-10-30 11:07:56, 177715, 468708
2025-10-30 11:07:57, 177715, 465888
2025-10-30 11:07:58, 177715, 440084
2025-10-30 11:07:59, 177715, 425920
2025-10-30 11:08:00, 177715, 460664
2025-10-30 11:08:01, 177715, 463024
2025-10-30 11:08:02, 177715, 483336
2025-10-30 11:08:03, 177715, 433832
2025-10-30 11:08:05, 177715, 491012
2025-10-30 11:08:06, 177715, 471748
```

After:

```
alexander@dev-1:~/LAPIS-SILO$ monitor_rss
Monitoring RSS of process 236049...
Timestamp, PID, RSS (KB)
2025-10-30 11:15:55, 236049, 61824
2025-10-30 11:15:56, 236049, 61824
2025-10-30 11:15:57, 236049, 254328
2025-10-30 11:15:58, 236049, 243360
2025-10-30 11:15:59, 236049, 315176
2025-10-30 11:16:00, 236049, 259712
2025-10-30 11:16:01, 236049, 269188
2025-10-30 11:16:02, 236049, 259024
2025-10-30 11:16:03, 236049, 223536
```

I let the stress tests running for >10 minutes and the memory did not increase above the peak anymore

PR #1080, opened 2025-12-02
feat(silo): use mimalloc as the default allocator

taepper:
I decided to split up #1045 for the allocator change and the mimalloc option tweaking


conanfile.py
```diff
@@ -22,6 +23,7 @@ class SiloRecipe(ConanFile):
     default_options = {
         "abseil/*:shared": False,
 
+        "arrow/*:with_mimalloc": False,
```
pflanze: Is this because mimalloc is used anyway globally?

Issue #798, opened 2025-06-10
Add multi-table support to SILO


taepper:
SILO only supports containing a single table. Since the refactoring in #720, it is now quite straight-forward to add support for multiple tables.

Actions should then reference which table they apply to. Later, mulit-table operators such as joins should be considered.

This issue will also be used as an epic for other changes to e.g. the schema, which should be done with this change.

PR #1092, opened 2025-12-15
Multi-table support

taepper:
first step for #798

### Summary

This introduces the possibility to create additional tables in SILO instances and query each of them. All these API function which are going to be e.g. exposed via python bindings are in the `Database` class (which will be the object kept around in python -> calling `database.createTable` in python will be largely be a direct translation)

I already fixed the ugly error handling in `lineage_definition_handler.cpp` in #1093

## PR Checklist

~~- [ ] All necessary documentation has been adapted or there is an issue to do so.~~ -> No public facing api yet


fengelniederhammer (approved):
Looks reasonable to me 👍

Issue #1049, opened 2025-11-04
EPIC: Add Python bindings


Taepper:
SILO should be able to be called directly from python without the necessity of communicating via http. This can make SILO usable directly from python projects as the storage backend for sequence alignments (i.e. fasta/bam/vcf alternative).

Therefore, we should directly distribute python bindings with SILO

PR #1105, opened 2026-01-13
fix(silo): only link mimalloc into the final executable, not the library, which might be included differently

taepper:
Move `mimalloc` linking from `silolib` shared library to the `silo` executable.

When `libsilolib.so` included mimalloc's `malloc` override symbols, it caused crashes in processes that had already allocated memory with a different allocator—such as Python with pyroaring loaded. Linking `mimalloc-static` to the executable only ensures the standalone `silo` binary still uses mimalloc, while `libsilolib.so` remains compatible with Python bindings and other host applications.

Therefore, we no longer invoke special mimalloc functions (see `allocator.h`), but instead fully rely on the malloc override

This also fixes incompatibility problems on macos, by disabling mimalloc on platforms other than linux

The changes to a shared library do not functionally change the setup, but are anyways required for generating python bindings in #1049

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,10 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 option(USE_MIMALLOC "Use mimalloc allocator" ON)
 
 # Only use mimalloc in Release mode, because of conflicts with ASAN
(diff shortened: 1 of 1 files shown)


Taepper:
Merging as discussed with @pflanze and then working on #1106 separately

Issue #1124, opened 2026-01-27
Handle NULLs in IndexedStringColumns correctly


taepper:
Currently, an empty string and null are interpreted the same in the case of `IndexedStringColumn`. This should be fixed!

PR #1125, opened 2026-01-27
547 null sequences

### LINKED ISSUE #547: Missing aligned sequences: make Fasta action return `null` instead of "all N" sequence
Currently, when there is an aligned sequence missing, we fill it up with a sequence that consists of `N` only (or `X` in the case of amino acid sequences).

Change it such that the Fasta action returns `null` instead for those sequences.

Open points to investigate:
* Can we simply adapt SILO to store those sequences as "absent" instead of filling them with `N`?
* Does it have effects on other parts, such as mutation computation?
  * Actually it should not matter, because `N` are omitted from mutation comparison computations anyway.

taepper:
resolves #547 and #1124

### Summary

Now, ALL columns contain a `null_bitmap` this simplifies logic. `TableScan`s already checked for `column.isNull(row_id)`. This makes the implementation of `Column::isNull` clearer for each column, and enables a generic implementation of the `IsNull`-expression

These changes are together in one branch to avoid many subsequent breaking changes and `CURRENT_SILO_SERIALIZATION_VERSION` bumps.

## PR Checklist

### DIFF
--- a/linter_changes.bash
+++ b/linter_changes.bash
@@ -30,6 +30,7 @@ docker run --rm \
   -e files="$files" \
   "$DOCKER_IMAGE" \
   /bin/bash -c '
+    set -euo pipefail
     cmake -G Ninja -D BUILD_WITH_CLANG_TIDY=ON -D CMAKE_BUILD_TYPE=Debug -B build/Debug
     targets=()
     for file in $files; do


src/silo/query_engine/filter/expressions/is_null.test.cpp
fengelniederhammer: Let's add a test for sequence columns?

documentation/input_format.md
```diff
@@ -0,0 +1,304 @@
...
+}
+```
+
+**Fields**:
+- `sequence` (required): The nucleotide sequence string
+- `insertions` (optional): Array of insertions in format `position:sequence`
+- `offset` (optional): Integer offset into reference genome (default: 0)
+
+**Valid Nucleotide Characters**:
+| Character | Meaning |
```
fengelniederhammer:
```suggestion
**Valid Nucleotide Characters**:

| Character | Meaning |
```
My IDE doesn't properly display the table without it.

documentation/input_format.md
```diff
@@ -0,0 +1,304 @@
...
+  "E": {
+    "sequence": "MYSF*",
+    "insertions": ["214:EPE"],
+    "offset": 0
+  }
+}
+```
+
+**Valid Amino Acid Characters**:
+| Character | Meaning |
```
fengelniederhammer:
```suggestion
**Valid Amino Acid Characters**:

| Character | Meaning |
```

documentation/input_format.md
```diff
@@ -0,0 +1,304 @@
...
+| `A-Z` | Standard amino acids (21 total) |
+| `*` | Stop codon |
+| `X` | Unknown/ambiguous |
+| `-` | Gap (deletion) |
+
+## Configuration Files
+
+### database_config.yaml
+
+Defines the schema for your SILO database:
```
fengelniederhammer: Should we maybe put that to the top? I think it's easier to explain the input data once you explained this.

PR #1129, opened 2026-01-30
feat: add compiled binaries for linux to GitHub releases

### LINKED ISSUE #575: Distributing self-contained binaries
Related to https://github.com/loculus-project/loculus/issues/2765

In Loculus we'd like to have zstd available in a container that does silo preprocessing.

There's a PR #574 but in general there's a question how to package/distribute silo.

It might be good to add the following release assets in addition to the current Alpine image:

- Release binaries, for Linux (glibc) x86 and ARM [possibly also for macOS (would simplify local testing as one doesn't have to mount things into docker)]

fengelniederhammer:
resolves #575

### Summary

I forked the repo to have a playground on which I can test creating releases: https://github.com/fengelniederhammer/LAPIS-SILO. If you want to test this, I can add you as a collaborator there.

It would actually be great if someone could test whether the arm binary works.

Since we already use Debian's `oldstable` as base images to build the binary for the containers, we can simply extract it from there and add it to the already existing GitHub release that Release Please created earlier.

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,7 +192,7 @@ jobs:
             PLATFORM="linux/$arch"
             echo "Extracting from $IMAGE ($PLATFORM)"
 
-            binary="./silo-linux-$arch"
+            binary="silo-linux-$arch"
 
             docker create --platform $PLATFORM --name extract-$arch $IMAGE
             docker cp extract-$arch:/app/silo "./$binary"


Taepper:
I think it would be good to test this with more than just the minimal setup (actually linking in our libraries), so we are certain that it works on more distros. But I also think we can just test this after we have merged this given that some broken binaries will not be too bad.

fengelniederhammer:
> I think it would be good to test this with more than just the minimal setup (actually linking in our libraries), so we are certain that it works on more distros. But I also think we can just test this after we have merged this given that some broken binaries will not be too bad.

I just did the 

            docker create --platform $PLATFORM --name extract-$arch $IMAGE
            docker cp extract-$arch:/app/silo "./$binary"

part locally and was able to successfully execute the binary locally that I extracted from the current `latest` image. I think that should be enough to show that it works (at least conceptually - it doesn't prove much about different distros since I tried it on my Ubuntu). The roundtrip through the GitHub release shouldn't break anything in that regard.

Different distros are a different topic that this PR doesn't try to solve anyway.

PR #1142, opened 2026-02-06
feat(silo): add a arrow query interface to the python bindings

### LINKED ISSUE #1141: Add arrow query interface to python bindings
It should be possible to write general SILO queries from python and receive the results directly through pyarrow (not ndjson)

taepper:
resolves #1141 

### Summary

This adds SILO queries to the python interface. Crucially, this directly makes them available to interact with in python through pyarrow, instead of going through ndjson as we do in the http interface

## PR Checklist

### DIFF
--- a/python/tests/test_database.py
+++ b/python/tests/test_database.py
@@ -4,6 +4,7 @@ import json
 import tempfile
 import shutil
 import pyroaring
+import pyarrow as pa
 
 
 # Path to test data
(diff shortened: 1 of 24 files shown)


python/tests/test_database.py
```diff
@@ -549,3 +633,274 @@ def test_gene_table_with_extra_columns(self):
...
+            sequence_name="seq",
+            reference_sequence="ACGT"
+        )
+
+        result = empty_database.get_tables()
+        data = result.to_pydict()
+        assert isinstance(data, dict)
+        assert "table_name" in data
+        assert isinstance(data["table_name"], list)
+
```
fengelniederhammer: This is also almost a duplication of `test_get_tables_single_table`

src/silo/query_engine/filter/expressions/or.cpp
```diff
@@ -112,6 +112,9 @@ ExpressionVector Or::rewriteSymbolInSetExpressions(ExpressionVector children) {
    return new_children;
 }
 
+template ExpressionVector Or::rewriteSymbolInSetExpressions<AminoAcid>(ExpressionVector children);
+template ExpressionVector Or::rewriteSymbolInSetExpressions<Nucleotide>(ExpressionVector children);
```
fengelniederhammer: Why do we need these now?

PR #1149, opened 2026-02-13
1054 error invalid positions

### LINKED ISSUE #1054: Also error if a user tries to insert invalid insertion positions
https://github.com/GenSpectrum/LAPIS-SILO/pull/1039 resolved the issue that user's could ask for invalid insertion position and received an empty result instead of an error. While that has been fixed, the maintainer is still able to insert positions to SILO that have a position outside the range [0, n]. 

Note that although positions are 1-based, an insertion at 0 is allowed to insert before the first symbol. The semantics of "insertion at position `x`" is "insert after position `x`" as documented by [nextstrain](https://docs.nextstrain.org/projects/nextclade/en/stable/user/algorithm/05-mutation-calling.html#mutations-relative-to-reference-sequence):
> 22030:ACT would indicate that the query sequence has the three bases ACT inserted between position 22030 and 22031 in the reference sequence (the indices are 1-based).

taepper:
resolves #1054

### Summary

We now correctly error if the position is not in `[0,genome_length]`

## PR Checklist


Taepper:
> Looks good, but when I try to add an insertion at a negative position, then I get this funny error:
> "preprocessing - error: preprocessing - exception when appending data: the insertion position (4294941595) is larger than the length of the reference sequence (29903)"
> 

Nice catch! Should be fixed and tested for now

PR #1150, opened 2026-02-13
refactor(silo): replace boost::join with fmt::join

### LINKED ISSUE #999: Replace `boost::join` with `fmt::join`
`fmt::join` is more flexible as it has support for ranges.

Therefore, this works:
```
#include <fmt/ranges.h>
...
std::map<std::filesystem::path, std::string> some_map = foo();
auto filenames =
   some_map | std::views::keys |
   std::views::transform([](const std::filesystem::path& p) { return p.filename(); });
auto joined = fmt::join(filenames, ",");
```

while
```
#include <boost/algorithm/string/join.hpp>
...
auto joined = boost::join(filenames, ",");
```

We should standardize on one version and reduce our dependence on boost in this regard

taepper:
resolves #999

### Summary

This replace `boost::join` with `fmt::join`. This makes us less reliant on boost and fmt has a nicer/more modern api as well

## PR Checklist

### DIFF
--- a/src/config/config_key_path.cpp
+++ b/src/config/config_key_path.cpp
@@ -1,7 +1,8 @@
 #include "config/config_key_path.h"
 
+#include <algorithm>
+
 #include <fmt/format.h>
-#include <boost/algorithm/string/join.hpp>
 
 namespace {
 bool isLowerCaseOrNumeric(char character) {
--- a/src/silo/common/string_utils.cpp
+++ b/src/silo/common/string_utils.cpp
@@ -3,7 +3,7 @@
 #include <stdexcept>
 
 #include <fmt/format.h>
-#include <boost/algorithm/string/join.hpp>
+#include <fmt/ranges.h>
 #include "silo/common/panic.h"
 
 namespace silo {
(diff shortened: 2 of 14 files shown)


src/silo/common/string_utils.cpp
```diff
@@ -81,7 +81,7 @@ std::string tieAsString(
    const std::vector<std::string>& elements2,
    std::string_view suffix
 ) {
-   return boost::join(tie(prefix, elements1, delimiter, elements2, suffix), "");
+   return fmt::format("{}", fmt::join(tie(prefix, elements1, delimiter, elements2, suffix), ""));
```
fengelniederhammer: Why do we need `fmt::format` here? Inserting into `"{}"` doesn't seem to do much?

src/config/config_key_path.cpp
```diff
@@ -1,7 +1,8 @@
 #include "config/config_key_path.h"
 
+#include <algorithm>
```
fengelniederhammer: Why do we need this include?

Taepper: So a compile error is generated if we leave it out :)

Issue #659, opened 2024-12-05
Error message doesn't state field name of field that has incorrectly formatted date


corneliusroemer:
I've encountered this in SILO prepro logs in Loculus, would be nice to state which field contained this wrongly formatted date:

```

2024-12-05 16:55:47.382 | [2024-12-05 15:55:47.382] [logger] [error] [api.cpp:289] Invalid Input Error: JSON transform error in file "/preprocessing/input/data.ndjson.zst", in line 1: date field value out of range: "2004", expected format is (YYYY-MM-DD)
-- | --
```

It says `date field value out of range` but it doesn't say _what that field is called_.

PR #1153, opened 2026-02-16
feat(silo): add a DateEquals expression

### LINKED ISSUE #414: DateEquals
Add a Filter for DateEquals. Then we have an equality filter for all column types

taepper:
resolves #414

### Summary

This adds a `DateEquals` expression for completeness

## PR Checklist

### DIFF
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -62,10 +62,16 @@ This filter is true, if the symbol of nucleotide sequence `sequenceName` at posi
 
 #### `DateBetween`: `{"column": string, "from": string | null, "to": string | null}`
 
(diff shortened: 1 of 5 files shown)


fengelniederhammer (changes requested):
I think we should validate the date format of `value` and error if it's not valid:

* `"value": "2021-03-00018"` returns the same as `"value": "2021-03-18"` (which is odd but might still be ok)
* `"value": "asdf"` silently returns nothing (which is something where we should show "bad request" IMO)

Taepper:
Okay, then I would first change our Date handling to error correctly in all cases, and move this to blocked.

Wanted to grab #659 next anyways

PR #1188, opened 2026-03-02
feat(silo): allow sequence-column inputs to be zstd-compressed in base64 format

### LINKED ISSUE #1186: Enable the input of zstd-compressed sequences in base64 encoding
This can speed up the ingestion of sequences, as no unnecessary de-compressions and compressions need to happen outside of silo

taepper:
resolves #1186

### Summary

Sometimes, users of SILO already have the sequences in zstd compressed format. Therefore, it can help processing times to avoid the decompression in the client software. This will also reduce input file-sizes by a lot in those cases

## PR Checklist

### DIFF
--- a/src/silo/common/base64.cpp
+++ b/src/silo/common/base64.cpp
@@ -1,0 +1,43 @@
+#include "silo/common/base64.h"
+
+#include <expected>
+#include <string>
+#include <string_view>
+
+#include <simdutf.h>
+
+namespace silo {
+
+std::expected<std::string, std::string> decodeBase64(std::string_view encoded) {
+   const size_t max_size =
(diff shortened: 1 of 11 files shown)


src/silo/common/base64.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <expected>
+#include <string>
+#include <string_view>
+
+#include <simdutf.h>
+
+namespace silo {
+
+std::expected<std::string, std::string> decodeBase64(std::string_view encoded) {
```
fengelniederhammer:
```suggestion
std::expected<std::string, std::string> decodeBase64(const std::string_view encoded) {
```

Issue #1151, opened 2026-02-14
Enable SILO to automatically cluster inputs


taepper:
SILO's core sequence storage already divides sequences into blocks of `2^16 = 65536` sequences, which are then stored in roaring containers (these also have `2^16` as maximum cardinality). On inserts we always append the data to the end of the last "block" or add a new block if the last block is full.

We can remove the necessity for most data-wrangling pre SILO-insertion, if instead of always appending, we also add _some_ clustering logic directly into SILO. Instead of appending to the LAST cluster, we have some sketch that helps us decide which cluster we want to add the sequence to. This can improve compression by a lot

And explicitly, this removes the need to (by decreasing order of relevance):
1. sort sequences by offset before inserting
2. cluster sequences by lineage before inserting
(3. sorting sequences by date before inserting)?

A scheme for clustering the input data can rely on:
- null-values for segments
- subtype as defined by a metadata field
- coverage information for segments (relevant for short-read silo)

PR #1226, opened 2026-03-27
refactor(silo): remove remaining parts of partitioning

taepper:
Partitioning in SILO has been "disabled" for some time now, and we do not plan on reintroducing it. Rather approaches such as #1151 will be followed regarding this direction.

This removes the partition code paths from SILO


src/silo/storage/column/sequence_column.cpp
fengelniederhammer: This log message still mentions a partition:
```suggestion
      "Sequence store info after filling it: {}, after local reference adaption: {}, and "
```

fengelniederhammer (approved):
Nice, only minor stuff.

also the `data_directories.md` says:

```
    default/
      ...  (partition data)
```
Which might be worth updating.

PR #1261, opened 2026-05-20
feat(silo)!: use date32 type for dates instead of converting to string

taepper:
resolves #1225 

### Summary

This contains a breaking change: in the arrow format, we will directly emit dates using arrow's date32 type. ndjson format will still convert them to strings

LAPIS will need to add a case statement for a DateVector here:
https://github.com/GenSpectrum/LAPIS/blob/05298cf0d2688ff13819db10f82022f3de698382/lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt#L184-L191

## PR Checklist


fengelniederhammer (approved):
LGTM

PR #1269, opened 2026-05-29
refactor(silo): remove internal helper operator ScanNode

taepper:
### Summary

This internal helper node made optimizer passes more difficult to be adapted in the future. It is a lot easier that the node is a `TableScan` with a proper bound output-schema from the start. This also allows better optimizations with regards to making filter pushdown more independent of node resolution

## PR Checklist

]Note that the removed test-case in `planner.test.cpp` was already present in `ast_to_query.test.cpp`, which is now the appropriate place for the test


src/silo/query_engine/planner.test.cpp
```diff
@@ -76,24 +76,10 @@ class ErrorQueryNode final : public operators::QueryNode {
    }
 
    [[nodiscard]] operators::NodeKind kind() const override {
-      return operators::NodeKind::TABLE_SCAN;
+      return operators::NodeKind::COUNT_FILTER;
    }
```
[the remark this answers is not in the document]
taepper: As this is only in a test file, I think this is fine. I will add a comment instead

src/silo/query_engine/planner.cpp
```diff
@@ -62,34 +61,33 @@ operators::QueryNodePtr wrapWithDecompressIfNeeded(
...
 
-/// Extracts table name and filter from a ScanNode or FilterNode(ScanNode) chain.
+/// Extracts the table and filter from a TableScanNode or FilterNode(TableScanNode) chain.
 std::optional<ExtractedScanInfo> extractScanInfo(operators::QueryNodePtr& node) {
-   auto* scan = dynamic_cast<operators::ScanNode*>(node.get());
-   if (scan != nullptr) {
-      return ExtractedScanInfo{
-         .table_name = scan->table_name, .filter = std::make_unique<filter::expressions::True>()
-      };
+   if (auto* scan = dynamic_cast<operators::TableScanNode*>(node.get())) {
```
fengelniederhammer, line 83: AI: 🟡 risk: `std::move(scan->filter)` leaves the `TableScanNode` in a moved-from state while still alive in the tree. Currently safe because callers always replace the node afterward, but this is a fragile contract — if anything later accesses the node's filter (e.g. error paths, debugging), it's a null deref. Consider adding a comment documenting this ownership transfer.

src/silo/query_engine/column_narrowing_pass.cpp
```diff
@@ -24,14 +24,14 @@ void applyToChild(operators::QueryNodePtr& child, ColumnNarrowingPass& pass) {
 }  // namespace
 
 // NOLINTNEXTLINE(readability-convert-member-functions-to-static,misc-no-recursion)
-operators::QueryNodePtr ColumnNarrowingPass::operator()(operators::ScanNode& node) {
+operators::QueryNodePtr ColumnNarrowingPass::operator()(operators::TableScanNode& node) {
    std::vector<schema::ColumnIdentifier> pruned;
-   for (const auto& col : node.output_schema) {
+   for (const auto& col : node.fields) {
       if (std::ranges::find(required, col) != required.end()) {
```
fengelniederhammer, line 30: AI: ❓ question: Now that `TableScanNode` carries both `fields` (output columns) and `filter` (filter expression), the narrowing pass prunes `fields` without considering which columns the filter references. This is currently safe because `computeFilter` evaluates the filter expression against the table's bitmaps directly — it doesn't go through the `fields` list. `fields` only controls which Arrow columns get materialized in the output. But if that invariant ever changes (e.g. filter evaluation starts consulting `fields` to know what's available), pruning could silently break filter evaluation. Worth a comment documenting that `fields` and `filter` operate on independent column sets?

fengelniederhammer (commented):
AI found a few things, looks good otherwise 👍

src/silo/query_engine/planner.cpp
taepper: https://github.com/GenSpectrum/LAPIS-SILO/pull/1275 changes the contract of this function already

src/silo/query_engine/column_narrowing_pass.cpp
taepper: I think this is very intuitive, but I will add a comment

fengelniederhammer (approved):

Issue #1139, opened 2026-02-06
Specify that any error in python bindings invalidate the database or make insert atomic


Taepper:
When an exception is thrown in the python bindings, the database might currently be left in an invalid state. This should be changed so either the database object is marked as invalid, or, ideally, insertions be fully atomic -> they can only go from valid state to valid state

Taepper:
Atomic would be the proper fix, as "only invalidating the database" can lead to throwing away data that if the python api user did not save the database state before inserting.

I will see how difficult it would be to fix this properly

taepper:
Data is now added to the column structs in groups of rows. All validation happen during the "building of chunks". The `appendChunk` method afterwards should be non-failing.

fengelniederhammer:
I think it makes sense to wait with this until #1295 is done. It solves a part of the problems for this issue.

PR #1278, opened 2026-06-02
Chunked insert

### LINKED ISSUE #1277: Append to columns in chunks
Currently the interface of a column offers an `insert` method. This should be changed to `appendChunk`, which adds a group of sequences (up to `1 << 16`).

taepper:
resolves #1277 

### Summary

This changes the insert interface of columns. 

Instead of directly appending to a `ColumnGroup`, the `TableInserter` has a `ColumnGroupBuilder` to which it appends each row

This enables many following changes which we want to have:
1. allow a more intelligent insertion strategy that also cluster's inputs and builds many (clustered) chunks at once which are then appended once full
2. allow the storage to change to immutable data structures for each chunk of data
2a. these chunks are rewritten on deletions or updates
2b. the chunk representation for `SequenceColumns` can eventually be streamed through the query engine

## PR Checklist


src/silo/storage/table.cpp
```diff
@@ -48,6 +65,16 @@ void Table::validate() const {
...
 }
 
+std::expected<void, std::string> Table::bulkInsert(ColumnGroupBuilder& block) {
+   for (const auto& column : columns.metadata) {
+      auto result = column::visit(column.type, BulkInsertVisitor{}, columns, block, column.name);
+      if (!result.has_value()) {
+         return result;
+      }
+   }
+   return {};
```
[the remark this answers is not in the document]
taepper: Possibly to be done in #1139

src/silo/append/database_inserter.cpp
```diff
@@ -165,22 +165,37 @@ std::expected<std::vector<TableInserter::SniffedField>, std::string> TableInsert
...
    return {};
 }
 
-TableInserter::Commit TableInserter::commit() const {
+TableInserter::Commit TableInserter::commit() {
+   // Flush the final, partially-filled chunk.
+   if (column_builder.numBufferedRows() > 0) {
+      auto bulk_insert_result = table->bulkInsert(column_builder);
+      if (!bulk_insert_result.has_value()) {
+         throw silo::append::AppendException(bulk_insert_result.error());
```
fengelniederhammer:
```suggestion
         throw AppendException(bulk_insert_result.error());
```

src/silo/append/database_inserter.cpp
```diff
@@ -165,22 +165,37 @@ std::expected<std::vector<TableInserter::SniffedField>, std::string> TableInsert
...
    for (const auto& sniffed_field : field_order_hint) {
       ASSIGN_OR_RAISE(auto column_value, findFieldWithFallbacks(object, sniffed_field));
       auto success_or_error =
-         table->columns.addJsonValueToColumn(sniffed_field.column_identifier, column_value);
+         column_builder.addJsonValueToColumn(sniffed_field.column_identifier, column_value);
       if (!success_or_error.has_value()) {
          return success_or_error;
       }
    }
    table->sequence_count++;
```
fengelniederhammer: AI: 🟡 risk: `sequence_count` is incremented per-row *before* `bulkInsert()` is called at the chunk boundary. If `bulkInsert()` fails, `sequence_count` has already been bumped for all rows in the chunk but column data may be partially applied — `validate()` would catch the mismatch, but the error path in `insert()` just returns the error without rolling back `sequence_count`.

Consider incrementing `sequence_count` inside `bulkInsert()` (after successful append), or tracking it via `column_builder.numBufferedRows()` and adding in bulk after flush.

src/silo/storage/column_group_builder.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+size_t ColumnGroupBuilder::numBufferedRows() const {
+   if (metadata.empty()) {
+      return 0;
+   }
+   const auto& first = metadata.front();
+   return column::visit(
+      first.type, NumValuesVisitor{}, const_cast<ColumnGroupBuilder&>(*this), first.name
+   );
+}
```
fengelniederhammer, line 326: AI: 🟡 risk: `numBufferedRows()` only checks the first column's builder count. If a row extraction fails mid-way (e.g., column A's builder got `insert()` but column B's extraction returned an error), builders go out of sync — the chunk size check could be wrong.

Consider: on extraction error, either roll back the partially-inserted row from all builders that already received it, or document that the builder is invalidated on error and must not be reused.

src/silo/storage/column/sequence_column.h
```diff
@@ -143,6 +157,51 @@ class SequenceColumn {
...
+
+   // The column's current local reference (possibly adapted by a previous
+   // finalize); inserted sequences are diffed against this so newly buffered
+   // rows share the same reference basis as the already-stored ones.
+   std::string local_reference;
+
+  public:
+   // Decompresses 'sequenceCompressed' input during phase-1 extraction. Uses the
+   // initial (unadapted) reference, which is what inputs are compressed against.
+   ZstdDecompressor compressed_input_decompressor;
```
fengelniederhammer, line 176: AI: 🔵 nit: `compressed_input_decompressor` is public here and accessed directly in `column_group_builder.cpp`. Consider exposing a `decompress()` wrapper method on the builder instead of making the decompressor a public member.

src/silo/storage/column/sequence_column.test.cpp
```diff
@@ -12,13 +16,31 @@ using silo::storage::InsertionFormatException;
...
+   std::string_view sequence,
+   uint32_t offset,
+   const std::vector<std::string>& insertions
+) {
+   SequenceColumn<Nucleotide>::Builder builder(
+      column.metadata, column.local_reference_sequence_string
+   );
+   builder.insert(sequence, offset, insertions);
+   auto result = column.appendChunk(builder.finalize());
+   (void)result;
```
fengelniederhammer: AI: 🟡 risk: `appendChunk` result is silently discarded with `(void)result`. If it ever returns an error, the test won't notice. Should be `ASSERT_TRUE(result.has_value())` or `SILO_ASSERT(result.has_value())`.

src/silo/storage/column/sequence_column.cpp
```diff
@@ -319,8 +300,89 @@ SequenceColumnMetadata<SymbolType>::SequenceColumnMetadata(
...
+      );
+   }
+   SILO_ASSERT(sequence.size() + offset < UINT32_MAX);
+
+   typename SequenceColumn<SymbolType>::BufferedSequence buffered;
+   buffered.insertions = insertions;
+
+   // Single pass over the sequence computes both the covered range / missing
+   // positions (coverage) and the deviations from the local reference
+   // (mutations), mirroring HorizontalCoverageIndex::insertSequenceCoverage and
```
fengelniederhammer: AI: 🔵 nit: Since the builder now computes coverage inline (this single-pass loop), `HorizontalCoverageIndex::insertSequenceCoverage()` is dead code in production — only `horizontal_coverage_index.test.cpp` still calls it. Consider removing it (tests can use `insertCoverage()` directly) or marking it as test-only.

fengelniederhammer (commented):
Some of the comments might be slop, didn't have time to check them thoroughly anymore.

src/silo/storage/column_group_builder.cpp
taepper: To be done in #1139

src/silo/storage/table.cpp
```diff
@@ -48,6 +65,17 @@ void Table::validate() const {
    validateMetadataColumns();
 }
 
+std::expected<void, std::string> Table::bulkInsert(ColumnGroupBuilder& block) {
+   sequence_count += block.numBufferedRows();
```
fengelniederhammer, line 74: AI: 🟡 risk: `sequence_count += block.numBufferedRows()` runs before the column append loop (L70-75). If any column's `appendChunk` fails, `sequence_count` is already bumped but columns are in an inconsistent state. Consider moving this after the loop, or incrementing inside each column's `appendChunk` and reconciling.

fengelniederhammer (approved):

PR #1289, opened 2026-06-08
refactor(silo): move towards unifying filter and scalar expressions

taepper:
Small refactoring commit for progress on #1252 and #filter #1273

PR #1290, opened 2026-06-08
refactor(silo): rename namespace silo::query_engine::filter::expressions to silo::query_engine::expressions

taepper:
follow-up to #1289


src/silo/query_engine/expressions/date_between.cpp
```diff
@@ -58,30 +59,36 @@ std::unique_ptr<operators::Operator> DateBetween::compile(const storage::Table&
...
+      )
+   );
+   predicates.emplace_back(
+      std::make_unique<filter::operators::CompareToValueSelection<Date32Column>>(
+         date_column,
+         filter::operators::Comparator::LESS_OR_EQUALS,
+         date_to.value_or(std::numeric_limits<silo::common::Date32>::max())
+      )
+   );
+   return std::make_unique<filter::operators::Selection>(
```
fengelniederhammer, line 81: `using` those classes would have resulted in less git noise ;)

fengelniederhammer (approved):

PR #1297, opened 2026-06-09
ci: add to changelog and to release notes when serialization version changed

### LINKED ISSUE #1224: Add serialization version changes to release notes
Brought up in #1223 

> A possible next step, but I would find it quite useful if the changelog would contain a note if the serialization version changed in a SILO version.

> Maybe we can also customize release please somehow via a plugin or whatever? Let's create an issue? We can throw an LLM at it, maybe it will figure out something.

It cannot directly amend the file in the repo, as release-please re-generates the release notes. So it must be via commit message or release please plugin

fengelniederhammer:
resolves #1224

### Summary

Adds two CI steps to the `release-please` job in `ci.yml` that automatically detect serialization version changes and surface them in release notes:

1. **Annotate CHANGELOG.md on the release PR branch** — After release-please creates/updates its PR, compares `serialization_version.txt` between the latest release tag and `origin/main`. If changed, commits a "⚠ Serialization Version Changed" section into `CHANGELOG.md` on the release branch.

2. **Annotate the GitHub Release** — After the release PR is merged, appends the same note to the GitHub Release body via `gh release edit`.

Both steps are no-ops when the serialization version hasn't changed.

### Why this approach

Release-please has no plugin/hook system for injecting custom content into the changelog — it only derives changelog entries from conventional commit messages. The available extension points (`changelog-sections`, strategies, updaters) don't support file-diff-based annotations.

We considered requiring developers to use a specific commit message format (e.g. `feat(serialization): ...`) when bumping the serialization version, but this is error-prone — bumps are often bundled into feature commits and would be silently missed.

Instead, this approach detects the change automatically by comparing the `serialization_version.txt` file between releases. It runs *after* release-please force-pushes its branch, so the note is re-applied on every update cycle and doesn't interfere with release-please's operation.

## PR Checklist

- [ ] ~~The implemented feature is covered by an appropriate test.~~

### DIFF
--- a/.github/scripts/annotate-changelog-serialization-version.sh
+++ b/.github/scripts/annotate-changelog-serialization-version.sh
@@ -2,12 +2,10 @@
 
 set -euo pipefail
 
-# Annotate CHANGELOG.md with a serialization version change note.
+# Check whether the serialization version changed since the last release,
+# and if so, annotate CHANGELOG.md on the release branch and commit+push.
 #
-# Compares the serialization version file between the latest release tag
-# and origin/main. If it changed, inserts a warning section into the
-# CHANGELOG.md entry for the given release version and commits+pushes
-# the change.
+# Delegates the actual changelog modification to insert-serialization-version-note.sh.
 #
 # Usage:
 #   ./annotate-changelog-serialization-version.sh --branch=<release-branch> --version=<release-version>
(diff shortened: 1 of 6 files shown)


fengelniederhammer:
I didn't test it e2e, I'm not 100% sure it works. But we can try and reiterate, if something doesn't work yet.

.github/scripts/annotate-changelog-serialization-version.sh
```diff
@@ -0,0 +1,65 @@
...
+"${SCRIPT_DIR}/insert-serialization-version-note.sh" --version="$VERSION" \
+  < CHANGELOG.md > CHANGELOG.md.tmp && mv CHANGELOG.md.tmp CHANGELOG.md
+
+if [ "$DRY_RUN" = "true" ]; then
+  echo "DRY_RUN: would commit and push CHANGELOG.md changes"
+  exit 0
+fi
+
+git config user.name "github-actions[bot]"
+git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
taepper, line 61: Do we really want to manually insert a `..noreply..` e-mail?

taepper (approved):

.github/scripts/annotate-changelog-serialization-version.sh
fengelniederhammer: At least the AI claimed that it's the standard thing.

PR #1302, opened 2026-06-10
refactor(silo): organize value columns in chunks so appendChunk does not modify existing data structures

### LINKED ISSUE #1301: Refactor int, float, date columns to store immutable chunks
These columns are easiest to implement and should be implemented together in one PR

taepper:
resolves #1301

### Summary

First of multiple PRs to make our storage layout more modular

## PR Checklist


src/silo/storage/column/chunked_value_buffer.h
```diff
@@ -0,0 +1,68 @@
...
+#include <boost/serialization/access.hpp>
+#include <boost/serialization/vector.hpp>
+
+namespace silo::storage::column {
+
+/// Stores a value column's data as the list of fixed-width buffers in which it was ingested, one
+/// buffer per appended chunk. Appending never touches previously ingested chunks, and a row id is
+/// mapped to its chunk with a binary search over `chunk_end_offsets`.
+template <typename T>
+class ChunkedValueBuffer {
```
fengelniederhammer, line 15: Is it really a "buffer" (makes me assume its temporary)? Or is it more a "storage" (makes me assume its persistent)?

src/silo/storage/column/float_column.cpp
```diff
@@ -1,22 +1,23 @@
...
+   std::vector<double> chunk;
+   chunk.reserve(buffer.size());
+   for (size_t i = 0; i < buffer.size(); ++i) {
+      if (buffer[i].has_value()) {
+         chunk.push_back(*buffer[i]);
       } else {
-         null_bitmap.add(values.size());
-         values.push_back(std::nan(""));
+         null_bitmap.add(base + i);
+         chunk.push_back(0.0);
```
fengelniederhammer, line 17: We just store something to get the size right? But this should never be read, because we check the `null_bitmap`? It seems a little dangerous to me.

src/silo/storage/column/date32_column.test.cpp
```diff
@@ -45,3 +45,73 @@ TEST(Date32Column, parseInvalidDateReturnsError) {
...
+   ASSERT_EQ(under_test.numValues(), 6);
+
+   const std::vector<bool> expected_null{false, true, true, false, true, false};
+   for (size_t row_id = 0; row_id < expected_null.size(); ++row_id) {
+      ASSERT_EQ(under_test.isNull(row_id), expected_null[row_id]) << "row_id=" << row_id;
+   }
+
+   ASSERT_FALSE(under_test.isNull(0));
+   ASSERT_EQ(silo::common::date32ToString(under_test.getValue(3)), "2021-03-21");
+   ASSERT_EQ(silo::common::date32ToString(under_test.getValue(5)), "2022-01-01");
```
fengelniederhammer, line 117: What does asking for a null value return (`under_test.getValue(1)`)?

src/silo/storage/column/float_column.cpp
taepper: This was already the case before

src/silo/storage/column/chunked_value_buffer.h
taepper: Once we introduce a buffer manager, these buffers, being in-memory, are indeed temporary. But in general, the term buffer can also be used for the range of (volatile) memory used for an in-memory data-structure

Do you happen to think of a better term we can use here?

fengelniederhammer: Convinced, sounds good :+1:

fengelniederhammer (approved):

src/silo/storage/column/date32_column.test.cpp
taepper: This would indeed return the "dummy value" introduced before. This same contract also exists for arrow arrays

PR #1303, opened 2026-06-10
refactor(silo): add equality operator to expressions

taepper:
This enables simplification to the expression rewrite / optimization logic


taepper:
This was supposed to offer better internal abstraction during query planning, but it might not work well in general use. Let's solve the problem at hand rather with a `kind()` method in expressions to allow cleaner branching in the optimizer / rewriter

Let's close this in favor of #1306

PR #1305, opened 2026-06-10
feat: add `unionAll` operator to SaneQL queries

### LINKED ISSUE #1221: Add a `UnionAll` operator
After changing to saneql the front-end for this new operator is clear. We should implement this to enable better multi-table support

fengelniederhammer:
resolves #1221

### Summary

Adds a `unionAll` operator to the SaneQL interface.

## PR Checklist

### DIFF
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -288,7 +288,7 @@ default.filter(pango_lineage = 'B.1.1.7').phyloSubtree('usherTree')
 
 Concatenates the output of two pipelines. `unionAll` can be called as a standalone function or with piped syntax:
 
-Both inputs must have the same schema (same column names and types). Column order does not matter — the right child is automatically reordered to match the left child's column order.
+Both inputs must have the same schema (same column names, types, and order).
 
 All rows from both inputs are included — duplicates are preserved (UNION ALL, not UNION).
 
@@ -353,7 +353,7 @@ unionAll(
 )
 ```
 
-**Output:** all rows from the left input followed by all rows from the right input, with the left input's column order.
+**Output:** all rows from both inputs. The order of rows is not guaranteed.
 
 ---
 
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -326,8 +326,25 @@ unionAll(
 
 **Restrictions:**
 
-- `.filter(...)` above a `unionAll` is not supported. Apply filters inside each child pipeline instead.
 - `mutations()`, `aminoAcidMutations()`, `insertions()`, and similar operators that require a table scan cannot be applied to the result of a `unionAll`. They can however be used inside each child.
+
+Filters above a `unionAll` are automatically pushed into both children:
+
+```
+unionAll(
+  default.project({primaryKey, country}),
+  default.project({primaryKey, country})
+).filter(country='CH')
+```
+
+is equivalent to:
+
+```
+unionAll(
+  default.filter(country='CH').project({primaryKey, country}),
+  default.filter(country='CH').project({primaryKey, country})
+)
+```
 
 **Output:** all rows from the left input followed by all rows from the right input, with the left input's column order.
 
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -20,10 +20,11 @@ The table name is `default` for now.
 
 Every operator takes a table as input and produces a table as output. Internally these tables are Apache Arrow record batches; externally they are streamed as NDJSON or Arrow IPC. The **response schema** — which fields are returned and their types — is always the output schema of the **last operator** in the pipeline.
 
-Operators fall into two categories:
+Operators fall into three categories:
 
 - **Schema-preserving** (`filter`, `orderBy`, `limit`, `offset`, `randomize`): pass all input columns through unchanged. They select or reorder rows but do not add, remove, or rename fields.
 - **Schema-defining** (`groupBy`, `project`, `map`, `mutations`, `aminoAcidMutations`, `insertions`, `aminoAcidInsertions`, `mostRecentCommonAncestor`, `phyloSubtree`): produce a changed output schema. Each operator's section below documents its output fields.
+- **Combining** (`unionAll`): takes two pipelines and concatenates their output. The output schema matches the children's schema.
 
 Simple example — count all sequences from Switzerland:
 
@@ -288,6 +289,47 @@ default.filter(pango_lineage = 'B.1.1.7').phyloSubtree('usherTree')
 ```json
 {"subtreeNewick": "((key_83:0.00027051)NODE_0000077:3.291e-05,(...)NODE_0000079:1e-06)NODE_0000076;", "missingNodeCount": 0}
 ```
+
+### `unionAll(left, right)`
+
+Concatenates the output of two pipelines. Unlike the other operators, `unionAll` is a **standalone function** — it is not chained on a single pipeline but takes two pipeline expressions as arguments.
+
+Both inputs must have the same schema (same column names and types). Column order does not matter — the right child is automatically reordered to match the left child's column order.
+
+All rows from both inputs are included — duplicates are preserved (UNION ALL, not UNION).
+
+```
+unionAll(
+  default.filter(division='Aargau').project({division}),
+  default.filter(division='Bern').project({division})
+)
+```
+
+The result can be piped into downstream operators:
+
+```
+unionAll(
+  default.filter(division='Aargau').project({division}),
+  default.filter(division='Bern').project({division})
+).groupBy({count:=count()}, {division})
+ .orderBy({asc(division)})
+```
+
+`unionAll` calls can be nested:
+
+```
+unionAll(
+  unionAll(pipelineA, pipelineB),
+  unionAll(pipelineC, pipelineD)
+)
+```
+
+**Restrictions:**
+
+- `.filter(...)` above a `unionAll` is not supported. Apply filters inside each child pipeline instead.
+- `mutations()`, `aminoAcidMutations()`, `insertions()`, and similar operators that require a table scan cannot be applied to the result of a `unionAll`. They can however be used inside each child.
+
+**Output:** all rows from the left input followed by all rows from the right input, with the left input's column order.
 
 ---
 
@@ -566,4 +608,14 @@ default
   .filter(aminoAcidInsertionContains(position:=214, value:='.*PE', sequenceName:='S'))
   .aminoAcidInsertions()
   .orderBy({insertedSymbols, position})
+```
+
+### Combine two filtered groups with unionAll
+
+```
+unionAll(
+  default.filter(division='Aargau').project({division}),
+  default.filter(division='Bern').project({division})
+).groupBy({count:=count()}, {division})
+ .orderBy({asc(division)})
 ```
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -20,12 +20,6 @@ The table name is `default` for now.
 
 Every operator takes a table as input and produces a table as output. Internally these tables are Apache Arrow record batches; externally they are streamed as NDJSON or Arrow IPC. The **response schema** — which fields are returned and their types — is always the output schema of the **last operator** in the pipeline.
 
-Operators fall into three categories:
-
-- **Schema-preserving** (`filter`, `orderBy`, `limit`, `offset`, `randomize`): pass all input columns through unchanged. They select or reorder rows but do not add, remove, or rename fields.
-- **Schema-defining** (`groupBy`, `project`, `map`, `mutations`, `aminoAcidMutations`, `insertions`, `aminoAcidInsertions`, `mostRecentCommonAncestor`, `phyloSubtree`): produce a changed output schema. Each operator's section below documents its output fields.
-- **Combining** (`unionAll`): takes two pipelines and concatenates their output. The output schema matches the children's schema.
-
 Simple example — count all sequences from Switzerland:
 
 ```
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -292,7 +292,7 @@ default.filter(pango_lineage = 'B.1.1.7').phyloSubtree('usherTree')
 
 ### `unionAll(left, right)`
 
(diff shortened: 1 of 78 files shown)


taepper:
As per arrow documentation there is a `union` we could reuse:
https://arrow.apache.org/docs/cpp/acero/user_guide.html#union

Although, there is a point to be made, where we want more control over our query execution layer

src/silo/query_engine/operators/union_all_node.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <arrow/result.h>
+
+#include "silo/query_engine/operators/query_node.h"
+#include "silo/schema/database_schema.h"
+#include "silo/storage/table.h"
+
+namespace silo::query_engine::operators {
+
+/// Concatenates the output of two child pipelines (UNION ALL semantics).
```
taepper, line 14: this mentions two child pipelines (and all tests are also using 2 child expressions), then why don't we just have
`QueryNodePtr left; QueryNodePtr right;`
instead of
`std::vector<QueryNodePtr> children;`

src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1224,6 +1269,8 @@ FunctionRegistry::FunctionRegistry() {
         named("contractUnaryNodes", false)}},
       handlePhyloSubtree
    );
+
+   registerFunction("unionAll", {{pos("left"), pos("right")}}, handleUnionAll);
```
taepper, line 1255: Can we add a tests that use pipeline'd syntax (`<left>.unionAll(<right>)`) and name expressions `unionAll(left := <left>, right := <right>)`

src/silo/query_engine/filter_pushdown_pass.cpp
```diff
@@ -160,6 +162,24 @@ operators::QueryNodePtr FilterPushdownPass::operator()(operators::UnresolvedPhyl
    return nullptr;
 }
 
+// NOLINTNEXTLINE(misc-no-recursion)
+operators::QueryNodePtr FilterPushdownPass::operator()(operators::UnionAllNode& node) {
+   // Filters cannot be pushed across a union all boundary.
```
taepper: I do not quite understand why this is the case. Couldn't we have applied all filter both left and right?

src/silo/query_engine/operators/union_all_node.cpp
```diff
@@ -0,0 +1,170 @@
...
+      arrow::acero::BackpressureMonitor* backpressure_monitor = nullptr;
+      const arrow::acero::SinkNodeOptions sink_options{
+         &current_generator,
+         arrow::acero::BackpressureOptions{
+            /*.resume_if_below =*/silo::common::S_16_KB,
+            /*.pause_if_above =*/silo::common::S_64_MB
+         },
+         &backpressure_monitor,
+         true
+      };
```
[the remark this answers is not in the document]
fengelniederhammer: Not true.

src/silo/query_engine/operators/union_all_node.cpp
```diff
@@ -0,0 +1,170 @@
...
+   while (true) {
+      if (state->child_idx >= state->child_plans.size()) {
+         return arrow::Future{std::optional<arrow::ExecBatch>{std::nullopt}};
+      }
+
+      if (!state->generator_active) {
+         auto status = state->startCurrentChild();
+         if (!status.ok()) {
+            return arrow::Future{arrow::Result<std::optional<arrow::ExecBatch>>{status}};
+         }
```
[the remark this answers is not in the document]
fengelniederhammer: AI says: `arrow::Future<T>` has an implicit converting constructor `Future(Result<T> res)` — declared in `arrow/util/future.h` with `// NOLINT runtime/explicit`. This is supported Arrow API, designed exactly for constructing already-finished futures from a result.

`TableScanGenerator` uses `Future<T>::Make()` + `MarkFinished()` for a different reason: it dispatches work to a detached `std::thread` (workaround for Arrow bugs #47641/#47642), so the future handle must exist *before* the result is ready. In `pullNext`, we already have the result synchronously in hand — the implicit constructor is the simpler and correct approach for that case.

src/silo/query_engine/operators/union_all_node.cpp
```diff
@@ -0,0 +1,170 @@
...
+      }
+
+      auto future = state->current_generator();
+      auto result = future.result();
+
+      if (!result.ok()) {
+         // Propagate all errors (including Cancelled) to the caller.
+         state->stopCurrentChild();
+         return arrow::Future{arrow::Result<std::optional<arrow::ExecBatch>>{result.status()}};
+      }
```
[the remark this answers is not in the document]
fengelniederhammer: AI says: Same reasoning as above — `Future(Result<T>)` is a supported implicit constructor in Arrow. It compiles and is the idiomatic way to create an already-finished future when the result is already available. `Future::Make()` + `MarkFinished()` is needed when the future must exist before async work completes (as in `TableScanGenerator`), which is not the case here.

src/silo/query_engine/operators/union_all_node.cpp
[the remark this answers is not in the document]
fengelniederhammer: @taepper what do you think? I can't really judge whether this is true or not.

fengelniederhammer:
> As per arrow documentation there is a `union` we could reuse: https://arrow.apache.org/docs/cpp/acero/user_guide.html#union
> 
> Although, there is a point to be made, where we want more control over our query execution layer

What do you think? I don't know Arrow well enough to judge this.

AI says: Arrow's `UnionNode` requires all input nodes to belong to the **same** `ExecPlan` instance (`UnionNode::Make(ExecPlan*, vector<ExecNode*>)`). In this codebase, `toQueryPlan()` returns a `PartialArrowPlan{top_node, plan}` where each node creates its own `ExecPlan` via `ExecPlan::Make()`. The left and right children each produce separate `ExecPlan` instances.

To use Arrow's `UnionNode`, `toQueryPlan()` would need to accept an existing `ExecPlan*` and add nodes to it, rather than each node creating its own plan. That's a signature change across all 18 `toQueryPlan()` implementations — an architectural refactor of the entire query engine. It would also change threading, backpressure, and error propagation semantics for every existing operator.

The current approach (build separate child plans, bridge them via a streaming source node) works within the existing architecture and already handles backpressure, cancellation, and ordered output correctly.

taepper:
> > As per arrow documentation there is a `union` we could reuse: https://arrow.apache.org/docs/cpp/acero/user_guide.html#union
> > Although, there is a point to be made, where we want more control over our query execution layer
> 
> What do you think? I don't know Arrow well enough to judge this.
> 
> AI says: Arrow's `UnionNode` requires all input nodes to belong to the **same** `ExecPlan` instance (`UnionNode::Make(ExecPlan*, vector<ExecNode*>)`). In this codebase, `toQueryPlan()` returns a `PartialArrowPlan{top_node, plan}` where each node creates its own `ExecPlan` via `ExecPlan::Make()`. The left and right children each produce separate `ExecPlan` instances.
> 
> To use Arrow's `UnionNode`, `toQueryPlan()` would need to accept an existing `ExecPlan*` and add nodes to it, rather than each node creating its own plan. That's a signature change across all 18 `toQueryPlan()` implementations — an architectural refactor of the entire query engine. It would also change threading, backpressure, and error propagation semantics for every existing operator.
> 
> The current approach (build separate child plans, bridge them via a streaming source node) works within the existing architecture and already handles backpressure, cancellation, and ordered output correctly.

In any case, this implementation does _not_ want to create a separate ExecPlan for the left child and right child. Instead they should be two nodes inside the same ArrowPlan. I also see now that this does in fact require some interface changes, but I do not think that they would be extensive

src/silo/query_engine/operators/union_all_node.cpp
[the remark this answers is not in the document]
taepper: This should be obsolete after changing the `toQueryPlan` interface

src/silo/query_engine/operators/union_all_node.test.cpp
```diff
@@ -0,0 +1,274 @@
...
+   .reference_genomes = REFERENCE_GENOMES
+};
+
+// Basic unionAll of two filtered pipelines from the same table
+const QueryTestScenario UNION_ALL_BASIC_SCENARIO = {
+   .name = "UNION_ALL_BASIC",
+   .query = R"(unionAll(
+      default.filter(country='CH').project({primaryKey, country}),
+      default.filter(country='DE').project({primaryKey, country})
+   ).orderBy({asc(primaryKey)}))",
```
fengelniederhammer, line 57: apparently Arrow's union node doesn't have a stable left-vs-right sort by itself.

fengelniederhammer:
> > > As per arrow documentation there is a `union` we could reuse: https://arrow.apache.org/docs/cpp/acero/user_guide.html#union
> > > Although, there is a point to be made, where we want more control over our query execution layer
> > 
> > 
> > What do you think? I don't know Arrow well enough to judge this.
> > AI says: Arrow's `UnionNode` requires all input nodes to belong to the **same** `ExecPlan` instance (`UnionNode::Make(ExecPlan*, vector<ExecNode*>)`). In this codebase, `toQueryPlan()` returns a `PartialArrowPlan{top_node, plan}` where each node creates its own `ExecPlan` via `ExecPlan::Make()`. The left and right children each produce separate `ExecPlan` instances.
> > To use Arrow's `UnionNode`, `toQueryPlan()` would need to accept an existing `ExecPlan*` and add nodes to it, rather than each node creating its own plan. That's a signature change across all 18 `toQueryPlan()` implementations — an architectural refactor of the entire query engine. It would also change threading, backpressure, and error propagation semantics for every existing operator.
> > The current approach (build separate child plans, bridge them via a streaming source node) works within the existing architecture and already handles backpressure, cancellation, and ordered output correctly.
> 
> In any case, this implementation does _not_ want to create a separate ExecPlan for the left child and right child. Instead they should be two nodes inside the same ArrowPlan. I also see now that this does in fact require some interface changes, but I do not think that they would be extensive

Changed it: `node.toQueryPlan` -> `node.addToExecPlan` that accepts a reference to a query plan. Using Arrow's union node seems to work.

documentation/query_documentation.md
```diff
@@ -20,10 +20,11 @@ The table name is `default` for now.
 
 Every operator takes a table as input and produces a table as output. Internally these tables are Apache Arrow record batches; externally they are streamed as NDJSON or Arrow IPC. The **response schema** — which fields are returned and their types — is always the output schema of the **last operator** in the pipeline.
 
-Operators fall into two categories:
+Operators fall into three categories:
 
 - **Schema-preserving** (`filter`, `orderBy`, `limit`, `offset`, `randomize`): pass all input columns through unchanged. They select or reorder rows but do not add, remove, or rename fields.
 - **Schema-defining** (`groupBy`, `project`, `map`, `mutations`, `aminoAcidMutations`, `insertions`, `aminoAcidInsertions`, `mostRecentCommonAncestor`, `phyloSubtree`): produce a changed output schema. Each operator's section below documents its output fields.
+- **Combining** (`unionAll`): takes two pipelines and concatenates their output. The output schema matches the children's schema.
```
taepper: This section was very messy already, and could probably be removed without substitution. (What are your thoughts here?)

If we keep it, `unionAll` would fit better into "schema-preserving"

src/silo/query_engine/operators/query_node.h
```diff
@@ -37,18 +37,17 @@ enum class NodeKind : uint8_t {
...
-   std::shared_ptr<arrow::acero::ExecPlan> plan;
+   UNION_ALL,
 };
 
 class QueryNode {
   public:
    virtual ~QueryNode() = default;
 
-   [[nodiscard]] virtual arrow::Result<PartialArrowPlan> toQueryPlan(
+   /// Add this node's exec nodes to an existing Arrow ExecPlan.
```
taepper:
```suggestion
   /// Translate this node (including its sub-nodes) to arrow::acero nodes and add them to an existing ExecPlan
```

src/silo/query_engine/operators/union_all_node.cpp
```diff
@@ -0,0 +1,35 @@
...
+}
+
+arrow::Result<arrow::acero::ExecNode*> UnionAllNode::addToExecPlan(
+   arrow::acero::ExecPlan& plan,
+   const std::map<schema::TableName, std::shared_ptr<storage::Table>>& tables,
+   const config::QueryOptions& query_options
+) const {
+   ARROW_ASSIGN_OR_RAISE(auto* left_node, left->addToExecPlan(plan, tables, query_options));
+   ARROW_ASSIGN_OR_RAISE(auto* right_node, right->addToExecPlan(plan, tables, query_options));
+   return arrow::acero::MakeExecNode("union", &plan, {left_node, right_node}, {});
```
taepper, line 24: Nice!

src/silo/query_engine/operators/union_all_node.test.cpp
taepper: Yes, exactly. It passes the (asynchronously) arriving batches through unchanged

src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1099,6 +1109,38 @@ operators::QueryNodePtr handlePhyloSubtree(
...
+   std::ranges::sort(right_sorted, {}, &schema::ColumnIdentifier::name);
+   CHECK_SILO_QUERY(
+      left_sorted == right_sorted,
+      "unionAll requires both inputs to have the same schema (same column names and types). "
+      "Left schema: [{}], right schema: [{}].",
+      fmt::join(namesWithTypes(left_sorted), ", "),
+      fmt::join(namesWithTypes(right_sorted), ", ")
+   );
+
+   // If the right child has different column order, wrap it in a ProjectNode to reorder columns
```
taepper: Do we want this? This seems like it can introduce weird behavior in edge cases. We already have a project node the user can use here

taepper (commented):
Looks good, a couple of comments

documentation/query_documentation.md
fengelniederhammer: Yes, let's remove it. I don't think it adds much (except having to maintain it)

taepper (approved):

PR #1309, opened 2026-06-11
Refactor to an explicit convertToScalar

taepper:
minor refactor to have an explicit function that converts part of an `ast` to a scalar expression


fengelniederhammer (approved):
I guess the first commit shouldn't actually be in this PR ;)

PR #1316, opened 2026-06-22
feat(silo): organize string columns in immutable chunks

### LINKED ISSUE #1307: Refactor string columns to store immutable chunks
Currently string columns use various storage backends, these should guarantee that a new data structure is started on every appendChunk call

taepper:
resolves #1307

### Summary

This refactors sequence columns for better mutability and chunked insertion. An `appendChunk` call should not interact with already stored data

## PR Checklist


src/silo/storage/column/string_column.h
```diff
@@ -58,45 +62,43 @@ class StringColumnMetadata : public ColumnMetadata {
...
+      if (value.size() <= SiloString::SHORT_STRING_SIZE) {
+         return fixed_string_data.insert(SiloString{value});
+      }
+      SILO_ASSERT(value.length() < UINT32_MAX);
+      auto suffix_id = variable_string_data.insert(value.substr(SiloString::PREFIX_LENGTH));
+      return fixed_string_data.insert(SiloString{
+         static_cast<uint32_t>(value.length()),
+         value.substr(0, SiloString::PREFIX_LENGTH),
+         suffix_id
+      });
```
fengelniederhammer: Do we want to keep the implementations here in the `.h` file?

src/silo/storage/column/string_column.h
```diff
@@ -117,6 +119,72 @@ class StringColumn {
...
+      return chunk_idx == 0 ? 0 : chunk_end_offsets[chunk_idx - 1];
+   }
+
+   /// Resolves a global row id to its chunk index and the row's offset within that chunk.
+   [[nodiscard]] std::pair<size_t, size_t> locate(size_t row_id) const {
+      // The chunk containing `row_id` is the first whose (exclusive) end offset exceeds it.
+      const auto chunk_idx = static_cast<size_t>(
+         std::ranges::upper_bound(chunk_end_offsets, row_id) - chunk_end_offsets.begin()
+      );
+      return {chunk_idx, row_id - chunkStart(chunk_idx)};
```
fengelniederhammer: nit: The offset-index math (`chunkStart`, `locate`, binary search over `chunk_end_offsets`, rebuild-on-load in `serialize`) is duplicated verbatim from `ChunkedValueBuffer`. Full reuse of `ChunkedValueBuffer` isn't possible here (it assumes one flat T per row, while `StringColumnChunk` is a two-level container with internal row-space), but the offset bookkeeping itself could be extracted into a tiny shared helper (e.g. `ChunkOffsetIndex`) to avoid maintaining the same logic in two places.

src/silo/storage/column/string_column.cpp
```diff
@@ -15,52 +15,49 @@ StringColumn::StringColumn(StringColumnMetadata* metadata)
...
          }
       } else {
-         null_bitmap.add(fixed_string_data.numValues());
-         fixed_string_data.insert(SiloString(""));
+         null_bitmap.add(row_id);
+         chunk.insertNull();
       }
    }
+   chunks.push_back(std::move(chunk));
+   chunk_end_offsets.push_back(base + buffer.size());
```
fengelniederhammer: nit: `base + buffer.size()` assumes a 1:1 mapping between buffer entries and inserted rows. `base + chunk.numValues()` would be more resilient — if insert semantics ever diverge (e.g. filtering/skipping), the offset stays correct by definition.

src/silo/storage/column/string_column.cpp
```diff
@@ -15,52 +15,49 @@ StringColumn::StringColumn(StringColumnMetadata* metadata)
...
          if (!result.has_value()) {
             return result;
          }
       } else {
-         null_bitmap.add(fixed_string_data.numValues());
-         fixed_string_data.insert(SiloString(""));
+         null_bitmap.add(row_id);
+         chunk.insertNull();
       }
    }
```
fengelniederhammer, line 155: question: On `registerPhyloNode` failure mid-buffer, the method returns early and the chunk is dropped (good — column stays unmodified). However, `null_bitmap.add(row_id)` calls from earlier iterations in this same buffer are NOT rolled back, leaving orphan bits at positions beyond `numValues()`. Are those harmless? `isNull` checks `null_bitmap.contains(row_id)` without bounds-checking against `numValues()`, so the bits are invisible in normal operation — but it's worth confirming that no code path iterates the bitmap directly and could be confused by them.

src/silo/storage/column/string_column.h
taepper: I already prepared a follow-up PR that will remove almost all the offset-index math

src/silo/storage/column/indexed_string_column.cpp
```diff
@@ -57,16 +57,35 @@ std::optional<const roaring::Roaring*> IndexedStringColumn::filter(
...
+   // Validate whole buffer before mutating anything
+   if (lineage_index.has_value() && !metadata->treat_unknown_lineages_as_null) {
+      for (const auto& maybe_value : buffer) {
+         if (maybe_value.has_value() && !metadata->dictionary.getId(*maybe_value).has_value()) {
+            return std::unexpected(fmt::format(
+               "The value '{}' is not a valid lineage value for column '{}'. "
+               "Is your lineage definition file outdated?",
+               *maybe_value,
+               metadata->column_name
+            ));
```
fengelniederhammer, line 69: Can we add a test for that case?
* add a chunk with lineages
* add a chunk with lineages that fails
* check that the last chunk didn't mutate anything

(Don't know how feasible that is)

src/silo/storage/column/string_column.test.cpp
```diff
@@ -123,6 +123,55 @@ TEST(StringColumn, rawInsertedValuesWithPhyloTreeRequeried) {
...
+
+   // A buffer that re-uses CHILD2 (already committed) with a null in between must fail, and must
+   // not leave the null in null_bitmap or grow the column.
+   StringColumn::Builder builder;
+   builder.insert("CHILD3");
+   builder.insertNull();
+   builder.insert("CHILD2");
+   auto result = under_test.appendChunk(builder.finalize());
+
+   ASSERT_FALSE(result.has_value());
```
fengelniederhammer, line 145: Isn't this supposed to have a failure then?

taepper: It reads `ASSERT_FALSE`

src/silo/storage/column/string_column.cpp
```diff
@@ -1,71 +1,183 @@
...
+   return chunk_idx == 0 ? 0 : chunk_end_offsets[chunk_idx - 1];
+}
+
+std::pair<size_t, size_t> StringColumn::locate(size_t row_id) const {
+   // The chunk containing `row_id` is the first whose (exclusive) end offset exceeds it.
+   const auto chunk_idx = static_cast<size_t>(
+      std::ranges::upper_bound(chunk_end_offsets, row_id) - chunk_end_offsets.begin()
+   );
+   return {chunk_idx, row_id - chunkStart(chunk_idx)};
+}
```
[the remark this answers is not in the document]
taepper: This function will be removed in #1325

src/silo/storage/column/string_column.test.cpp
taepper: I added an assertion for the error string

fengelniederhammer (approved):

PR #1319, opened 2026-06-22
feat(silo): add an at expression

### LINKED ISSUE #1292: Add `at` method to ScalarExpressions
`at` should be a scalar method taking a string as an input and outputting a string containing just the character at the respective position.

This will allow the following expressions:

`table.map({x := sequence.at(123)})`

taepper:
resolves #1292

### Summary

This adds an `at` expression that works on strings (1-indexed) and returns the character at a given position in the string. This can be used to get a character symbol from a sequence

## PR Checklist

### DIFF
--- a/src/silo/query_engine/operators/map_node.test.cpp
+++ b/src/silo/query_engine/operators/map_node.test.cpp
@@ -7,17 +7,34 @@ using silo::ReferenceGenomes;
 using silo::test::QueryTestData;
 using silo::test::QueryTestScenario;
 
-nlohmann::json createData(const std::string& primaryKey, int value) {
+nlohmann::json createData(
+   const std::string& primaryKey,
+   int value,
+   const std::string& str_value,
+   const nlohmann::json& segment1 = nullptr
+) {
    return {
       {"primaryKey", primaryKey},
       {"int_value", value},
-      {"segment1", nullptr},
+      {"str_value", str_value},
+      {"segment1", segment1},
       {"gene1", nullptr},
       {"unaligned_segment1", nullptr}
    };
 }
 
-const std::vector<nlohmann::json> DATA = {createData("id_0", 1), createData("id_1", 2)};
+nlohmann::json alignedSequence(const std::string& sequence) {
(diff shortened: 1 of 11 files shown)


src/silo/query_engine/operators/map_node.test.cpp
```diff
@@ -80,6 +93,34 @@ const QueryTestScenario MAP_FIELD_REF_SCENARIO = {
...
 };
 
+// `at` extracts the (1-indexed) character of a string column at a position.
+const QueryTestScenario MAP_AT_SCENARIO = {
+   .name = "MAP_AT",
+   .query = "default.map({second := primaryKey.at(4)}).project({primaryKey, second})",
+   .expected_query_result = nlohmann::json(
+      {{{"primaryKey", "id_0"}, {"second", "0"}}, {{"primaryKey", "id_1"}, {"second", "1"}}}
+   )
+};
```
fengelniederhammer, line 109: Let's add a test for what happens when one of the values isn't long enough?

Something like
* values of string column: ["short", "longlonglong"]
* `map({ mapped := stringColumn.at(8) })`

src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -412,16 +446,49 @@ FilterPtr handleInsertionContains(const BoundArguments& args) {
...
+   // Resolve the referenced column against the available schema so the At carries the
+   // column's actual {name, type}. If it is unknown the column is kept with a
+   // placeholder type; the caller reports the (contextual) unknown-column error.
+   const auto found =
+      std::ranges::find_if(schema, [&](const auto& col) { return col.name == input_column; });
+   schema::ColumnIdentifier column =
+      found != schema.end() ? *found
+                            : schema::ColumnIdentifier{
+                                 .name = std::move(input_column), .type = schema::ColumnType::STRING
+                              };
```
fengelniederhammer: Do we need this default? Or should we rather assert that we never reach it?

fengelniederhammer (approved):

PR #1320, opened 2026-06-22
build(silo): add WebAssembly build

chaoran-chen:
### Summary

Adds an WebAssembly (Emscripten) build of SILO so preprocessing and SaneQL queries can run in the browser or Node, alongside the unchanged native build.

- **New `wasm/` build target** — separate CMake entry point and Conan recipe that reuse the SILO core sources (excluding the CLI, HTTP API, append mode, tests, and benchmarks) and emit an embind module `silo_wasm.js`/`.wasm`.
- **Browser API** (`wasm/src/silo_wasm.cpp`) — `preprocess`, `save`, `load`, `query`, `info`, `dispose`, exchanging files through Emscripten's virtual filesystem. Uses a smaller query materialization cutoff than native, since browser memory is limited.
- **Shared CMake** (`cmake/silo_core.cmake`) — C++ standard, common compile definitions, core dependencies, and source globbing are factored out so the native and wasm builds cannot drift.
- **Build/test entry points** — `make wasm` and `make wasm-test` (Node smoke test against the `unitTestDummyDataset` fixture), a CI job that builds/tests/uploads artifacts, plus a browser example under `wasm/example/`.
- **Portability fixes in existing sources** — see the per-file notes below. Behavior of the native build is unchanged; the differences are guarded by `__EMSCRIPTEN__` / `SILO_WASM`.

The wasm build supports raw and `.zst` inputs. `.xz` is rejected.

## PR Checklist

---

<details>
<summary>AI sumary of the changes</summary>

### `LAPIS-SILO/wasm/example/`

Adds a minimal browser example with one HTML file, one JavaScript file, and one CSS file. It demonstrates the low-level browser integration without React:

- writing uploaded files into Emscripten&#39;s virtual filesystem;
- calling `preprocess(...)`;
- calling `save(...)` and downloading the saved SILO state as a dependency-free JSON archive;
- uploading that JSON archive, reconstructing the virtual filesystem state, and calling `load(...)`;
- calling `query(...)` and showing the raw NDJSON result.

### `LAPIS-SILO/wasm/CMakeLists.txt`

Adds a separate CMake entry point for the browser build. It reuses SILO source files but excludes the CLI entry point, HTTP API, append mode, tests, and benchmarking code. The target links the SILO core plus the dependencies needed for preprocessing and SaneQL query execution, then emits an embind module with filesystem and pthread support.

This keeps the WASM experiment isolated from SILO&#39;s normal native CMake build.

### `LAPIS-SILO/wasm/conanfile.py`

Adds a WASM-specific Conan recipe. It keeps only the dependencies and features needed by the browser target, disables unnecessary Arrow and Boost features, and requests static libraries. This reduces the amount of code compiled by Emscripten and avoids native-only dependency features.

### `LAPIS-SILO/Makefile`

Adds the `wasm` target. It follows the existing native build pattern: create a Conan profile, install dependencies, configure CMake, build the target, then copy `silo_wasm.js` and `silo_wasm.wasm` to `LAPIS-SILO/dist/wasm/`.

### `LAPIS-SILO/buildScripts/create-wasm-conanprofile`

Adds the Emscripten equivalent of `create-conanprofile`. It assumes `emcc` is installed and on `PATH`, reads its version with `emcc -dumpversion`, and writes `build/wasm/conanprofile-emscripten`.

### `LAPIS-SILO/buildScripts/install-wasm-dependencies`

Adds the Emscripten equivalent of `install-dependencies`. It uses the existing `uv run --project buildScripts conan ...` pattern, installs the reduced WASM dependency set from `wasm/conanfile.py`, and writes Conan outputs to `build/wasm`.

### `LAPIS-SILO/wasm/src/silo_wasm.cpp`

Adds the browser-facing embind API:

- `preprocess(preprocessingConfigPath)` reads SILO preprocessing config from the Emscripten filesystem and returns a database handle.
- `save(handle, outputDirectory)` saves an in-memory database to SILO&#39;s normal processed-state files in the Emscripten filesystem.
- `load(stateDirectory)` loads a saved SILO state from the Emscripten filesystem and returns a database handle.
- `query(handle, saneqlQuery)` plans and executes SaneQL against an in-memory SILO database and returns NDJSON.
- `info(handle)` returns database info as JSON.
- `dispose(handle)` releases a loaded database.

The WASM binding uses a smaller query materialization cutoff than native SILO. Browser memory is limited, and detail queries that project full nucleotide sequences can otherwise materialize large intermediate batches before a downstream `.limit(...)` is applied.

### `LAPIS-SILO/src/silo/common/input_stream_wrapper.cpp`

Guards the `.xz`/lzma input path with `#ifndef __EMSCRIPTEN__`. The browser fixture path uses raw and `.zst` inputs; lzma support is not strictly needed and pulls in a native library that complicates the Emscripten build. On Emscripten the `.xz`-ending branch now fails fast with a clear `PreprocessingException` instead of letting an `.xz` file fall through to the raw-file path and be read as if uncompressed.

### `LAPIS-SILO/src/silo/query_engine/saneql/lexer.cpp`

Changes `from_chars` end-pointer checks from `ptr != num_str.end()` to `ptr != num_str.data() + num_str.size()`. `fast_float::from_chars` returns a raw pointer, while `std::string::end()` is an iterator. Some native toolchains accept the original comparison, but the Emscripten/libc++ build rejects it.

### `LAPIS-SILO/src/silo/query_engine/exec_node/table_scan.h`

Uses synchronous table-scan batch production only when compiling the WASM target. Native SILO keeps the existing detached-thread path, while the browser build avoids creating an extra pthread per query batch, which can starve or deadlock the limited Emscripten pthread worker pool on larger browser datasets.

### `LAPIS-SILO/src/silo/storage/buffer/page.h`

Renames the internal page-size constant from `PAGE_SIZE` to `SILO_PAGE_SIZE`. Emscripten/system headers can expose `PAGE_SIZE` as a macro, so the old name conflicts in the browser build. The deserialization allocation now uses `new[]`, matching the existing `delete[]` destructor.

### `LAPIS-SILO/src/silo/storage/vector/variable_data_registry.cpp`

Updates references to the renamed page-size constant, from `buffer::PAGE_SIZE` to `buffer::SILO_PAGE_SIZE`.

### `LAPIS-SILO/src/silo/storage/column/vertical_sequence_index.h`

Adjusts the `SequenceDiff` size assertion for wasm32. On 64-bit native builds the struct remains 16 bytes; on wasm32 pointer-size/layout differences make it 12 bytes. The assertion now documents both expected layouts.

</details>


wasm/src/silo_wasm.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+constexpr uint64_t QUERY_TIMEOUT_SECONDS = 120;
+// Native SILO defaults to DEFAULT_ARROW_BATCH_SIZE (32767, see
+// src/silo/config/runtime_config.cpp) rows before it stops collecting a query
+// result in memory and starts streaming it instead. Browser memory is far
+// more constrained than a server process, and detail queries that project
+// full nucleotide sequences can materialize large intermediate batches before
+// a downstream `.limit(...)` is applied, so the browser build uses a much
+// smaller cutoff.
+constexpr size_t DEFAULT_MATERIALIZATION_CUTOFF = 256;
```
taepper: Let's not call this `DEFAULT_MATERIALIZATION_CUTOFF` as it is used to overwrite the default in the function `browserQueryOptions()`. `WASM_MATERIALIZATION_CUTOFF` sounds more fitting to me

src/silo/query_engine/saneql/lexer.cpp
```diff
@@ -131,15 +131,15 @@ Token Lexer::readNumber() {
...
-      if (ec != std::errc() || ptr != num_str.end()) {
+      if (ec != std::errc() || ptr != num_str.data() + num_str.size()) {
          throw ParseException(start, "Invalid float literal");
       }
       return makeToken(TokenType::FLOAT_LITERAL, val, start);
    }
 
    uint64_t val = 0;
    auto [ptr, ec] = fast_float::from_chars(num_str.data(), num_str.data() + num_str.size(), val);
-   if (ec != std::errc() || ptr != num_str.end()) {
```
taepper: Can we put these changes into a separate PR?

src/silo/storage/buffer/page.h
```diff
@@ -1,19 +1,20 @@
 #pragma once
 
+#include <cstddef>
 #include <cstdint>
 
 #include <boost/serialization/binary_object.hpp>
 
 namespace silo::storage::buffer {
 
-const size_t PAGE_SIZE = 16384;
```
taepper: Can we put these changes into a separate PR?

src/silo/common/input_stream_wrapper.cpp
```diff
@@ -32,9 +35,18 @@ InputStreamWrapper::InputStreamWrapper(const std::filesystem::path& file_path) {
       file_stream = std::ifstream(withZSTending(file_path), std::ios::binary);
       boost_input_stream->push(boost::iostreams::zstd_decompressor());
    } else if (std::filesystem::is_regular_file(withXZending(file_path))) {
+#ifndef __EMSCRIPTEN__
```
taepper, line 38: Here we guard with `__EMSCRIPTEN__` but below with `SILO_WASM`. Which do we want?

wasm/conanfile.py
taepper, line 1: We repeat the full `conanfile.py` with just `"boost/*:lzma": False` and Poco removed. Is there not a cleaner way?

src/silo/storage/buffer/page.h
```diff
@@ -42,7 +43,7 @@ class Page {
       } else {
          // Loading: ensure buffer is allocated, then read from archive
          if(!buffer) {
-            buffer = reinterpret_cast<uint8_t*>(malloc(SILO_PAGE_SIZE));
+            buffer = new uint8_t[SILO_PAGE_SIZE];
```
fengelniederhammer, line 46: @taepper AI claimed that it fixed an undefined behavior bug there. What do you think about it?

fengelniederhammer: The bug: `Page` allocates `buffer` with `new uint8_t[SILO_PAGE_SIZE]` in the constructor and frees it with `delete[]` in the destructor. But the deserialization path allocated with `malloc(...)`:

```cpp
Page() { buffer = new uint8_t[SILO_PAGE_SIZE]; }   // ctor: new[]
~Page() { delete[] buffer; }                        // dtor: delete[]
...
buffer = reinterpret_cast<uint8_t*>(malloc(SILO_PAGE_SIZE));  // load path: malloc
```

When a `Page` is loaded from an archive with `buffer == nullptr`, it gets a `malloc`'d pointer that is later freed with `delete[]`. Mixing allocation families (`malloc`/`free` vs `new[]`/`delete[]`) is undefined behavior — the two need not use the same underlying bookkeeping, so this can corrupt the heap.

The fix just makes the load path use `new uint8_t[SILO_PAGE_SIZE]` too, so all three sites (ctor, load, dtor) are consistent `new[]`/`delete[]`.

Note this is unrelated to the WASM build — it's a pre-existing correctness issue. The rest of the `page.h` changes (the `SILO_PAGE_SIZE` rename + `<cstddef>` include) already landed on `main` via another PR, so after merging main this one-line fix is all that remains here.

taepper: Oh, that is correct, introduced in `7359eeaf`

wasm/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+# Shared C++ standard, common compile definitions, core dependencies and source
+# collection (shared with the native build).
+include(${CMAKE_CURRENT_LIST_DIR}/../cmake/silo_core.cmake)
+
+add_compile_definitions(SILO_DEBUG_ASSERTIONS=0)
+add_compile_definitions(NO_EVOBENCH)
+
+# Collect the SILO core sources, excluding the CLI entry point, HTTP API, append
+# mode, and evobench, which are not part of the browser target.
+silo_collect_sources(SILO_CORE_SOURCES
```
taepper: I think we should straighten the problems leading to these specific excludes instead and structure it properly. I do not think the cmake division is sufficient for now and I am working to alleviate it

taepper:
Importantly, [‎cmake/silo_core.cmake‎](https://github.com/GenSpectrum/LAPIS-SILO/pull/1320/changes#diff-98a9041dab325524049f1d55d6343bf4e816950d4e8850e578b66bd6deb842d5) should disappear after the merge and it should only use the `CMakeLists.txt` and `wasm/CMakeLists.txt`

CMakeLists.txt
```diff
@@ -68,10 +82,32 @@ target_compile_definitions(silolib_deps INTERFACE
...
+if(EMSCRIPTEN)
+    target_compile_definitions(silolib_deps INTERFACE SILO_WASM NO_EVOBENCH)
+    target_compile_options(silolib_deps INTERFACE -fexceptions -pthread)
+endif()
+
 file(GLOB_RECURSE SILOLIB_SOURCES CONFIGURE_DEPENDS "src/*.cpp")
 file(GLOB_RECURSE SILO_TEST_SOURCES CONFIGURE_DEPENDS "src/*.test.cpp")
 list(REMOVE_ITEM SILOLIB_SOURCES ${SILO_TEST_SOURCES})
 
+# The browser target does not include append mode, runtime config, or evobench.
```
taepper: I think this should still work if we leave these files in (it did for me when working on #1362 with minor edits). Otherwise, let it fix the outstanding errors here

src/silo/common/input_stream_wrapper.cpp
```diff
@@ -6,12 +6,15 @@
 
 #include <spdlog/spdlog.h>
 #include <boost/iostreams/detail/error.hpp>
-#include <boost/iostreams/filter/lzma.hpp>
 #include <boost/iostreams/filter/zstd.hpp>
 #include <boost/iostreams/filtering_stream.hpp>
 
 #include "silo/preprocessing/preprocessing_exception.h"
 
+#ifndef SILO_WASM
```
taepper: Let's use `__EMSCRIPTEN__` as a guard afterall as it removes the special declaration in `CMakeLists.txt`

CMakeLists.txt
```diff
@@ -68,10 +82,32 @@ target_compile_definitions(silolib_deps INTERFACE
...
 )
 
+# WebAssembly-only usage requirements. These are INTERFACE so they compile into
+# the silolib objects too: the SILO_WASM guards live in silolib sources
+# (input_stream_wrapper.cpp, exec_node/table_scan.h), NO_EVOBENCH neutralises the
+# evobench header pulled in by silolib, and -fexceptions must reach every silolib
+# object (they throw). -pthread keeps SILO objects ABI-consistent with the
+# Emscripten dependency builds.
+if(EMSCRIPTEN)
+    target_compile_definitions(silolib_deps INTERFACE SILO_WASM NO_EVOBENCH)
```
taepper: Let's use `__EMSCRIPTEN__` as a guard afterall as it removes the special `SILO_WASM` declaration in `CMakeLists.txt`

CMakeLists.txt
```diff
@@ -109,29 +145,38 @@ endif ()
 # Tests
 # ---------------------------------------------------------------------------
 
-enable_testing()
-find_package(GTest REQUIRED)
-include_directories(${GTest_INCLUDE_DIRS})
+# GTest is not part of the Emscripten dependency set; the browser build ships
+# its own node smoke test under wasm/test instead.
+if(NOT EMSCRIPTEN)
```
taepper: Let's make this opt-in instead

Default the option's value to `ON`

(also maybe remove the second sentence of the comment)

taepper (approved):
LGTM!

PR #1324, opened 2026-06-23
refactor: extract generic pipeline optimizer base class

### LINKED ISSUE #1282: Generic Pipeline Optimizer propagation
We should add a parent class to NodeResolutionPass and FilterPushdownPass which implements the propagation for any pipelined operator (e.g. fetch, map, project, zstddecompress). This removes boilerplate and ensures that optimizations are always properly propagated

fengelniederhammer:
resolves #1282

### Summary

`NodeResolutionPass` + `FilterPushdownPass` had identical boilerplate `operator()` per pipeline operator (Fetch, Map, Project, ZstdDecompress, OrderBy, Filter, `Unresolved*`, `UnionAll`) — each just recurse into child. New operator = edit every pass, easy to forget.

Extract CRTP base `PipelinePassBase` (`pipeline_pass_base.h`): default propagation for all single-child operators + `UnionAll`, shared `run()` + `propagateToNode()`. Each pass inherits, pulls defaults via `using PipelinePassBase::operator();`, overrides only what it needs. `ColumnNarrowingPass` is migrated onto the base too (overriding `makePass()` to seed its required-column state from the root output schema).

### Traversal

`propagateToNode()` visits a node once, replaces it if the handler returns non-null. Handlers recurse into their own children before returning, so a single visit per node is enough — no fix-point loop needed.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/column_narrowing_pass.cpp
+++ b/src/silo/query_engine/column_narrowing_pass.cpp
@@ -19,9 +19,9 @@ namespace silo::query_engine {
 
 namespace {
 // NOLINTNEXTLINE(misc-no-recursion)
-void applyToNode(operators::QueryNodePtr& child, ColumnNarrowingPass& pass) {
-   if (auto new_child = operators::visit(*child, pass)) {
-      child = std::move(new_child);
+void applyToNode(operators::QueryNodePtr& node, ColumnNarrowingPass& pass) {
+   if (auto new_node = operators::visit(*node, pass)) {
+      node = std::move(new_node);
    }
 }
 }  // namespace
--- a/src/silo/query_engine/column_narrowing_pass.cpp
+++ b/src/silo/query_engine/column_narrowing_pass.cpp
@@ -1,35 +1,17 @@
 #include "silo/query_engine/column_narrowing_pass.h"
 
-#include "silo/query_engine/operator_visitor.h"
 #include "silo/query_engine/operators/aggregate_node.h"
-#include "silo/query_engine/operators/fetch_node.h"
-#include "silo/query_engine/operators/filter_node.h"
 #include "silo/query_engine/operators/map_node.h"
 #include "silo/query_engine/operators/order_by_node.h"
 #include "silo/query_engine/operators/project_node.h"
 #include "silo/query_engine/operators/table_scan_node.h"
 #include "silo/query_engine/operators/union_all_node.h"
-#include "silo/query_engine/operators/unresolved_insertions_node.h"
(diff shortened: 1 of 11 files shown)


src/silo/query_engine/column_narrowing_pass.cpp
```diff
@@ -19,13 +19,19 @@ namespace silo::query_engine {
 
 namespace {
 // NOLINTNEXTLINE(misc-no-recursion)
-void applyToChild(operators::QueryNodePtr& child, ColumnNarrowingPass& pass) {
+void applyToNode(operators::QueryNodePtr& child, ColumnNarrowingPass& pass) {
```
taepper: Maybe let's also call the argument `node`

src/silo/query_engine/pipeline_pass_base.h
```diff
@@ -0,0 +1,146 @@
...
+      return nullptr;
+   }
+
+  protected:
+   /// Repeatedly visit `child` with this pass, replacing it each time a non-null
+   /// replacement is returned, until the pass returns nullptr.
+   // NOLINTNEXTLINE(misc-no-recursion)
+   void propagateToNode(operators::QueryNodePtr& child) {
+      auto& derived = static_cast<Derived&>(*this);
+      while (auto replacement = operators::visit(*child, derived)) {
```
taepper: Let's maybe add a limit here and log using `SPDLOG_WARN` that cyclic optimization behavior was found.

Also logging the changed plans might be very useful

taepper (approved):
Looks good, some minor comments

fengelniederhammer:
I removed the loop in the passes now and replaced it by a single if. I think it should be fine, at least the tests that I added pass: The passes already traversed the whole node tree, i.e. it shouldn't be necessary to visit any node more than once.

src/silo/query_engine/pipeline_pass_base.h
```diff
@@ -0,0 +1,151 @@
...
+   /// Visit `node` with this pass once, replacing it if the pass returns a non-null
+   /// replacement. Handlers are responsible for recursing into their own children
+   /// before returning a replacement, so a single visit per node is sufficient.
+   // NOLINTNEXTLINE(misc-no-recursion)
+   void propagateToNode(operators::QueryNodePtr& node) {
+      auto& derived = static_cast<Derived&>(*this);
+      if (auto replacement = operators::visit(*node, derived)) {
+         node = std::move(replacement);
+      }
+   }
```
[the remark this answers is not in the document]
fengelniederhammer: outdated

taepper (approved):
LGTM!

PR #1325, opened 2026-06-23
1296 chunked row ids

### LINKED ISSUE #1296: Change RowIds to explicitly refer to their chunk
Currently, rows in the database can be refered to by a single integer. With a chunked storage layout, we will want to split the lower 16 bits out of this id. By storing the upper bits (referring to the chunk) and lower bits (offset within a chunk), we will not need to track chunk boundaries as explicitly anymore.

taepper:
resolves #1296

### Summary

Previously RowIds (e.g. used for fetching data of a specific row) were continuous. We want to subdivide the space to reflect our chunked storage architecture (also reflected in roaring container chunking)

## PR Checklist


src/silo/query_engine/filter/operators/range_selection.cpp
```diff
@@ -44,25 +53,50 @@ Type RangeSelection::type() const {
 CopyOnWriteBitmap RangeSelection::evaluate() const {
    EVOBENCH_SCOPE("RangeSelection", "evaluate");
    roaring::Roaring result_bitmap;
-   for (const auto& range : ranges) {
-      result_bitmap.addRange(range.start, range.end);
+   for (const auto& [start, end] : ranges) {
+      SILO_ASSERT(
+         end.chunk_id == row_layout.numChunks() ||
+         end.row_in_chunk < row_layout.chunkSize(end.chunk_id)
+      );
```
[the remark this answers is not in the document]
taepper: These range-ends should be represented as first element of next chunk, to make comparisons consistent

src/silo/storage/column/row_layout.h
```diff
@@ -56,31 +84,50 @@ class RowLayout {
...
+         skipEmptyChunks();
+      }
 
-      RowId operator*() const { return row_id; }
+      RowId operator*() const { return RowId{.chunk_id = chunk_id, .row_in_chunk = row_in_chunk}; }
 
       Iterator& operator++() {
-         ++row_id;
+         SILO_ASSERT_LT(chunk_id, chunk_sizes->size());
+         if (row_in_chunk == UINT16_MAX || row_in_chunk == chunk_sizes->at(chunk_id) - 1) {
```
fengelniederhammer, line 97: Missing test: the `row_in_chunk == UINT16_MAX` wrap branch is never exercised. All `RowLayout` iterator tests use small chunks, so iteration only hits the `== chunkSize - 1` path. A chunk of size `COLUMN_CHUNK_SIZE` (65536) is needed to reach `row_in_chunk == UINT16_MAX` here. Add a full-chunk iteration test.

src/silo/storage/column/horizontal_coverage_index.h
```diff
@@ -19,43 +20,56 @@ namespace silo::storage::column {
...
-   void insertCoverage(const Coverage& coverage);
+   void insertCoverage(RowId row_id, const Coverage& coverage);
+
+   void insertNullSequence(RowId row_id);
 
-   void insertNullSequence();
+   [[nodiscard]] size_t numChunks() const { return start_end.size(); }
+
+   [[nodiscard]] uint32_t chunkSize(uint16_t chunk_id) const {
+      return static_cast<uint16_t>(start_end.at(chunk_id).size());
```
fengelniederhammer: AI: Missing test: `chunkSize` is never exercised on a full `COLUMN_CHUNK_SIZE` (65536) chunk. A full-chunk test here would assert the size round-trips correctly through this conversion.

src/silo/query_engine/filter/operators/range_selection.cpp
```diff
@@ -44,25 +54,55 @@ Type RangeSelection::type() const {
...
+            auto chunk_start = RowId::chunkStart(chunk_id);
+            auto chunk_end = RowId::chunkStart(chunk_id) + row_layout.chunkSize(chunk_id);
+            result_bitmap.addRange(chunk_start, chunk_end);
+         }
+
+         auto start_of_end_chunk = RowId{.chunk_id = end.chunk_id, .row_in_chunk = 0};
+         if (end != start_of_end_chunk) {
+            result_bitmap.addRange(start_of_end_chunk.toGlobal(), end.toGlobal());
+         }
+      }
```
fengelniederhammer, line 84: AI: Missing test: the cross-chunk branch (partial-first / full-middle / partial-last range expansion) is never exercised. All `range_selection.test.cpp` cases build ranges with `RowId::fromGlobal` on small ids, so `start.chunk_id == end.chunk_id` always holds and only the single-chunk path on L62-63 runs. Add a test with a multi-chunk `RowLayout` and a range spanning chunk boundaries.

fengelniederhammer (approved):

PR #1327, opened 2026-06-24
refactor(silo): explicitly model row layout of a table

taepper:
Small refactoring commit that reduces the diff of #1325.

This explicitly models the row layout of a table. As of now, this seems a little unnecessary because it is just implied by the row count, but this dense format will be remodelled in #1325


taepper:
I would advocate for ignoring the linter errors, as they will disappear once `RowLayout` is no longer a trivial type (which happens in #1325)

src/silo/storage/column/row_layout.h
```diff
@@ -0,0 +1,87 @@
+#pragma once
+
+#include <algorithm>
+#include <cstdint>
+#include <iterator>
+#include <vector>
+
+#include <boost/serialization/access.hpp>
+#include <boost/serialization/vector.hpp>
```
fengelniederhammer, line 9: AI: nit: these includes look unused in this header — `<algorithm>`, `<iterator>` (no algorithm/iterator usage; `iterator_category`/`forward_iterator_tag` come from `<iterator>` actually — keep that one), `<vector>` and `boost/serialization/vector.hpp` (no vector member; `num_rows` is scalar). Drop the genuinely unused ones.

fengelniederhammer (approved):
Good to have this as a separate PR ;)

src/silo/storage/column/row_layout.h
fengelniederhammer: Ok, I think AI got a little confused here, but my IDE also shows unused includes ;)

taepper: Will do in follow-up PR #1325

PR #1333, opened 2026-06-29
feat(SaneQL): add `schema` operator to get query result schema

### LINKED ISSUE #1331: Query result schema: implement basic version
As a first step of #1312, let's implement the simplest version to get a feeling for how to implement more complicated things. Let's start with this: https://github.com/GenSpectrum/LAPIS-SILO/issues/1312#issuecomment-4796768193

Keep in mind: can we distinguish sequence vs non-sequence strings at this stage?

fengelniederhammer:
resolves #1331

### Summary

Adds a `schema()` SaneQL operator that returns the output schema of a query (or any pipeline result) as rows, without executing the underlying query plan:

```
default.filter(country='CH').groupBy({count:=count()}, {age}).schema()
// → {"fieldName":"age","type":"INT32"}
//   {"fieldName":"count","type":"INT64"}
```

Each row has `fieldName` (STRING) and `type` (STRING). The child plan is never executed — only its declared output schema is inspected.

**Limitations (documented):**
- Sequence columns report as `STRING` (decompressed to string before `schema()` observes them).
- `filter()` after `schema()` is rejected at planning time — there is no data source above `schema()` to push a predicate into.

**Implementation notes for reviewers:**
- `SchemaNode` is a source-type leaf; it hits `FilterPushdownPass`, `NodeResolutionPass`, and `ColumnNarrowingPass` with per-pass overrides so the child subtree is validated (bad sequence names, mutations-not-on-scan, etc.) but never executed.
- `ColumnNarrowingPass` recurses into the child seeded with the child's *full* output schema (no pruning), preserving the structural Map/Project collapses that `NodeResolutionPass` depends on.

## PR Checklist

### DIFF
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -358,59 +358,33 @@ unionAll(
 
 ### `schema()`
 
-Reports the output schema of whatever it is applied to — a base table or any
-intermediate query result. It emits one row per output field with two columns:
-`fieldName` and `type`, where `type` is the field's column type as a string.
-
-Schema of a base table:
+Describes the output schema of whatever it is applied to — a table or the result of any pipeline.
+It does not read or return any data; it only reports the fields that the input would produce.
 
 ```
 default.schema()
-```
-
-```
-{"fieldName":"primaryKey","type":"STRING"}
-{"fieldName":"country","type":"STRING"}
-{"fieldName":"date","type":"DATE32"}
-{"fieldName":"age","type":"INT32"}
-```
-
-Schema of an intermediate result (the same operator, after a pipeline). The
-aggregate `count` is `INT64`:
-
-```
 default.filter(country='CH').groupBy({count:=count()}, {age}).schema()
-```
-
-```
-{"fieldName":"count","type":"INT64"}
-{"fieldName":"age","type":"INT32"}
-```
-
-For `mutations()` the `count` field is `INT32`:
-
-```
 default.mutations(minProportion:=0.1).schema()
 ```
 
-```
-{"fieldName":"mutation","type":"STRING"}
-{"fieldName":"mutationFrom","type":"STRING"}
-{"fieldName":"mutationTo","type":"STRING"}
-{"fieldName":"sequenceName","type":"STRING"}
-{"fieldName":"position","type":"INT32"}
-{"fieldName":"proportion","type":"FLOAT"}
-{"fieldName":"coverage","type":"INT32"}
-{"fieldName":"count","type":"INT32"}
+**Output:** one row per field of the described result, with two columns:
+
+| Field       | Type   | Description                                  |
+|-------------|--------|----------------------------------------------|
+| `fieldName` | string | Name of the field                            |
+| `type`      | string | Type of the field (e.g. `STRING`, `INT32`, `INT64`, `DATE32`, `BOOL`, `FLOAT`, `INDEXED_STRING`) |
+
+```json
+{"fieldName": "age", "type": "INT32"}
+{"fieldName": "count", "type": "INT64"}
 ```
 
-**Limitation:** sequence columns are reported as `STRING`. When a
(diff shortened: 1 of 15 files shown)


documentation/query_documentation.md
```diff
@@ -356,6 +356,38 @@ unionAll(
...
+```
+
+`schema()` produces an ordinary two-column relation,
+so schema-preserving and schema-defining operators (such as `project` and `orderBy`) can be chained after it.
+
+**Limitation:** sequence columns are reported with type `STRING`.
+When a sequence column is read into a pipeline it is decompressed to a string before `schema()` observes it,
+so nucleotide and amino acid sequences cannot be distinguished from ordinary strings at this point.
+
+**Limitation:** `schema()` is a result-producing source, so `filter(...)` cannot be applied to its output.
```
taepper: This limitation is weirdly worded. What does "result-producing source" mean? I would leave it as a limitation that 
> `filter(...)` cannot be applied to schema()

This should also be tested, that the optimizer does not "pass through" this node by accident.

A term we can (and probably should) introduce here is "pipeline breaker". I think these generally behave the same with regards to the optimizer. Current pipeline breakers are `groupBy`, `mutations`, `insertions`, `schema`. `orderBy` is a odd-one out, as it is a pipeline breaker that does allow push-through of filter (and probably many other) operators

src/silo/query_engine/operators/schema_node.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <arrow/result.h>
+
+#include "silo/query_engine/operators/query_node.h"
+#include "silo/schema/database_schema.h"
+#include "silo/storage/table.h"
+
+namespace silo::query_engine::operators {
+
+/// Terminal node that reports the output schema of its child as data rows.
```
taepper: This need not be terminal. `limit` and `orderBy` should work on this node's result out-of-the box. `filter` currently can only applied to tables directly. But should be changed in the near future.

src/silo/query_engine/operators/schema_node.test.cpp
```diff
@@ -0,0 +1,270 @@
...
+   .query = "default.map({tag := 42, label := 'x'}).project({primaryKey, tag, label}).schema()",
+   .expected_query_result = nlohmann::json(
+      {{{"fieldName", "primaryKey"}, {"type", "STRING"}},
+       {{"fieldName", "tag"}, {"type", "INT64"}},
+       {{"fieldName", "label"}, {"type", "STRING"}}}
+   )
+};
+
+// project() before schema() controls field selection and order
+const QueryTestScenario SCHEMA_AFTER_PROJECT_ORDER_SCENARIO = {
```
taepper, line 142: Ah, I would have forgotten about the (very good!) requirement that the result should be ordered by field order of the input node! Nice that this is covered by tests

src/silo/query_engine/optimizer/pipeline_pass_base.h
```diff
@@ -124,6 +125,19 @@ class PipelinePassBase {
       return nullptr;
    }
 
+   // Default for `SchemaNode`: do NOT propagate into its child. `schema()` never executes
+   // the child plan; it only inspects `child->getOutputSchema()`. Pure optimization passes
+   // (e.g. ColumnNarrowingPass) must not touch the child: narrowing in particular would
```
taepper: But the `ColumnNarrowingPass` specifically does recurse into its child nodes (but requiring all columns). Is this default necessary?

taepper: (`NodeResolution` and `FilterPushdown` overwrite it also)

taepper (commented):
Looks good, a final consideration

src/silo/query_engine/optimizer/pipeline_pass_base.h
fengelniederhammer: True, this is slop.

taepper (approved):

Issue #1336, opened 2026-06-29
Aggressive Map Pull-up through Project, OrderBy and Aggregate


fengelniederhammer:
Sub-issue of #1258. Follow-up to #1335 (depends on it). Mostly AI generated implementation plan.

Extend `MapPullupPass` (introduced in #1335) to also pull a `MapNode` up through:

- **`ProjectNode`**
- **`OrderByNode`**
- **`AggregateNode`**

when the parent references **none** of the columns the `MapNode` produces (same `requiredColumns(P) ∩ producedColumns(M) == ∅` rule).

## Why a separate issue

v1 (#1335) deliberately only pulls through `FetchNode` + `FilterNode`, which covers the main win (zstd decompress after `limit`) with minimal risk. The remaining operators need extra correctness handling and are scoped here to keep #1335 small.

## Correctness considerations

- **`ProjectNode`**: a Project may drop columns. Pulling a Map above a Project that drops one of the Map's produced columns would change the output schema. Only swap when the Project keeps all of (or does not depend on) the produced columns, and ensure the resulting output schema is unchanged. Note `MapNode::getOutputSchema` appends newly-added columns in assignment order — preserve column ordering.
- **`OrderByNode`**: safe only when the order-by fields don't reference produced columns; ordering is otherwise preserved.
- **`AggregateNode`**: aggregation collapses rows — only valid when neither `group_by_fields` nor `aggregates[].source_column` reference produced columns, and the Map's per-row computation is still meaningful above the aggregate.

## Tests

- pull through `Project` / `OrderBy` / `Aggregate` when produced columns are unused.
- no swap when the parent references a produced column.
- output-schema and ordering correctness, especially across `Project` dropping columns.

fengelniederhammer:
Scope update: `FilterNode` pull-through was originally planned for #1335 but moved here.

Pulling a `MapNode` up through a `FilterNode` requires knowing which columns the filter predicate references, via `Expression::freeIUs()`. Currently only `FieldRef`, `At` and `ZstdDecompressScalar` implement `freeIUs()`; all filter predicate types (`StringEquals`, `IntEquals`, `And`, `Or`, `Negation`, ...) inherit the base default `{}`. With that gap, a swap decision based on `freeIUs()` would incorrectly pull a `MapNode` above a filter that depends on a produced column, breaking the plan.

So this issue should additionally cover:

- **`FilterNode`** pull-through, gated on `freeIUs()` being complete.
[...]

PR #1338, opened 2026-06-29
refactor(SaneQL): map pull up: pull map expressions through fetch nodes

### LINKED ISSUE #1335: Map Pull-up through Fetch and Filter (v1)
Sub-issue of #1258 - mostly AI generated implementation plan.

Add a new optimizer `Pass` (`MapPullupPass`) that moves a `MapNode` **up** the plan tree, so expensive per-row computation (especially zstd decompression, now a `ZstdDecompressScalar` inside `MapNode` after #1311) runs on as few rows as possible.

## Motivation

Decompress/map assignments are inserted near the bottom of the tree, so they run before row-reducing nodes shrink the stream. E.g. a `limit` causes every row to be decompressed and then discarded. Pulling the `MapNode` above a `FetchNode` fixes this.

## Scope (v1)

Pull `MapNode` up through:
- **`FetchNode`** (limit/offset — the main win)
- **`FilterNode`** (needed as a pass-through hop so a Map can reach a `FetchNode` sitting above a filter)

Core transform for parent `P` over `MapNode M`:

```
P(M(child))  →  M(P(child))
```

applied only when `P` references **none** of the columns `M` produces:

- **producedColumns(M)** = `MapNode.assignments[].output_column` (by name).
- **requiredColumns(P)**: `FilterNode.filter->freeIUs()`; `FetchNode` requires none.
- **Swap allowed** iff `requiredColumns(P) ∩ producedColumns(M) == ∅`.
- Both `Fetch` and `Filter` preserve row count and ordering, and `MapNode` is row-count/order preserving, so the swap is semantics-preserving.

## Blocked by (leave Map in place, v1)

`ProjectNode`, `OrderByNode`, `AggregateNode`, `MapNode`, `UnionAllNode`, and the root. Aggressive pull-through of these is deferred to the follow-up (sub-issue B).

## Implementation notes

- New files `src/silo/query_engine/optimizer/map_pullup_pass.{h,cpp,test.cpp}`, namespace `silo::query_engine::optimizer` (passes live there since #1330; CMake auto-scans `src/`).
- `class MapPullupPass : public PipelinePassBase<MapPullupPass>` (CRTP base from #1282), stateless, `using PipelinePassBase<MapPullupPass>::operator();`. Override only `operator()(FetchNode&)` and `operator()(FilterNode&)`.
- Recurse child-first (`propagateToNode(node.child)`) so a Map below already bubbled to be the direct child, then attempt the swap. Drive repeated bubbling so a Map keeps rising through stacked Fetch/Filter (base `propagateToNode` visits a node only once and does not re-visit a swapped-in replacement).
- Register in `Planner::planQuery` (`planner.cpp`) after `ColumnNarrowingPass`, before `FilterPushdownPass`, with a matching `log_plan` line.

## Tests

- `fetch(map(scan))` → `map(fetch(scan))`.
- `filter(map(scan))`: independent filter → swap; filter referencing a produced column → no swap.
- chained `fetch(filter(map(scan)))` → Map bubbles to the top.
- blocked at `Project` / `OrderBy` / `Aggregate` / `UnionAll` / root.
- one `QUERY_TEST` E2E: query with a zstd sequence column + `limit`, verifying correctness and that decompression runs after the limit.

fengelniederhammer:
resolves #1335

### Summary

New optimizer pass `MapPullupPass` moves a `map` expression above an adjacent `limit`/`offset`
so it runs only on the rows kept, not every row. Rewrites `Fetch(Map(child)) → Map(Fetch(child))`;
the swap is always safe because `Fetch` references no columns and preserves ordering.

- **`MapPullupPass`** — pulls a `MapNode` up through a directly-adjacent `FetchNode`; blocked at every other node and at the root.
- **Wired into `Planner::planQuery`** after `FilterPushdownPass` (which can expose a `Fetch(Map(...))` shape) and before `NodeResolutionPass`.
- **Tests** — unit coverage of the swap, stacked-fetch bubbling, `Fetch(non-Map)` no-op, nested `Map(Fetch(Map))`, and all blocked nodes; one E2E (`MAP_WITH_LIMIT_TRIGGERS_PULLUP`) where a real `map({...}).limit()` query triggers the rewrite.

v1 does **not** yet move the zstd-decompression `MapNode` above a `limit`: it sits over the `TableScan` with Project/OrderBy between it and any `Fetch`. Pull-through of Filter/Project/OrderBy to reach it is deferred to #1336.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/AGENTS.md
+++ b/src/silo/query_engine/AGENTS.md
@@ -9,7 +9,7 @@ IMPORTANT: Changes here may require updates to query docs: `documentation/query_
 SaneQL string
   → Parser (saneql/parser.cpp) → AST (saneql/ast.h)
   → ast_to_query.cpp → QueryNode tree (operators/*.h)
-  → ColumnNarrowingPass → FilterPushdownPass → NodeResolutionPass
+  → ColumnNarrowingPass → MapPullupPass → FilterPushdownPass → NodeResolutionPass
   → addToExecPlan() → Arrow Acero ExecPlan (single shared plan)
   → QueryPlan execution → send result
 ```
--- a/src/silo/query_engine/AGENTS.md
+++ b/src/silo/query_engine/AGENTS.md
@@ -9,7 +9,7 @@ IMPORTANT: Changes here may require updates to query docs: `documentation/query_
 SaneQL string
   → Parser (saneql/parser.cpp) → AST (saneql/ast.h)
   → ast_to_query.cpp → QueryNode tree (operators/*.h)
-  → ColumnNarrowingPass → MapPullupPass → FilterPushdownPass → NodeResolutionPass
+  → ColumnNarrowingPass → FilterPushdownPass → MapPullupPass → NodeResolutionPass
   → addToExecPlan() → Arrow Acero ExecPlan (single shared plan)
   → QueryPlan execution → send result
 ```
--- a/src/silo/query_engine/operators/map_node.test.cpp
+++ b/src/silo/query_engine/operators/map_node.test.cpp
@@ -194,14 +194,21 @@ const QueryTestScenario DECOMPRESS_WITH_USER_MAP_SCENARIO = {
 };
 
 // Regression guard: selecting the (zstd-compressed) sequence column together with a
-// `limit` must still return the correct, order-stable result. (For this particular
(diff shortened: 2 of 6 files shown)


taepper (approved):
Looks good!

PR #1344, opened 2026-06-30
refactor(SaneQL): map pull up: pull filter nodes across map nodes

### LINKED ISSUE #1343: map pull up: pull filter nodes across map nodes
Separate issue for:

> Scope update: `FilterNode` pull-through was originally planned for #1335 but moved here.
> 
> Pulling a `MapNode` up through a `FilterNode` requires knowing which columns the filter predicate references, via `Expression::freeIUs()`. Currently only `FieldRef`, `At` and `ZstdDecompressScalar` implement `freeIUs()`; all filter predicate types (`StringEquals`, `IntEquals`, `And`, `Or`, `Negation`, ...) inherit the base default `{}`. With that gap, a swap decision based on `freeIUs()` would incorrectly pull a `MapNode` above a filter that depends on a produced column, breaking the plan.
> 
> So this issue should additionally cover:
> 
> - **`FilterNode`** pull-through, gated on `freeIUs()` being complete.
> - Implementing `freeIUs()` on all filter predicate expression types (prerequisite). `ColumnNarrowingPass` would also benefit from complete predicate `freeIUs()`.
> 
> #1335 therefore ships `FetchNode`-only pull-up (Fetch references no columns, always safe). 

 _Originally posted by @fengelniederhammer in [#1336](https://github.com/GenSpectrum/LAPIS-SILO/issues/1336#issuecomment-4830684758)_

fengelniederhammer:
resolves #1343

### Summary

Fixes a correctness bug where a filter on a decompressed column produced wrong results, by teaching the optimizer which columns a filter predicate references.

The auto-inserted decompression `MapNode` produces a `STRING` column with the same name as its compressed source (e.g. `seq` → `seq`). `FilterPushdownPass` pushed every filter down past `MapNode`s, so `default.filter(seq = 'ACGT')` ran the filter against the still-compressed column.

Changes:

- **Complete `Expression::freeIUs()`** on all predicate types (previously only `FieldRef`/`At`/`ZstdDecompressScalar`). It reports the columns a predicate references. For filter predicates only `.name` is reliable; `.type` is a placeholder, except sequence predicates which carry their sequence type.
- **`FilterPushdownPass`** no longer pushes a filter below a `Map` that produces a column the filter consumes. Such filters stay above the `Map`; independent filters still push down. Sequence predicates (e.g. `hasMutation`) consume the Map's sequence-typed *input*, so they keep pushing down to the scan.
- **`ColumnNarrowingPass`** now keeps `Map` assignments alive when a surviving filter above the `Map` references the produced column.

Pass order is unchanged (`ColumnNarrowing → FilterPushdown → MapPullup → NodeResolution`).

## PR Checklist

### DIFF
--- a/src/silo/query_engine/expressions/string_equals.cpp
+++ b/src/silo/query_engine/expressions/string_equals.cpp
@@ -22,7 +22,7 @@ StringEquals::StringEquals(std::string column_name, std::optional<std::string> v
       value(std::move(value)) {}
 
 std::vector<schema::ColumnIdentifier> StringEquals::freeIUs() const {
-   return {{column_name, schema::ColumnType::BOOL}};
+   return {{.name = column_name, .type = schema::ColumnType::BOOL}};
 }
 
 std::string StringEquals::toString() const {
--- a/src/silo/query_engine/expressions/string_equals.cpp
+++ b/src/silo/query_engine/expressions/string_equals.cpp
@@ -3,6 +3,7 @@
 
 #include <optional>
 #include <utility>
+#include <vector>
 
 #include "silo/common/panic.h"
 #include "silo/query_engine/expressions/expression.h"
@@ -19,6 +20,10 @@ namespace silo::query_engine::expressions {
 StringEquals::StringEquals(std::string column_name, std::optional<std::string> value)
     : column_name(std::move(column_name)),
       value(std::move(value)) {}
+
+std::vector<schema::ColumnIdentifier> StringEquals::freeIUs() const {
+   return {{column_name, schema::ColumnType::BOOL}};
+}
 
 std::string StringEquals::toString() const {
    if (value.has_value()) {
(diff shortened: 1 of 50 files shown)


src/silo/query_engine/expressions/string_equals.cpp
```diff
@@ -20,6 +21,10 @@ StringEquals::StringEquals(std::string column_name, std::optional<std::string> v
     : column_name(std::move(column_name)),
       value(std::move(value)) {}
 
+std::vector<schema::ColumnIdentifier> StringEquals::freeIUs() const {
+   return {{.name = column_name, .type = schema::ColumnType::BOOL}};
```
fengelniederhammer, line 25: Is there actually a way to figure out the correct column type? It can be STRING or INDEXED_STRING here. `isNull` has a similar issue.

PR #1352, opened 2026-07-15
fix(silo): null sequences should not match N symbol queries

taepper:
### Summary

We interpreted `null` sequences as equivalent to full N sequences. 

Since https://github.com/GenSpectrum/LAPIS-SILO/pull/1125, we return `null` for sequences that were null in the input. We should also not match null sequences in filters for `N`.

## PR Checklist


fengelniederhammer (approved):
👍

PR #1354, opened 2026-07-15
feat(SaneQL): pipeline optimization step: map pull up through filter

### LINKED ISSUE #1343: map pull up: pull filter nodes across map nodes

fengelniederhammer:
resolves #1343

### Summary

Test-only. No production change needed: `FilterPushdownPass` already reorders a filter beneath a `MapNode`. It drops the `FilterNode`, leaves the Map in place, and pushes the filter into the `TableScan` — yielding `Map(scan+filter)`, so a filter runs before an (expensive, e.g. decompression) Map. `MapPullupPass` stays a no-op on filters.

- **`filter_pushdown_pass.test.cpp`** — `pushesFilterThroughDecompressMapIntoTableScan`: filter above a decompression Map lands in the scan, Map stays on top.
- **`map_node.test.cpp`** — end-to-end scenarios for both orderings, decompression + filter, and decompression + filter + `limit()`.

Known limitation (`FILTER_ON_MAPPED_COLUMN`): a filter on a column the Map *produces* is pushed below it to the scan, where the column is absent, and the query errors.

## PR Checklist


taepper:
Hmm, I am wondering whether the complexity is worth it here, maybe we can discuss again? Can we find any cases where this part of map pull-up has an effect that was not already covered by filter-pushdown (which is a more important optimization in the general case)

fengelniederhammer:
> Hmm, I am wondering whether the complexity is worth it here, maybe we can discuss again? Can we find any cases where this part of map pull-up has an effect that was not already covered by filter-pushdown (which is a more important optimization in the general case)

As discussed: The filter pushdown currently doesn't swap map and filter and we should still do that (imagine someone wants to download a single unaligned sequence by filtering for its primary key - we don't want to decompress the whole DB). We'll change this PR: Move the swap to the filter pushdown, make the map pullup a no-op on filters. Then we can also revert the order of the passes to their previous state.

taepper (approved):
LGTM!

PR #1355, opened 2026-07-15
feat(SaneQL): add `isoWeek` scalar function

### LINKED ISSUE #1351: Add scalar function `isoWeek()`
SILO should be able to map any date to its iso-week using a new scalar function `isoWeek()`

fengelniederhammer:
resolves #1351

### Summary

Adds a new SaneQL scalar function `isoWeek()` that maps a date column to its ISO 8601 week number (1–53). Usable inside `map()`, e.g. `default.map({week := date.isoWeek()})`.

- **New `isoWeek(column)` scalar function** — takes a date column, returns its ISO week as an integer; a `null` date yields `null`. Like `at()`, it is a value-producing (non-boolean) function, so it can only be used in `map()`, not as a `filter()` predicate.
- **Validation** — errors if the referenced column is missing or is not a date column.
- **Execution** — translates to Arrow's `iso_week` compute kernel over the `date32` column in the projection.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/expressions/expression.h
+++ b/src/silo/query_engine/expressions/expression.h
@@ -28,6 +28,7 @@ class Expression {
       AND,
       OR,
       AT,
+      ISO_WEEK,
       N_OF,
       NEGATION,
       MAYBE,
--- a/documentation/query_documentation.md
+++ b/documentation/query_documentation.md
@@ -394,7 +394,9 @@ so nucleotide and amino acid sequences cannot be distinguished from ordinary str
 
 ## Scalar Functions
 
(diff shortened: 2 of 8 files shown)


src/silo/query_engine/expressions/expression.h
```diff
@@ -28,6 +28,7 @@ class Expression {
       AND,
       OR,
       AT,
+      ISO_WEEK,
```
taepper, line 31: At some point we want to stop adding a member to this enum for every function, but have a generalized instead, but at this point let's stay with the simpler explicit form

taepper (approved):
LGTM

PR #1356, opened 2026-07-15
fix(silo): remove instance of undefined behavior

taepper:
remove instance of UB


taepper:
Let's just do it as part of #1320

PR #1359, opened 2026-07-16
feat(silo): add update function to python bindings

### LINKED ISSUE #1358: Add `update` function to python bindings
Before allowing writing queries in saneql, we should make that functionality available in the python bindings

taepper:
resolves #1358

### Summary

As a first step this only allows the updating of value columns with _literal_ values.

We already add the functions with the intended interface to the python bindings and use the same saneql parser for parsing the literal values.

## PR Checklist

### DIFF
--- a/python/tests/test_database.py
+++ b/python/tests/test_database.py
@@ -1,6 +1,7 @@
 import pytest
 import os
 import json
+import datetime
 import tempfile
 import shutil
 import pyroaring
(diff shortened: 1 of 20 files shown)


src/silo/database.cpp
```diff
@@ -274,6 +276,35 @@ roaring::Roaring Database::getFilteredBitmap(
    return bitmap;
 }
 
+void Database::updateColumn(
```
fengelniederhammer, line 280: AI: q: any locking vs concurrent queries during this mutation, or is single-threaded mutation guaranteed at the API layer?

python/tests/test_database.py
fengelniederhammer, line 1: AI: nit: `update_column` is missing from the `expected_methods` list in `test_database_has_expected_methods`. Add `'update_column'` so the method-presence test covers the new binding.

src/silo/database.cpp
taepper: We should document it properly, but I do not think it requires a lot of care

fengelniederhammer (approved):

PR #1362, opened 2026-07-17
refactor(build): split the cmake setup

taepper:
### Summary

this clearly separates (source location, deps, cmake structure) a portable silolib which can be used by many consumers (e.g. python, performance benchmarks, `silo`, but also wasm in the future)

as a perparation for #1320 

## PR Checklist


CMakeLists.txt
```diff
@@ -157,90 +113,25 @@ enable_testing()
...
 
 # ---------------------------------------------------------------------------
-# Benchmarks
+# Targets
 # ---------------------------------------------------------------------------
 
-add_subdirectory(performance)
+# The native command-line application (the `silo` executable) is a consumer
+# of the shared silolib declaration
+add_subdirectory(app)
```
fengelniederhammer, line 128: Intuitively, I would expect that the silo lib just offers a library and then consumers use this. This is the other way round: The lib knows its consumers and offers them as build targets.

Is this idiomatic in CMake? Would it be possible to create a lib that doesn't know it's consumers, instead the consumers import the libs CMake file?

app/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+
+find_package(GTest REQUIRED)
+if (NOT GTest_LIBRARIES)
+    set(GTest_LIBRARIES gtest gmock)
+endif ()
+
+# GTest is built without its own main (gtest/*:no_main), so both test binaries
+# share the harness entry point in src/main.test.cpp (repo-root cwd + arrow
+# init + RUN_ALL_TESTS). Each binary compiles it to run only its own tests.
+add_executable(silo_app_test ${SILO_APP_TESTS} "${CMAKE_SOURCE_DIR}/src/main.test.cpp")
```
fengelniederhammer, line 88: This is a bit weird but I guess ok for now? It's just the test entry point.

app/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
+# The native SILO command-line application: the `silo` executable and the
+# Poco-based HTTP API layer it serves. This is a consumer of the portable silolib
+# and built by the root CMakeLists before this subproject is added)
```
fengelniederhammer: AI: nit: stray unmatched `)` at end of sentence ("...before this subproject is added)"). Drop the paren.

CMakeLists.txt
taepper: The "importing" part is not really possible. I find this very much acceptable and also a new LLM session without access to the current repo recommended the same layout. A different case can be made if the lib were in a subdirectory (i.e. we move `src/` folder at the root to `silo/src/` instead). Then we would ideally add a root-level scaffold that first includes the lib's `CMakeLists.txt` and then the consumer's

app/CMakeLists.txt
taepper: I also think it is acceptable and it required us to touch the least amount of sources. I added #1365

CMakeLists.txt
fengelniederhammer: IMO that is where we should land in the long term (organizing all "packages" under `src/`). But I guess this is good enough for now.

fengelniederhammer (approved):

PR #1366, opened 2026-07-20
feat(SaneQL): add sublineage-inclusive aggregation groupBy

### LINKED ISSUE #1192: Add method to compute the count for every lineage
SILO has the `Lineage` query, which computes the count for a lineage (optionally) including its descendants:
[https://github.com/GenSpectrum/LAPIS-SILO/blob/main/documentation/query_documentation.md#lineage-column-string-value-[…]g-includesublineages-boolean](https://github.com/GenSpectrum/LAPIS-SILO/blob/main/documentation/query_documentation.md#lineage-column-string-value-string-includesublineages-boolean)

Loculus wants to receive this count for every lineage in the `LineageDefinitions`. This should be supported by silo

corneliusroemer:
closes #1192.

## What

Adds a `groupBy` column form that returns, for **every defined lineage**, the count of
sequences in that lineage **or any of its sublineages** — in a single query:

```
default
  .filter(...)
  .groupBy({count := count()}, {lineage(pango_lineage, includeSublineages := true)})
```

Output: one row per defined lineage `{<column>: string, count: int64}`, **including
zero-count lineages**. An optional `recombinantFollowingMode := 'alwaysFollow' |
'followIfFullyContainedInClade' | 'doNotFollow'` mirrors the `lineage(...)` filter (default
`doNotFollow`).

## Motivation

Loculus's search-filter lineage dropdown shows a per-lineage count that can include
sublineages. Today it computes that **client-side**: it fetches flat per-lineage counts from
`/aggregated` plus the full `lineageDefinition`, then rolls up the lineage DAG — resolving
aliases and summing descendants across multi-parent (recombinant) edges — in JavaScript. That
duplicates, in the client, tree/alias logic that SILO already owns. This feature lets the
rollup run server-side in a single query, making the authoritative lineage tree in SILO the
single source of truth. (It's a consolidation/correctness win, not a fanout fix — Loculus
issues only ~2 requests today.)

Loculus code: https://github.com/loculus-project/loculus/blob/b88f15f4475467ba8f439cda09a0be61c1099df7/website/src/components/SearchPage/fields/AutoCompleteOptions.ts#L101-L163

## Approach

This is the more performant formulation floated in the issue discussion, rather than the
self-join sketch (SaneQL has no join primitive today, so that would be substantial new
surface area). Each lineage's count is `|filter_bitmap ∩ subtree_bitmap[lineage]|`, computed
with roaring `and_cardinality` over the **precomputed** `LineageIndex` subtree bitmaps — the
same index the existing `includeSublineages` filter uses. Cost is O(#defined lineages)
bitmap intersections, no per-row join, no new preprocessing.

Implemented with the existing unresolved→resolved node pattern (as `mostRecentCommonAncestor`
/ `phyloSubtree`): `UnresolvedLineageAggregateNode` at parse time is resolved to
`LineageAggregateNode` once the optimizer has flattened `(table, filter)` out of the scan.

Canonical lineages are enumerated via the tree id range `[0, numberOfLineages())`, which
excludes aliases by construction (aliases are assigned ids beyond that range), guarded by a
`resolveAlias` identity assertion — so aliases are never double-emitted.

**Non-partitioning:** a row in `A.1` is counted under both `A.1` and `A`, so counts do not
sum to the filtered total. Documented in `query_documentation.md`.

**Scope (v1):** exactly one `lineage(...)` group-by column plus a single `count()` aggregate;
other combinations are rejected with a clear query error.

## Tests

- 10 new `QUERY_TEST` scenarios (`lineage_aggregate_node.test.cpp`): base sublineage rollup,
  all three recombinant-following modes, with/without a filter, alias-not-double-emitted,
  zero-count lineage present, ordering + limit over the result, and the error cases.
- Full suite green locally (1050/1050).

## Notes

Verified with a local build and a container-free end-to-end run against LAPIS (the companion
LAPIS PR adds the `includeSublineagesFor` REST param that emits this SaneQL).


taepper:
Thank you very much for the contribution, I would opt for a different approach however:

1. Expose the internal lineage table:
SILO has a known set of lineage systems with configured names, these are indexed and should be queryable by metadata and the indexed values

2. Query the indexed lineage table directly:
The resulting query will look very similar to the suggested format, but the group-by-fields will instead contain the indexed lineage ancestor. (i.e. it is cheap to look-up the indexed lineage table entry for each row in a database table. After referencing the correct ancestor the special case will be a general `groupBy` operation without any special logic required)

The advantage of this approach is less technical debt within the very special operator. The quoted `unresolved` -> `resolved` pattern was handy in getting our new query interface up and running (it was released just 5 days ago), but it is not sustainable to keep adding each piece of new functionality in this way as the query translation logic will be filled with more special cases.

Additionally, the queryable lineage table offers a lot of flexibility in the future

Issue #1371, opened 2026-07-21
Do not invalidly reorder `filter` nodes


taepper:
The current SILO optimizers reorders this query:

`default.map({week := date.isoWeek()}).filter(week = 20)`

to 

`default.filter(week = 20).map({week := date.isoWeek()})`

which triggers an error during query execution.

The optimizer should not transform valid query plans into invalid ones (even if the `filter` node is not currently executable on inner nodes)

PR #1377, opened 2026-07-21
feat(query_engine): implement free IUs for boolean expressions

### LINKED ISSUE #1374: Properly implement free IUs for boolean expressions
To prepare #1371: properly implement the `freeIUs` function for all expressions that can appear in a filter. We need to know which columns are referenced by a filter to decided whether it can be swapped with a map.

fengelniederhammer:
resolves #1374

### Summary

Implements `freeIUs()` (the set of columns an expression references) for every boolean/filter expression, so the query optimizer can tell which columns a filter depends on. This is groundwork for #1371 (deciding when a `filter` may be safely reordered past a `map`).

- **`freeIUs()` on all filter expressions** — recursive nodes (`And`/`Or`/`NOf` union their children; `Negation`/`Maybe`/`Exact` forward theirs) and leaf predicates (`Between`, `IsNull`, `StringInSet`, `StringSearch`, `LineageFilter`, `PhyloChildFilter`, and the sequence predicates) now report the columns they reference.
- **Leaf predicates carry a typed `ColumnIdentifier`** (name + type) resolved from the input schema at parse time, instead of a bare column name. The type is load-bearing: a filter may run on any node (e.g. a computed `map()` output), not only when pushed into a table scan, so a guessed type is not acceptable.
- **Unknown columns in a filter now fail at parse time** with `The database does not contain the column '<name>'` rather than being deferred to compile. Slightly breaking: a few type-specific messages changed (e.g. an existing non-string column used in `like()` now reports "is not of type string" instead of "does not contain the string column").

## PR Checklist


src/silo/query_engine/scalar_expressions/free_ius.test.cpp
fengelniederhammer, line 1: I wonder whether we want to have this test or rather unit tests at the individual filters (i.e. in `and.test.cpp`, etc.).

src/silo/query_engine/scalar_expressions/mutation_profile.cpp
```diff
@@ -54,6 +54,13 @@ std::string MutationProfile<SymbolType>::toString() const {
    return fmt::format("MutationProfile({}distance={},{})", seq_prefix, distance, input_str);
 }
 
+template <typename SymbolType>
+std::vector<schema::ColumnIdentifier> MutationProfile<SymbolType>::freeIUs() const {
+   // The referenced sequence column. sequence_name is always present (no default
+   // sequence), so this is always resolvable.
+   return {schema::ColumnIdentifier{sequence_name, SymbolType::COLUMN_TYPE}};
```
taepper: Can't we resolve `sequence_name` against the input before constructing this expression? (also in `has_mutation.cpp`, `insertion_contains.cpp`, `symbol_in_set.cpp`, `symbol_equals.cpp`)

Otherwise this looks very good!

src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -86,6 +86,22 @@ schema::ColumnIdentifier resolveColumn(
...
+      std::ranges::find_if(schema, [&](const auto& col) { return col.name == sequence_name; });
+   CHECK_SILO_QUERY(
+      found != schema.end(),
+      "Database does not contain the {} Sequence with name: '{}'",
+      SymbolType::SYMBOL_NAME,
+      sequence_name
+   );
+   return schema::ColumnIdentifier{sequence_name, SymbolType::COLUMN_TYPE};
+}
+
```
taepper: I do not see why we need a specialized `resolveSequenceColumn`? There is no difference between `resolveColumn` and `resolveSequenceColumn` other than (with the type having the possiblity to type-check the result). But this type-checking has nothing to do with the resolved column being of a sequence type.

Arguable, if we want resolveColumn to report type errors, we should also do this in cases like this:
```cpp
ScalarExpressionPtr convertIntComparison(
   const std::string& column_name,
   ast::BinaryOp binary_op,
   uint32_t value,
   const std::vector<schema::ColumnIdentifier>& schema
) {
   auto column = resolveColumn(column_name, schema);
...
         return std::make_unique<scalar_expressions::IntBetween>(column, std::nullopt, value);
```

but maybe this validation should instead happen in the constructor of the ScalarExpression classes? Where is this currently done?

fengelniederhammer: True, we don't need it. It's okay if the sequence filters report "type string" in the free IUs.

taepper (approved):
Very important standardization! Thank you!

PR #1383, opened 2026-07-21
perf(silo): always prime Selection with most-selective predicate

taepper:
A large oversight when I added optimizations to the new storage backend for short-reads:

When no child-operator is present in a `Selection` it will iterate and check the predicates of ALL row-ids. It should skip this step with a `makeBitmap` call to its first predicate, which _might_ provide a specialized implementation of the bitmap computation (`IsInCoveredRegion` does provide this)

### DIFF
--- a/src/silo/query_engine/filter/operators/selection.cpp
+++ b/src/silo/query_engine/filter/operators/selection.cpp
@@ -4,6 +4,7 @@
 #include <cmath>
 #include <compare>
 #include <iterator>
+#include <ranges>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


taepper:
This reduces the time of the `many_short_read_filter` benchmark from ~159s to 6.7s

src/silo/query_engine/filter/operators/selection.cpp
```diff
@@ -95,47 +95,44 @@ bool Selection::matchesPredicates(const PredicateVector& predicates, uint32_t ro
...
+   }
+
+   // `candidates` already satisfies predicates.front(); apply the remaining predicates row by row.
+   if (predicates.size() == 1) {
+      return CopyOnWriteBitmap{std::move(candidates)};
+   }
+   PredicateVector remaining_predicates;
+   remaining_predicates.reserve(predicates.size() - 1);
+   for (size_t i = 1; i < predicates.size(); ++i) {
+      remaining_predicates.push_back(predicates.at(i)->copy());
```
fengelniederhammer: Do we really need to copy all of those here?

fengelniederhammer (approved):

PR #1384, opened 2026-07-22
Optimize aggregation when bitmaps are available

### LINKED ISSUE #1252: Efficient co-occurence calculation
SILO should have an operator that calculates mutational co-occurences from a given set of sequences and mutations.

This should be expressible to the user as an aggregation by multiple positions, such as this:

```
default.groupBy({count:=count()},{main[1], main[5]})
```
which should produce a result such as this:

| main[1] | main[5] | count |
| ------- | ------- | ----- |
| A       | N       | 1     |
| G       | T       | 1     |
| A       | T       | 2     |

This should be computable in an efficient way that can handle >10 different mutations

taepper:
resolves #1252

### Summary

Recognize `groupBy({count()}, ...)` whose grouping keys can be computed
directly from roaring bitmaps and route them through a dedicated
BitmapAggregationNode instead of materializing one row per sequence and
hashing it in the generic map/groupBy pipeline.

Grouping keys may be sequence-position lookups (`main.at(123)`), indexed
string columns (`division`), or a mix of the two. The node partitions the
filtered row-set recursively through a pluggable GroupingDimension,
pruning empty combinations so it scales to many dimensions without a full
Cartesian blow-up, and emits its result in pipeline-sized batches. The
BitmapAggregationRewritePass recognizes the shape after filter pushdown
and declines (falling back to the generic pipeline) whenever it cannot
resolve it, so no valid query is ever turned into an error.

## PR Checklist


src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.h
```diff
@@ -0,0 +1,45 @@
...
+///
+/// Each grouping key must be one of:
+///   * a sequence-position lookup produced by an `At` assignment of a directly preceding `map`
+///     (the mutation co-occurrence pattern), e.g.
+///
+///         ... | map({symbol_123 := main.at(123)}) | groupBy({count := count()}, {symbol_123})
+///
+///   * an indexed string column read straight from the table scan, e.g.
+///
+///         ... | groupBy({count := count()}, {division})
```
fengelniederhammer, line 24: And potentially any other indexed column (e.g. dates in the future)?

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
```diff
@@ -0,0 +1,148 @@
...
+         return nullptr;
+      }
+      // A plain table column: only indexed string columns can be grouped straight from their index.
+      const auto column = scan->table->schema->getColumn(group_by_field.name);
+      if (!column.has_value() || column->type != schema::ColumnType::INDEXED_STRING) {
+         return nullptr;
+      }
+      dimensions.push_back(
+         std::make_unique<operators::IndexedColumnDimension>(column.value(), group_by_field.name)
+      );
```
fengelniederhammer: Would it be possible to solve this via free IUs?

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
```diff
@@ -0,0 +1,148 @@
...
+   operators::QueryNode* aggregate_child,
+   const operators::MapNode* map_node
+) {
+   operators::QueryNode* below = map_node != nullptr ? map_node->child.get() : aggregate_child;
+   if (auto* decompress = dynamic_cast<operators::MapNode*>(below);
+       decompress != nullptr && isDecompressMap(*decompress)) {
+      below = decompress->child.get();
+   }
+   return dynamic_cast<operators::TableScanNode*>(below);
+}
```
fengelniederhammer: That would also mean that this optimization doesn't trigger, if there is e.g. an unrelated `orderBy` or something in between somewhere?

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
```diff
@@ -0,0 +1,148 @@
...
+      dimensions.push_back(
+         std::make_unique<operators::IndexedColumnDimension>(column.value(), group_by_field.name)
+      );
+   }
+
+   return std::make_unique<operators::BitmapAggregationNode>(
+      std::move(scan->table),
+      std::move(scan->filter),
+      std::move(dimensions),
+      node.aggregates[0].output_name
```
fengelniederhammer, line 205: Why do we take `[0]` here? Since we validated that the aggregate node only has a single aggregate?

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
```diff
@@ -0,0 +1,148 @@
...
+         // A sequence-position grouping key. The `At` reads a column that keeps the sequence
+         // column's name (a decompress map only remaps its type to STRING), so resolving that name
+         // against the table schema tells us whether it is really a sequence column.
+         const auto column = scan->table->schema->getColumn(at->input_column.name);
+         if (!column.has_value() || !isSequenceColumn(column.value())) {
+            return nullptr;
+         }
+         dimensions.push_back(std::make_unique<operators::SequencePositionDimension>(
+            column.value(),
+            at->position - 1,
```
fengelniederhammer: AI: `at->position - 1` underflows to `UINT32_MAX` when `position == 0`, producing the confusing error `SymbolInSet<...> position is out of bounds 0 > 5` instead of a clear message. Guard `at->position >= 1` (decline, or clearer error) before the subtraction.

src/silo/query_engine/operators/bitmap_aggregation_node.test.cpp
```diff
@@ -0,0 +1,274 @@
...
+   ])"),
+   .query_options = silo::config::QueryOptions{.materialization_cutoff = 0}
+};
+
+}  // namespace
+
+QUERY_TEST(
+   BitmapAggregation,
+   TEST_DATA,
+   ::testing::Values(
```
fengelniederhammer, line 275: AI: Coverage gap: no scenario groups on a position where rows carry ambiguous nucleotide codes (R/Y/etc.). `buildSymbolBitmaps` iterates all `SYMBOLS` and relies on each stored symbol landing in exactly one group (reference vs. explicit vs. missing vs. null). An ambiguity-code test would pin the disjointness/no-double-count invariant (sum of counts == filtered rows) for that path.

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
fengelniederhammer: Is there a test that covers this case?

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
taepper: Yes, in `isBareCountGroupBy`

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
taepper: Yes, such reorderings should be removed independently of this optimization

src/silo/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
taepper: Unfortunately that does not really give us much, I refactored this method instead and hope it is clearer now

fengelniederhammer (approved):

PR #1385, opened 2026-07-22
feat(query_engine): add join operator

### LINKED ISSUE #1379: Add a `join` operator
Add a join operator:

```
join(left table, right table, on expression, type symbol := inner)
```

that should be able to join 2 inputs on a given expression

taepper:
resolves #1379 

### Summary

Add a `join` pipeline operator to the SaneQL query engine that joins two
input pipelines on an equi-join condition:

    join(left, right, on, type := inner)

- `on` is an equality (or `&&`-conjunction of equalities) between a left
  and a right column; the key pairs are extracted into an Arrow hash join.
- `type` selects the join kind: inner (default), left, right, full,
  leftSemi, rightSemi, leftAnti, rightAnti.

Implemented as a new two-child JoinNode operator backed by Arrow Acero's
"hashjoin" exec node. The optimizer passes are extended for the second
child: FilterPushdownPass rejects any predicate sitting above the join
(which input a predicate belongs to could be derived from freeIUs(), but
pushing into the null-supplying side of an outer join, or pushing a
predicate that references no column, would change the result -- apply
filters to the join inputs instead), and ColumnNarrowingPass keeps both
inputs' columns (including the join keys) intact below the join.

Known limitation: the two inputs must have disjoint column names.
Overlapping names would make column references ambiguous and produce a
result with duplicate column names, so such joins are rejected; rename
one side (e.g. via map()) or drop the duplicates (e.g. via project())
first.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/operators/join_node.cpp
+++ b/src/silo/query_engine/operators/join_node.cpp
@@ -1,0 +1,108 @@
+#include "silo/query_engine/operators/join_node.h"
+
+#include <string_view>
+
+#include <arrow/acero/exec_plan.h>
+#include <arrow/acero/options.h>
+#include <arrow/type.h>
+#include <nlohmann/json.hpp>
+
+#include "silo/common/panic.h"
+
+namespace silo::query_engine::operators {
+
+using arrow::acero::JoinType;
+
+JoinNode::JoinNode(
+   QueryNodePtr left,
+   QueryNodePtr right,
+   std::vector<schema::ColumnIdentifier> left_keys,
+   std::vector<schema::ColumnIdentifier> right_keys,
+   JoinType join_type
+)
+    : left(std::move(left)),
+      right(std::move(right)),
+      left_keys(std::move(left_keys)),
+      right_keys(std::move(right_keys)),
+      join_type(join_type) {
+   // Equi-join key vectors are paired positionally (left_keys[i] == right_keys[i]), so they
+   // must have the same length
+   SILO_ASSERT_EQ(this->left_keys.size(), this->right_keys.size());
+}
+
+std::string_view joinTypeToString(JoinType join_type) {
+   switch (join_type) {
+      case JoinType::INNER:
+         return "inner";
+      case JoinType::LEFT_OUTER:
+         return "left";
+      case JoinType::RIGHT_OUTER:
+         return "right";
+      case JoinType::FULL_OUTER:
+         return "full";
+      case JoinType::LEFT_SEMI:
+         return "leftSemi";
+      case JoinType::RIGHT_SEMI:
+         return "rightSemi";
+      case JoinType::LEFT_ANTI:
+         return "leftAnti";
+      case JoinType::RIGHT_ANTI:
+         return "rightAnti";
+   }
+   SILO_UNREACHABLE();
+}
+
+std::vector<schema::ColumnIdentifier> JoinNode::getOutputSchema() const {
+   // Semi/anti joins act as a filter on one input and emit only that side's columns.
+   if (join_type == JoinType::LEFT_SEMI || join_type == JoinType::LEFT_ANTI) {
+      return left->getOutputSchema();
+   }
+   if (join_type == JoinType::RIGHT_SEMI || join_type == JoinType::RIGHT_ANTI) {
+      return right->getOutputSchema();
+   }
+   auto output = left->getOutputSchema();
+   auto right_schema = right->getOutputSchema();
+   output.insert(output.end(), right_schema.begin(), right_schema.end());
+   return output;
+}
+
+arrow::Result<arrow::acero::ExecNode*> JoinNode::addToExecPlan(
+   arrow::acero::ExecPlan& plan,
+   const std::map<schema::TableName, std::shared_ptr<storage::Table>>& tables,
+   const config::QueryOptions& query_options
+) const {
+   ARROW_ASSIGN_OR_RAISE(auto* left_node, left->addToExecPlan(plan, tables, query_options));
+   ARROW_ASSIGN_OR_RAISE(auto* right_node, right->addToExecPlan(plan, tables, query_options));
+
+   std::vector<arrow::FieldRef> left_key_refs;
+   left_key_refs.reserve(left_keys.size());
+   for (const auto& key : left_keys) {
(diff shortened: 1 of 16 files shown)


src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1511,6 +1683,10 @@ FunctionRegistry::FunctionRegistry() {
    );
 
    registerFunction("unionAll", {{pos("left"), pos("right")}}, handleUnionAll);
+
+   registerFunction(
+      "join", {{pos("left"), pos("right"), pos("on"), named("type", false)}}, handleJoin
+   );
```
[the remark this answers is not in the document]
taepper: @fengelniederhammer `AGENTS.md` seems to catch these well :)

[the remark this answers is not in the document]
fengelniederhammer: Nice!

src/silo/query_engine/operators/join_node.cpp
```diff
@@ -0,0 +1,109 @@
...
+namespace silo::query_engine::operators {
+
+using arrow::acero::JoinType;
+
+JoinNode::JoinNode(
+   QueryNodePtr left,
+   QueryNodePtr right,
+   std::vector<schema::ColumnIdentifier> left_keys,
+   std::vector<schema::ColumnIdentifier> right_keys,
+   arrow::acero::JoinType join_type
```
fengelniederhammer:
```suggestion
   JoinType join_type
```

src/silo/query_engine/operators/join_node.cpp
```diff
@@ -0,0 +1,109 @@
...
+      left_keys(std::move(left_keys)),
+      right_keys(std::move(right_keys)),
+      join_type(join_type) {
+   // Equi-join key vectors are paired positionally (left_keys[i] == right_keys[i]), so they
+   // must have the same length. Downstream code (toJson, the Arrow hash-join configuration)
+   // relies on this; a mismatch would be an out-of-bounds read or an invalid join.
+   SILO_ASSERT_EQ(this->left_keys.size(), this->right_keys.size());
+}
+
+std::string_view joinTypeToString(arrow::acero::JoinType join_type) {
```
fengelniederhammer:
```suggestion
std::string_view joinTypeToString(JoinType join_type) {
```

src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1365,6 +1366,177 @@ operators::QueryNodePtr handlePhyloSubtree(
...
+   const ChildConverter& convert_child
+) {
+   auto left = convert_child(args.at("left"), tables);
+   auto right = convert_child(args.at("right"), tables);
+
+   auto left_schema = left->getOutputSchema();
+   auto right_schema = right->getOutputSchema();
+
+   JoinKeys keys;
+   collectJoinKeys(args.at("on"), left_schema, right_schema, keys);
```
fengelniederhammer, line 1565: AI: risk: paired key columns are matched by name only, with no check that their types agree. Joining a string key against an int key falls through to Arrow, which surfaces a raw Arrow error instead of a friendly `CHECK_SILO_QUERY`. Consider validating `first.column.type == second.column.type` in `collectJoinKeys` (where both sides are resolved) and emitting a query error.

src/silo/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1365,6 +1366,177 @@ operators::QueryNodePtr handlePhyloSubtree(
...
+   CHECK_SILO_QUERY(
```
fengelniederhammer, line 1566: AI: nit: this `!keys.left.empty()` check is unreachable. `collectJoinKeys` either throws (non-BinaryExpr) or pushes at least one key pair, so `keys.left` is never empty on success. Drop it or turn it into a `SILO_ASSERT`.

src/silo/query_engine/operators/join_node.cpp
```diff
@@ -0,0 +1,109 @@
...
+         return "leftSemi";
+      case JoinType::RIGHT_SEMI:
+         return "rightSemi";
+      case JoinType::LEFT_ANTI:
+         return "leftAnti";
+      case JoinType::RIGHT_ANTI:
+         return "rightAnti";
+   }
+   SILO_UNREACHABLE();
+}
```
fengelniederhammer, line 53: AI: q: overlapping non-key column names between the two inputs produce duplicate field names in this concatenated output schema. When the result is serialized to JSON records keyed by column name, the duplicates collide/overwrite silently. The `on`-expression ambiguity case is already rejected upfront — should overlapping output columns be rejected the same way (or at least pinned by a test), rather than only documented as "rename downstream"?

src/silo/query_engine/operators/join_node.test.cpp
```diff
@@ -0,0 +1,327 @@
...
+      JOIN_FULL_OUTER_SCENARIO,
+      JOIN_RIGHT_SEMI_SCENARIO,
+      JOIN_RIGHT_ANTI_SCENARIO,
+      JOIN_DOWNSTREAM_FILTER_SCENARIO,
+      JOIN_WITH_GROUPBY_SCENARIO,
+      JOIN_MULTI_KEY_SCENARIO,
+      JOIN_AMBIGUOUS_COLUMN_SCENARIO,
+      JOIN_UNKNOWN_COLUMN_SCENARIO,
+      JOIN_INVALID_TYPE_SCENARIO
+   )
```
fengelniederhammer, line 341: AI: test coverage gaps (scenarios are otherwise thorough — nice). Consider adding:
1. **Key type mismatch** — join on a string key vs a numeric key; pins the error behavior (currently a raw Arrow error).
2. **Non-equality op in `on`** — e.g. `primaryKey > pk`; the `EQUALS`-only path exists but is untested.
3. **Overlapping non-key output names** — join without renaming a shared column; pins the documented "pass-through" behavior.
4. **`JoinNode::toJson()`** — plan JSON output is not directly asserted.

fengelniederhammer (commented):
Looks good! Just a few things.

fengelniederhammer (approved):

PR #1387, opened 2026-07-22
refactor(query_engine): unify per-type equals into a single Equals expression

### LINKED ISSUE #1373: Unify `XEquals` expressions
Currently we have explicit `StringEquals`, `DateEquals` ... expressions. These should be unified into a single `Equals` expression with members `left` and `right`.

It will be more difficult to find out which filters are able to be efficiently compiled to bitmap filters

The `freeIUs` of this expression is `left.freeIUs() union right.freeIUs()`.

taepper:
resolves #1373 

### Summary

This removes a lot of classes that were not necessary before. Functionality is still limited, but at least `Equals` now reports correct IU usage

## PR Checklist


src/silo/query_engine/scalar_expressions/int_equals.cpp
```diff
@@ -1,63 +0,0 @@
...
-      return std::make_unique<filter::operators::Selection>(
-         std::make_unique<filter::operators::CompareToValueSelection<IntColumn>>(
-            int_column, filter::operators::Comparator::EQUALS, value.value()
-         ),
-         table.row_layout
-      );
-   }
-   return std::make_unique<filter::operators::IndexScan>(
-      CopyOnWriteBitmap{&int_column.null_bitmap}, table.row_layout
-   );
```
fengelniederhammer, line 60: The old expressions had this explicit null case - why don't we need it anymore?

fengelniederhammer: Ah I see, they are already rerouted into `IsNull` during parsing (even before this change).

fengelniederhammer (approved):
LGTM 👍

PR #1391, opened 2026-07-27
refactor(query_engine): compose scalar expressions and collapse stacked maps

taepper:
A refactoring commit that makes the optimization step in #1384 easier.

We first collapse multiple Map-nodes that follow one another, so that the optimization step does not need to look at more than one Map node

### DIFF
--- a/src/silo/query_engine/optimizer/map_pullup_pass.cpp
+++ b/src/silo/query_engine/optimizer/map_pullup_pass.cpp
@@ -1,14 +1,84 @@
 #include "silo/query_engine/optimizer/map_pullup_pass.h"
 
+#include <cstddef>
 #include <memory>
 #include <string>
+#include <unordered_map>
 #include <unordered_set>
+#include <utility>
+#include <vector>
 
 #include "silo/query_engine/operators/fetch_node.h"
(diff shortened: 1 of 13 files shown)


src/silo/query_engine/optimizer/map_pullup_pass.cpp
```diff
@@ -60,4 +130,59 @@ operators::QueryNodePtr MapPullupPass::operator()(operators::OrderByNode& node)
...
+   propagateToNode(node.child);
+
+   if (node.child->kind() != operators::NodeKind::MAP) {
+      return nullptr;
+   }
+   auto& lower = static_cast<operators::MapNode&>(*node.child);
+
+   // The columns the lower map produces, mapped to their defining expressions: an upper expression
+   // referencing one of these reads the lower map's computed value, so it must be inlined.
+   std::unordered_map<std::string, const ScalarExpression*> produced;
```
fengelniederhammer:
```suggestion
   std::unordered_map<std::string, const ScalarExpression*> lower_produced_expressions_by_output_name;
```

fengelniederhammer (approved):

PR #1392, opened 2026-07-27
feat(silo): allow updating of string columns

### LINKED ISSUE #1389: Enable `update` of `string` columns
#1359 added an update method which is restricted by applicable column types. string columns should also be supported

taepper:
resolves #1389

### Summary

This adds the missing `update` method to the three types of string columns

## PR Checklist

### DIFF
--- a/src/silo/storage/column/indexed_string_column.cpp
+++ b/src/silo/storage/column/indexed_string_column.cpp
@@ -5,6 +5,7 @@
 #include <fmt/format.h>
 
 #include "silo/common/bidirectional_string_map.h"
+#include "silo/storage/column/row_id.h"
 
 namespace silo::storage::column {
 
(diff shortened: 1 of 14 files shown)


src/silo/storage/column/indexed_string_column.cpp
```diff
@@ -106,6 +107,39 @@ std::expected<void, std::string> IndexedStringColumn::appendChunk(const Buffer&
    return {};
 }
 
+void IndexedStringColumn::update(
```
fengelniederhammer, line 110: AI: risk: `update()` is public but does not touch `lineage_index`, so calling it on a lineage-backed column silently corrupts the lineage index. The guard currently lives only in the caller (`scalar_column_update.cpp`). Consider a defensive `SILO_ASSERT(!lineage_index.has_value())` here (and analogously a phylo-tree assert in `StringColumn::update`) so misuse fails loudly instead of producing a corrupt index.

fengelniederhammer (approved):
LGTM 👍

Issue #800, opened 2025-06-11
Arrow select-K does not support null-odering


Taepper:
Draft of #799 added support for select-k algorithm instead of fully sorting the result, when orderByFields and limit are supplied simultaneously. When adding tests, I unfortunately discovered that the select-k implementation by arrow (more precisely arrow acero) does not support null values, which are dropped instead. Therefore we cannot use the select-k implementation for now.

When resolving this issue, we will supply an alternate solution to the problem, which also does not require sorting the full set.

The following (stale) [PR in arrow](https://github.com/apache/arrow/pull/38584/files#diff-2465b4eee6998d9bfe055e53dd57510e1a73b529a792e3e75edfb9aaf494a164) also resolved the issue. We will solve it similarly by providing a custom select-k node.

Taepper:
Blocked by https://github.com/apache/arrow/pull/46926

taepper:
Arrow 25.0.0, containing the referenced fix is now released and available on conan: https://conan.io/center/recipes/arrow?version=25.0.0

chaoran-chen:
Amazing! Congratulations on successfully contributing to Arrow! Do we already have issues to make use of this? (Ah, now I see that there are three issues in "is blocking")

PR #1393, opened 2026-07-27
fix(silo): sort nulls consistently as smallest value

### LINKED ISSUE #849: Null-ordering is misleading in multi-field sort
Sorting by multiple fields with different ordering (ascending / descending) can lead to misleading results.

Our API was "nulls are ordered as the smallest element", this was broken in 0.7.0.

With the new arrow library, `NullPlacement` (`AtStart` / `AtEnd`) can only be set globally for all fields. We set it based on the first field to mimic "null = smallest" (Ascending -> `AtStart`, descending -> `AtEnd`), but this does not work for multi-field sort. Here the nulls of the second fields are placed based on the ordering of the first field. This can mean `AtEnd` even when the second field is ascending or `AtStart` even when the second field is descending.

taepper:
resolves #849

### Summary

The blocking issue #800 was finally resolved up-stream. Now we pass the null placement individually for every sort key, so nulls are again ordered as the smallest element

## PR Checklist

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -14,7 +14,7 @@ class SiloRecipe(ConanFile):
     default_options = {"wasm": False}
 
     def requirements(self):
-        self.requires("arrow/22.0.0")
+        self.requires("arrow/25.0.0")
         self.requires("boost/1.85.0")
         self.requires("fast_float/8.1.0")
         self.requires("nlohmann_json/3.12.0")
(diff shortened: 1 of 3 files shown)


fengelniederhammer (approved):

corneliusroemer:
IIUC this PR also enables not-full-sorting for limit requests? Or is this not yet done and just unblocked now?

taepper:
This is now unblocked, but not fixed in this PR

PR #1407, opened 2026-07-28
feat(query_engine): add a specialized OrderByWithLimit node

### LINKED ISSUE #1056: Sampling 10 samples randomly takes a long time.
I did this LAPIS call:

```
curl -X 'GET' \
  'https://lapis.cov-spectrum.org/open/v2/sample/details?orderBy=random&limit=10&dataFormat=JSON&downloadAsFile=false' \
  -H 'accept: application/json'
```

which I think should be this in SILO:

```
{"action":{"randomize":true,"limit":10,"type":"Details"},"filterExpression":{"type":"True"}}
```

and it took very long.

I think with the limit, it should be quite fast.

I'm reporting this here because I think it's a SILO issue, but of course I'm reporting it specifically with the instance running cov-spectrum, so it might also be specific to that instance.

taepper:
resolves #1056

### Summary

This adds another rewrite-pass that removes `limit(orderBy(..))` patterns with a specialized node, which will map to an `OrderByWithLimit` node instead. This new node can map to a selectK implementation instead of doing a full sort of the input data

## PR Checklist

### DIFF
--- a/src/silo/query_engine/operators/order_by_with_limit_node.cpp
+++ b/src/silo/query_engine/operators/order_by_with_limit_node.cpp
@@ -1,0 +1,164 @@
+#include "silo/query_engine/operators/order_by_with_limit_node.h"
+
+#include <cstdint>
+#include <map>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include <arrow/acero/exec_plan.h>
+#include <arrow/acero/options.h>
+#include <arrow/compute/api.h>
+#include <arrow/compute/ordering.h>
+#include <arrow/type.h>
+#include <arrow/util/async_generator_fwd.h>
+#include <nlohmann/json.hpp>
+
+#include "silo/common/panic.h"
+#include "silo/query_engine/exec_node/arrow_util.h"
+#include "silo/query_engine/operators/order_by_randomize.h"
+#include "silo/schema/database_schema.h"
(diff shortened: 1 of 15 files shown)


src/silo/query_engine/operators/order_by_with_limit_node.cpp
```diff
@@ -0,0 +1,164 @@
...
+      const auto null_placement =
+         order_by_field.ascending ? NullPlacement::AtStart : NullPlacement::AtEnd;
+      sort_keys.emplace_back(order_by_field.field.name, sort_order, null_placement);
+   }
+   // A randomize seed adds the per-row random hash as the lowest-priority (tie-breaking) sort key,
+   // matching OrderByNode; with no explicit fields it becomes the sole key, yielding a random
+   // order.
+   if (randomize_seed.has_value()) {
+      sort_keys.emplace_back(RANDOMIZE_HASH_FIELD_NAME);
+   }
```
fengelniederhammer, line 73: This part is also duplicated - is it worth extracting it?

fengelniederhammer (approved):
LGTM

src/silo/query_engine/operators/order_by_with_limit_node.cpp
taepper: I want to remove this in the nearer future anyways. See #1419

PR #1412, opened 2026-07-30
test(performance): split data generation of tests and run with one command

taepper:
This helps the general utilities of the repo, where we can add more features and CI integration to performance tests. 

Importantly this also helps #1151 to provide easier benchmarks, as we want to run the ~same benchmark on sorted and unsorted input

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,30 @@ ${SILO_RELEASE_TEST_EXECUTABLE}: build/Release/build.ninja $(shell find src -typ
 ${SILO_RELEASE_APP_TEST_EXECUTABLE}: build/Release/build.ninja $(shell find src app/src -type f)
 	$(CMAKE) --build build/Release --parallel $(CMAKE_BUILD_PARALLEL_LEVEL) --target silo_app_test
 
+GENERATE_TEST_DATA_EXECUTABLE=./build/Release/performance/generate_test_data
+
+${GENERATE_TEST_DATA_EXECUTABLE}: build/Release/build.ninja $(shell find src performance -type f \( -name '*.cpp' -o -name '*.h' \))
+	$(CMAKE) --build build/Release --parallel $(CMAKE_BUILD_PARALLEL_LEVEL) --target generate_test_data
+
+# Generate the datasets consumed by the performance benchmarks and write them to disk under
+# localTestData/performance/ (see the *_NDJSON_PATH constants in performance/sequence_generator.h).
+# Run this once before running any performance benchmark.
+.PHONY: generateTestData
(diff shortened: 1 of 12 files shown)


performance/CMakeLists.txt
```diff
@@ -4,9 +4,22 @@ function(add_benchmark BENCH_NAME)
...
+add_benchmark(generate_test_data)
+
+set(BENCHMARK_NAMES
+    mutation_benchmark
+    many_string_equals
+    many_short_read_filters
+    nof_sequence_filter
+    sequence_column_insert
+    co_occurrence_benchmark
+)
```
fengelniederhammer, line 17: Can't we autodetect them somehow? So that we don't have to manually edit this list every time we add a benchmark?

taepper: Many repos do not want the auto-globbing that we have in `CMakeLists.txt`. 

Instead listing source files explicitly can be much more maintainable

fengelniederhammer: To me that concept seems like outdated technology. I don't know any other build tool where you manually need to list your source files. The file ending is usually specific enough to determine what is code and what not.
(But ok, I also don't know any other language besides C++ where you manually need to write your build process ;) )

Makefile
fengelniederhammer, line 1: Sort of unrelated: I tried `make benchmarks` and it failed due to some stale cmake cache. `make clean` didn't help. `rm -rf build/Release` did.
-> Why doesn't `make clean` also trash the builds?

performance/run_benchmarks.sh
fengelniederhammer, line 1: The logs are quite noisy. Every benchmark logs things and then its result. The result of the previous benchmark is hidden in the flood of logs of the next one. A focused summary at the end would be nice (as a next step, not now).

fengelniederhammer (approved):

performance/run_benchmarks.sh
taepper: > A focused summary at the end would be nice 

Yep, that is a planned follow-up

Makefile
taepper: Added #1418

performance/CMakeLists.txt
taepper: Well, on the other hand you are not forced into abstractions ^^ But yes, in most cases it is just a way to shoot yourself in the foot

For the explicit source file listing:
E.g.: https://github.com/duckdb/duckdb/blob/main/src/optimizer/CMakeLists.txt

where every directory lists its sources. Requires a disciplined development but can make build rules more "hygienic"

PR #1413, opened 2026-07-30
fix(silo): allow `limit` on `groupBy`

### LINKED ISSUE #1409: Allow `limit` without `orderBy` after `groupBy`
Currently SILO sends an error when a user puts a `limit` on the result of a `groupBy` expression, because that expression's output does not have any ordering.

We should automatically mark this unordered result's Ordering as _implicitly ordered_ if the user requests a limit

taepper:
resolves #1409

### Summary

Note that this `marking` does not move/copy the data, it just moves references to batches through auxiliary nodes

## PR Checklist


fengelniederhammer (approved):
LGTM 👍

PR #1426, opened 2026-08-03
refactor(silo): introduce `RoaringContainer` abstraction

taepper:
small refactoring that introduces a reusable abstraction for representing a single roaring container (part of a roaring bitmap. can represent `2^16` values)


fengelniederhammer (approved):
Good idea 👍

PR #1429, opened 2026-08-03
feat(silo): add opt-in N-way clustered ingestion buffering

### LINKED ISSUE #1151: Enable SILO to automatically cluster inputs

taepper:
resolves #1151

### Summary

Current SILO short-read deployments need to sort their input before ingestion. This is superfluous. SILO has an internally chunked storage layout. It is enough to batch sequences together that have similar coverage characteristics to achieve optimal performance. This adds an opt-in coverage-aware clustering during ingestion.

The flow of data changes from
```
file  ->  buffer  ->   column
                (chunk)
```

```
file -> buffer  ->        n buffers  ->   column
              (distribute)         (chunk)

```

This leads to the following performance metrics:

```
[2026-08-03 11:08:09.550] [info] [clustered_ingestion_benchmark.cpp:181] === Summary (ingestion / 10000 queries) ===
[2026-08-03 11:08:09.550] [info] [clustered_ingestion_benchmark.cpp:188] amplicon-sorted, ingestion clustering off        ingest    4.543s   queries    7.633s
[2026-08-03 11:08:09.550] [info] [clustered_ingestion_benchmark.cpp:188] amplicon-shuffled, ingestion clustering off      ingest    4.855s   queries   55.329s
[2026-08-03 11:08:09.550] [info] [clustered_ingestion_benchmark.cpp:188] amplicon-shuffled, 128-way clustered ingestion   ingest    6.255s   queries    7.361s
```

We optionally pay some cost on insert, but get all the performance back for shuffled input datasets

## PR Checklist


src/silo/append/table_inserter.h
```diff
@@ -1,21 +1,82 @@
...
+   std::optional<std::string> driver_column_name;
+   /// Number of simultaneously open data buffers (>=1). A separate buffer holds null-driver rows.
+   size_t num_buffers = 8;
+   /// A row joins an existing buffer when doing so widens that buffer's covered span by at most
+   /// this fraction of the genome length; otherwise it opens a new buffer (until num_buffers is
+   /// reached).
+   double span_growth_threshold_fraction = 0.10;
+};
+
 class TableInserter {
```
fengelniederhammer, line 43: Nit: file name does not match contained class name

src/silo/append/database_inserter.test.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+TEST(ClusteredBuffering, flushesBufferWhenItReachesChunkSize) {
+   auto table = makeTable(1);
+   constexpr size_t CHUNK_SIZE = silo::storage::column::COLUMN_CHUNK_SIZE;
+   std::string ndjson;
+   ndjson.reserve((CHUNK_SIZE + 1) * 40);
+   for (size_t i = 0; i < CHUNK_SIZE + 1; ++i) {
+      ndjson += row(fmt::format("row-{}", i), coveredSequence(1, 0, 1));
+      ndjson += "\n";
```
fengelniederhammer:
```suggestion
      ndjson += row(fmt::format("row-{}", i), coveredSequence(1, 0, 1));
```
Why not let `row()` already append the `\n`?

fengelniederhammer (approved):
LGTM 👍

src/silo/append/database_inserter.test.cpp
taepper: Good call

Issue #1372, opened 2026-07-21
Allow filters on subexpression


taepper:
Currently filters are not executable on intermediate results within a SaneQL expression.

e.g. `x.map(...).filter(...)` is only executable when the optimizer reorders it to `x.filter(...).map(...)`

This should be generalized and filters should be executable at any point within a plan

taepper:
We should add a function

``` 
   [[nodiscard]] virtual arrow::Expression toArrowExpression() const = 0;
```

to `ScalarExpression`

PR #1431, opened 2026-08-03
fix(query_engine): do not invalidly reorder filter nodes

### LINKED ISSUE #1371: Do not invalidly reorder `filter` nodes

fengelniederhammer:
resolves #1371

### Summary

The optimizer could reorder `map({x := ...}).filter(x = ...)` into an invalid plan that filters on `x` before it exists (runtime error), and for a value-changing replace-in-place map could silently filter on the original column value (wrong result).

- **`FilterPushdownPass` now keeps a filter above a `map`** when it references a column the map produces, unless the assignment is pushdown-transparent (a zstd-decompress / identity of the same-named column, which the scan still evaluates by name). Filters on genuinely derived columns now fail with a clear error instead of miscompiling.
  - It doesn't decompose filters yet: `.map(week = date.isoWeek).filter(hasMutation && week == ...)` is problematic. The mutation filter needs to be pushed down on the table scan to operate on the bitmaps while the isWeek filter must stay above the column producing map.
- **`FilterPushdownPass` runs before `ColumnNarrowingPass`**, so narrowing no longer deletes a producing map that a stuck filter still needs; `ColumnNarrowingPass` marks a surviving filter's referenced columns as required.

Pushing sequence filters (e.g. `hasMutation`) through the implicit decompression map into the scan (#1343) is preserved.

## PR Checklist

### DIFF
--- a/src/silo/query_engine/optimizer/column_narrowing_pass.cpp
+++ b/src/silo/query_engine/optimizer/column_narrowing_pass.cpp
@@ -4,6 +4,7 @@
 #include <vector>
 
 #include "silo/query_engine/operators/aggregate_node.h"
+#include "silo/query_engine/operators/filter_node.h"
 #include "silo/query_engine/operators/join_node.h"
 #include "silo/query_engine/operators/map_node.h"
 #include "silo/query_engine/operators/order_by_node.h"
@@ -16,6 +17,21 @@ namespace silo::query_engine::optimizer {
 
 ColumnNarrowingPass ColumnNarrowingPass::makePass(const operators::QueryNodePtr& node) {
    return ColumnNarrowingPass{node->getOutputSchema()};
+}
+
+// NOLINTNEXTLINE(misc-no-recursion)
+operators::QueryNodePtr ColumnNarrowingPass::operator()(operators::FilterNode& node) {
+   // A filter needs every column its predicate references from its child, in addition to the
+   // columns the parent requires. FilterPushdownPass runs before this pass, so the only
+   // FilterNodes still present are ones that could not be pushed below a producing map (e.g.
+   // filter on an isoWeek/at column).
+   for (const auto& column : node.filter->freeIUs()) {
+      if (std::ranges::find(required, column) == required.end()) {
+         required.push_back(column);
+      }
+   }
+   propagateToNode(node.child);
+   return nullptr;
 }
 
 // NOLINTNEXTLINE(readability-convert-member-functions-to-static,misc-no-recursion)
--- a/src/silo/query_engine/optimizer/column_narrowing_pass.cpp
+++ b/src/silo/query_engine/optimizer/column_narrowing_pass.cpp
@@ -22,9 +22,8 @@ ColumnNarrowingPass ColumnNarrowingPass::makePass(const operators::QueryNodePtr&
 // NOLINTNEXTLINE(misc-no-recursion)
 operators::QueryNodePtr ColumnNarrowingPass::operator()(operators::FilterNode& node) {
    // A filter needs every column its predicate references from its child, in addition to the
-   // columns the parent requires. FilterPushdownPass runs before this pass, so the only
-   // FilterNodes still present are ones that could not be pushed below a producing map (e.g.
-   // filter on an isoWeek/at column).
+   // columns the parent requires. Keeping them required prevents a node below the filter that
+   // produces such a column (e.g. a map) from being pruned away while the filter still needs it.
    for (const auto& column : node.filter->freeIUs()) {
       if (std::ranges::find(required, column) == required.end()) {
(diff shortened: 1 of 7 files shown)


src/silo/query_engine/optimizer/column_narrowing_pass.cpp
```diff
@@ -18,6 +19,21 @@ ColumnNarrowingPass ColumnNarrowingPass::makePass(const operators::QueryNodePtr&
    return ColumnNarrowingPass{node->getOutputSchema()};
 }
 
+// NOLINTNEXTLINE(misc-no-recursion)
+operators::QueryNodePtr ColumnNarrowingPass::operator()(operators::FilterNode& node) {
+   // A filter needs every column its predicate references from its child, in addition to the
+   // columns the parent requires. FilterPushdownPass runs before this pass, so the only
+   // FilterNodes still present are ones that could not be pushed below a producing map (e.g.
+   // filter on an isoWeek/at column).
```
taepper: > ... FilterPushdownPass runs before this pass, so the only
   // FilterNodes still present are ones that could not be pushed below a producing map (e.g.
   // filter on an isoWeek/at column).

This doc will become outdated

src/silo/query_engine/optimizer/filter_pushdown_pass.cpp
```diff
@@ -8,18 +11,47 @@
...
+bool isPushdownTransparent(const operators::MapNode::Assignment& assignment) {
+   const scalar_expressions::ScalarExpression* expression = assignment.expression.get();
+   // See through nested physical realizations down to the underlying column reference.
+   while (const auto* decompress =
+             scalar_expressions::dynCast<scalar_expressions::ZstdDecompressScalar>(expression)) {
+      expression = decompress->input.get();
+   }
+   const auto* field_ref = scalar_expressions::dynCast<scalar_expressions::FieldRef>(expression);
+   return field_ref != nullptr && field_ref->column.name == assignment.output_column.name;
+}
```
taepper: Maybe instead just mention that currently, as a hack, we treat the `zstddecompress` like an identity. 

Also `isIdentity` would be a more fitting name for this function. I do not think it needs a doc above, only for the `while` the exaplanation:

```
Because sequence columns cannot be read into an arrow plan directly, and will be zstd-compressed & decompressed instead, we will treat the zstd-decompression as an identity. Might produce misleading error messages
```

src/silo/query_engine/optimizer/filter_pushdown_pass.cpp
```diff
@@ -28,6 +60,59 @@ operators::QueryNodePtr FilterPushdownPass::operator()(operators::FilterNode& no
...
+   // Matching is by NAME: the produced column and a referencing filter carry identical
+   // {name, type} identifiers (e.g. a decompression map's {segment1, STRING} output and a
+   // hasMutation predicate resolved against it), so only the assignment's expression kind can
+   // distinguish a transparent realization from a genuinely derived value.
+   std::unordered_set<std::string> blocking_columns;
+   for (const auto& assignment : node.assignments) {
+      if (!isPushdownTransparent(assignment)) {
+         blocking_columns.insert(assignment.output_column.name);
+      }
+   }
```
taepper, line 74: Maybe we want to compute the opposite? Just give the list of `field_refs`:

`std::unordered_map<std::string, std::string> field_refs`

and then we can also detect pure renames.

--> the bottom loop body can become something like

```
for (col : free_ius) {
  if (fields_refs.contains(col)){
     // We can continue, and if field_ref renames the col, we need to rename col in the filter expression
  }
  else {
    // Stop, cannot compute it below the filter
  }
}
```
____________________

The full described feature in a separate PR, but putting the logic the other way around already makes sense IMO

taepper (commented):
Looks quite good, the refactoring before really helped!!

src/silo/query_engine/optimizer/filter_pushdown_pass.cpp
taepper: Maybe `isFieldRef` afterall

src/silo/query_engine/optimizer/filter_pushdown_pass.cpp
taepper: After further consideration we will want to implement #1433 here, maybe you can add a TODO (#1433 is blocked by #1372 however)

taepper (approved):
LGTM!

PR #1434, opened 2026-08-03
feat(silo): materialize per-lineage relation tables during preprocessing

### LINKED ISSUE #1368: Expose internal lineage tables
The lineages are currently managed purely internally and are not queryable through the standard interface. Instead we should allow direct queries on the lineages, join them with data and filter accordingly. Notably, the ancestor for each of the lineage-querying modes should be contained within a lineage table entry

taepper:
resolves #1368

### Summary

This adds the option to create a proper database table from lineage definitions. This allows better scalability and functionality. Real joins can be performed on the 

## PR Checklist

### DIFF
--- a/src/silo/config/database_config.cpp
+++ b/src/silo/config/database_config.cpp
@@ -34,6 +34,34 @@ ValueType silo::config::toDatabaseValueType(std::string_view type) {
    throw silo::config::ConfigException("Unknown metadata type: " + std::string(type));
 }
 
+silo::config::LineageIndexType silo::config::toLineageIndexType(std::string_view type) {
+   if (type == "columnMetadata") {
+      return LineageIndexType::COLUMN_METADATA;
+   }
+   if (type == "table") {
+      return LineageIndexType::TABLE;
+   }
+   if (type == "both") {
+      return LineageIndexType::BOTH;
+   }
+   throw silo::config::ConfigException(
+      "Unknown lineageIndexType: '" + std::string(type) +
+      "'. Must be one of 'columnMetadata', 'table', 'both'."
+   );
+}
+
+std::string_view silo::config::lineageIndexTypeToString(LineageIndexType type) {
+   switch (type) {
+      case LineageIndexType::COLUMN_METADATA:
+         return "columnMetadata";
+      case LineageIndexType::TABLE:
+         return "table";
+      case LineageIndexType::BOTH:
+         return "both";
+   }
+   SILO_UNREACHABLE();
+}
+
 namespace {
 
 std::string toString(ValueType type) {
@@ -49,7 +77,7 @@ std::string toString(ValueType type) {
       case ValueType::FLOAT:
          return "float";
    }
-   throw std::runtime_error("Non-exhausting switch should be covered by linter");
+   SILO_UNREACHABLE();
 }
 }  // namespace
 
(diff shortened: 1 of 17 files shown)


src/silo/database.test.cpp
```diff
@@ -297,3 +300,67 @@ TEST(DatabaseTest, canCreateMultipleTablesAndAddData) {
       silo::test::executeQueryToJsonArray(query_plan_2), nlohmann::json::array({{{"count", 1}}})
    );
 }
+
+TEST(DatabaseTest, buildsQueryableLineageRelationTable) {
```
fengelniederhammer: Could we use a `QUERY_TEST` here instead?

src/silo/initialize/initializer.cpp
```diff
@@ -46,7 +57,108 @@ void Initializer::createTableInDatabase(
...
+      if (!config_metadata.generatesLineageTable()) {
+         continue;
+      }
+      auto lineage_tree =
+         findLineageTreeForName(lineage_trees, config_metadata.generate_lineage_index.value());
+      if (!lineage_tree.has_value()) {
+         auto keys =
+            lineage_trees | std::views::keys |
+            std::views::transform([](const std::filesystem::path& path) { return path.string(); });
+         throw silo::initialize::InitializeException(
```
fengelniederhammer:
```suggestion
         throw InitializeException(
```

src/silo/database.cpp
```diff
@@ -332,8 +331,9 @@ DatabaseInfo Database::getDatabaseInfo() const {
       .vertical_bitmaps_size = 0,
       .horizontal_bitmaps_size = 0
    };
-   for (const auto& [_, table] : tables) {
-      addTableStatisticsToDatabaseInfo(database_info, *table);
+   const auto default_table = tables.find(schema::TableName::getDefault());
+   if (default_table != tables.end()) {
+      addTableStatisticsToDatabaseInfo(database_info, *default_table->second);
    }
```
fengelniederhammer, line 341: AI: `getDatabaseInfo` now reports only the `default` table (was: sum of all tables). This correctly keeps lineage-relation-table rows out of `sequence_count`, but it also silently drops stats for any other data table and returns all-zero stats if `default` is absent. Is default-only intentional, or should this aggregate over tables that actually have sequence columns?

src/silo/config/database_config.cpp
```diff
@@ -34,6 +34,34 @@ ValueType silo::config::toDatabaseValueType(std::string_view type) {
...
+std::string_view silo::config::lineageIndexTypeToString(LineageIndexType type) {
+   switch (type) {
+      case LineageIndexType::COLUMN_METADATA:
+         return "columnMetadata";
+      case LineageIndexType::TABLE:
+         return "table";
+      case LineageIndexType::BOTH:
+         return "both";
+   }
+   throw std::runtime_error("Non-exhausting switch should be covered by linter");
```
fengelniederhammer: AI: nit: unreachable throw after an exhaustive switch. `SILO_UNREACHABLE()` would be clearer than `runtime_error`. (Matches the existing `toString` pattern, so optional.)

src/silo/database.cpp
taepper: I think that we will want it like this.

src/silo/database.test.cpp
taepper: This is not quite as easy. `QUERY_TEST` is currently only building databases without lineage tables

taepper: I will give it a try

taepper: This is now a proper test scenario!

src/silo/database.cpp
fengelniederhammer: For now, probably yes. Once we have more table (several sequence data tables + several lineage tables), I think we need to rework the info endpoint anyway. Simply returning a sequence count won't really be meaningful anymore.

taepper: Precisely!

src/silo/initialize/lineage_relation_table.test.cpp
fengelniederhammer, line 1: Nice :)

fengelniederhammer (approved):

PR #1452, opened 2026-08-05
refactor(silo): add a set-algebra api to CopyOnWriteBitmap

### LINKED ISSUE #1191: Filter operator interface improvement
Currently, the interface that is used between filter operators is `CopyOnWriteBitmap`. This is a roaring bitmap, with the additional feature that it might avoid a copy of the original bitmap, if it is never mutated by the consumer.

This abstraction level might be too high. As of #993, SILO's sequence storage operates directly on roaring containers, which lead to huge storage improvements for wastewater data. This abstraction level can improve the performance of `SymbolEquals` and `MutationProfile` queries if applied to the operator interface.

An additional feature could be making negations implicit, by having a "negated" flag directly in the container. We never want to actually compute negations and this might make it possible to model this between operators in a better way

taepper:
Another minor refactor following #1441.

Now we will be able to consolidate and resolve #1191.

This adds set-operation directly to `CopyOnWriteBitmap`. The accessors to the underlying bitmap are now changed to private methods. All callers now directly formulate their computations on our abstraction -> next we can change the internal behavior without touching all call sites!

### DIFF
--- a/src/silo/database.cpp
+++ b/src/silo/database.cpp
@@ -275,7 +275,7 @@ roaring::Roaring Database::getFilteredBitmap(
    auto rewritten_filter_expression =
       filter_expression->rewrite(*table, ScalarExpression::AmbiguityMode::NONE);
    auto filter_operator = rewritten_filter_expression->compile(*table);
-   roaring::Roaring bitmap = filter_operator->evaluate().getConstReference();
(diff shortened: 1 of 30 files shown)


fengelniederhammer (approved):
LGTM

PR #1466, opened 2026-08-06
refactor(silo): move the primary key uniqueness change to `preprocessing`

### LINKED ISSUE #1465: Move the primary key uniqueness check to preprocessing
Currently, `table.h` owns about the primary key uniqueness check. This should be moved to `preprocessing`. Some consumers of SILO (e.g. through python bindings) might decide that they do not want this behavior

taepper:
resolves #1465

### Summary

Currently, table.h owns about the primary key uniqueness check. This should be moved to preprocessing. Some consumers of SILO (e.g. through python bindings) might decide that they do not want this behavior.

This allows are more succinct `createTable` api for the python bindings, where the primary key will not necessarily be specified

### DIFF
--- a/documentation/incremental_preprocessing.md
+++ b/documentation/incremental_preprocessing.md
@@ -52,9 +52,7 @@ When `rhydb append` is invoked, the following steps are performed:
 
 5. Each subsequent line is parsed and its values are inserted into the database. Progress is logged every 10,000 records.
 
-6. After all records have been inserted, the database is finalized and validated. This includes checking for duplicate primary keys across both old and new data.
-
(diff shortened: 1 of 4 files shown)


fengelniederhammer:
AI: 🔴 Docs now stale: `documentation/incremental_preprocessing.md` (L9, L55) still states the `append` operation checks for duplicate primary keys across old+new data. After removing the PK check from `Table::validate()`, the `append` command (`append.cpp:56` → `appendData`) no longer validates uniqueness. Either restore the check in `runAppend` or update the docs to reflect that append no longer validates primary key uniqueness.

fengelniederhammer (approved):

PR #1473, opened 2026-08-12
feat(rhydb): return ISO week date string from isoWeek()

### LINKED ISSUE #1471: Change `isoWeek` to also return year
Currently, the function `isoWeek` returns a number between 1-53 for every date. In almost all cases, we are not interested in this number and would like to get the value including the year, ideally formatted as `2026-W08`. As there are no users of `isoWeek` yet, I would make this breaking change as soon as possible.

If we decide to bring back the old functionality, we could use `isoWeekNumber` for a new scalar expression, but will do so only when the need for it arises

taepper:
resolves #1471 

### Summary

This changes the scalar function `isoWeek` to e.g. return `"2026-W07"`, `"2026-W12"` instead of `7`,`12`. This function is not exposed through LAPIS yet, so has no external dependents

## PR Checklist

### DIFF
--- a/src/silo/query_engine/operators/map_node.cpp
+++ b/src/silo/query_engine/operators/map_node.cpp
@@ -11,6 +11,7 @@
 #include <arrow/acero/options.h>
 #include <arrow/compute/api.h>
 #include <arrow/datum.h>
+#include <arrow/type.h>
 #include <nlohmann/json_fwd.hpp>
 
 #include "silo/common/size_constants.h"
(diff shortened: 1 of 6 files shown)


src/silo/query_engine/operators/map_node.cpp
chaoran-chen, line 1: I thought about this again and wonder whether we should really call the function `isoWeek` which is usually defined differently (as just the week number) or whether it would be better to call it `isoYearWeek`.

taepper: `isoWeekNumber` and `isoWeek` was our thought, but `isoWeek` and `isoYearWeek` also works of course. No strong feelings about this

chaoran-chen: ok, sure, no strong feeling from my side either

chaoran-chen, line 1: Would you like to update the [query_documentation](https://github.com/GenSpectrum/LAPIS-SILO/blob/cb7e952e8c493a9b45ea5dd545a72786333c1a7d/documentation/query_documentation.md#isoweekcolumn) as well?

chaoran-chen (approved):
I think everything looks good and straightforward, so I'm approving this as Fabian is not here at the moment.

PR #1480, opened 2026-08-13
feat(python): add generic create_table binding for arbitrary column types

### LINKED ISSUE #1091: Allow users to create new tables via CLI
The CLI should allow the creation of new tables within a SILO instance

taepper:
resolves #1091

### Summary

Before, we had specialized methods for creating tables via the python api. Now we have a generic method for creating arbitrary tables

## PR Checklist


src/rhydb/database.cpp
```diff
@@ -70,17 +72,127 @@ std::string symbolVectorToString(const std::vector<typename SymbolType::Symbol>&
...
+Database::Database() {
+   createReferencesTable();
+}
+
 Database::Database(schema::DatabaseSchema database_schema)
     : schema(std::move(database_schema)) {
    for (const auto& [table_name, table_schema] : schema.tables) {
       tables.emplace(table_name, std::make_shared<storage::Table>(table_name, table_schema));
    }
 }
```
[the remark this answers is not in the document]
taepper: We instead bump the serialization version to guarantee this

python/rhydb/database.pyx
```diff
@@ -80,93 +81,147 @@ cdef class PyDatabase:
...
+        "date",
+        "bool",
+        "int",
+        "float",
+        "nucleotide_sequence",
+        "amino_acid_sequence",
+        "zstd_compressed_string",
+    })
+
+    def create_table(self, str table_name, list columns=None):
```
fengelniederhammer: Why do we have the `None` default here when we throw an error in that case below anyway?

python/tests/test_database.py
```diff
@@ -38,6 +38,47 @@ def main_reference_sequence(reference_genomes):
...
+    """
+    db.register_reference(name, reference)
+
+
+def create_nucleotide_sequence_table(
+    db, table_name, primary_key_name, sequence_name, reference_sequence, extra_columns=None
+):
+    """Create a nucleotide sequence table through the generic ``create_table`` interface.
+
+    Replaces the removed ``create_nucleotide_sequence_table`` short-hand: the primary key is the
```
fengelniederhammer: Sounds like AI put its prompt into those comments ;) Do we need them?

python/tests/test_database.py
```diff
@@ -549,19 +550,172 @@ def test_gene_table_with_extra_columns(self):
         # If no exception, table was created
 
 
+class TestCreateTable:
```
fengelniederhammer: We could also think about splitting this file and move each test class into its own file. This grows rather long.

python/tests/test_database.py
```diff
@@ -549,19 +550,172 @@ def test_gene_table_with_extra_columns(self):
...
+        """Creating a sequence column with no matching `reference_genomes` entry fails."""
+        from rhydb import Database
+
+        db = Database()
+        self._populate_references(db, {"other": "ACGT"})
+        with pytest.raises(RuntimeError, match="no entry named 'seq'"):
+            db.create_table(
+                "samples",
+                [{"name": "id", "type": "string"}, {"name": "seq", "type": "nucleotide_sequence"}],
+            )
```
fengelniederhammer: Those tests are basically equal, aren't they?

src/rhydb/database.h
```diff
@@ -12,8 +15,38 @@
...
+struct ColumnDefinition {
+   std::string name;
+   std::string type;
+};
+
+/// One row of the built-in `reference_genomes` table: a reference sequence keyed by the name of the
+/// column it belongs to. `type` is the sequence type name ("nucleotide_sequence" or
+/// "amino_acid_sequence"): the preprocessing/initialize path always sets it (see
+/// `Initializer::loadReferences`), while the generic `createTableFromColumns` path does not consult
+/// it (it takes each column's type from its `ColumnDefinition`).
```
fengelniederhammer: Let's maybe put the field specific parts as doc strings on the fields in the struct themselves?

src/rhydb/database.cpp
```diff
@@ -70,17 +74,140 @@ std::string symbolVectorToString(const std::vector<typename SymbolType::Symbol>&
    }
    return result;
 }
+
+rhydb::schema::ColumnType parseColumnTypeName(const std::string& type_name) {
+   using rhydb::schema::ColumnType;
```
fengelniederhammer: Why not move it up so that the return type of the function can also use it?

fengelniederhammer: and the function below has the same `using` again

src/rhydb/database.test.cpp
```diff
@@ -298,3 +301,175 @@ TEST(DatabaseTest, canCreateMultipleTablesAndAddData) {
...
+       {.name = "lineage", .type = "indexed_string"},
+       {.name = "seq", .type = "nucleotide_sequence"},
+       {.name = "gene", .type = "amino_acid_sequence"}}
+   );
+
+   std::stringstream data;
+   data << R"({"key":"id_1","age":42,"qc":0.5,"collected":"2021-03-15","passed":true,)"
+        << R"("country":"Switzerland","lineage":"B.1","seq":{"sequence":"ACGT","insertions":[]},)"
+        << R"("gene":{"sequence":"MFV","insertions":[]}})"
+        << "\n";
```
fengelniederhammer: Nit: we could also construct a json object and serialize it into the stream, that would make it a bit more readable.

src/rhydb/initialize/initializer.cpp
```diff
@@ -98,6 +117,40 @@ void Initializer::createTableInDatabase(
    }
 }
 
+void Initializer::loadReferences(const ReferenceGenomes& reference_genomes, Database& database) {
```
fengelniederhammer: AI: risk — `loadReferences` appends into `reference_genomes` on every `createTableInDatabase` call. With multi-table preprocessing (parent #798) a second table will re-append and hit a duplicate primary key on `name`. Also, a nucleotide and an amino-acid reference sharing the same name collide within a single call (same PK). Fine for the current single-table path, but worth guarding/deduping before multi-table lands.

python/rhydb/database.pyx
```diff
@@ -80,93 +81,147 @@ cdef class PyDatabase:
...
+            cpp_columns.push_back(cpp_column)
 
         try:
-            self.c_database.createNucleotideSequenceTable(cpp_table_name, cpp_primary_key_name, cpp_sequence_name, cpp_reference_sequence, cpp_extra_columns)
+            self.c_database.createTableFromColumns(cpp_table_name, cpp_columns)
         except Exception as e:
             raise RuntimeError(f"Failed to create table '{table_name}': {e}")
 
-    def create_gene_table(self, str table_name, str primary_key_name, str gene_name, str reference_sequence, list extra_columns=None):
+    def register_reference(self, str name, str reference, str sequence_type=None):
```
fengelniederhammer, line 196: AI: risk — `register_reference` writes unvalidated `{name, reference, type}` rows straight into the built-in `reference_genomes` table with no type check. These rows drive column typing in the preprocessing path via `getReferences`, and nothing prevents junk `type` values or duplicate `name`s. Consider validating `type` (and rejecting duplicates) on this path.

python/rhydb/database.pyx
```diff
@@ -80,93 +81,147 @@ cdef class PyDatabase:
...
+        if sequence_type is not None and not isinstance(sequence_type, str):
+            raise TypeError("sequence_type must be a string or None")
 
-        if extra_columns:
-            for col in extra_columns:
-                if not isinstance(col, str):
-                    raise TypeError(f"extra_columns must contain strings, got {type(col)}")
-                cpp_extra_columns.push_back(col.encode('utf-8'))
+        row = {"name": name, "reference": reference, "type": sequence_type or ""}
+        cdef string cpp_table_name = "reference_genomes".encode('utf-8')
```
fengelniederhammer: AI: nit — hardcoded `"reference_genomes"` string instead of the C++ `Database::REFERENCE_GENOMES_TABLE_NAME` constant. Silent drift if the constant is ever renamed; consider exposing and reusing it.

fengelniederhammer: Is that possible?

PR #1483, opened 2026-08-19
refactor(rhydb): add type() function to columns

### LINKED ISSUE #1458: Add `type()` to columns
Columns should receive a `type()` function that returns the stored data-type, which can become independent of the class used for encoding (#1222)

taepper:
resolves #1458

### Summary

This minor refactor adds a dynamic type() function to all columns. Also the `ValueTyp` enum is moved to the `schema` namespace from the `config` namespace, because we will need to differentiate `ColumnType` from `ValueType` in the future. Consequently, nucleotide-sequences and amino-acid-sequences have been added to the `ValueType` enum

## PR Checklist

### DIFF
--- a/src/rhydb/config/database_config.cpp
+++ b/src/rhydb/config/database_config.cpp
@@ -12,7 +12,7 @@
 
 #include "config/config_exception.h"
 
-using rhydb::config::ValueType;
+using rhydb::schema::ValueType;
 
 ValueType rhydb::config::toDatabaseValueType(std::string_view type) {
    if (type == "string") {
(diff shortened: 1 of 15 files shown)


src/rhydb/config/database_config.cpp
```diff
@@ -166,7 +147,7 @@ YAML::Node YAML::convert<rhydb::config::DatabaseMetadata>::encode(
 ) {
    Node node;
    node["name"] = metadata.name;
-   node["type"] = toString(metadata.type);
+   node["type"] = std::string{rhydb::schema::valueTypeToString(metadata.type)};
```
fengelniederhammer: Let's add a `using` for this?

src/rhydb/schema/database_schema.h
```diff
@@ -6,6 +6,7 @@
 #include <ranges>
 #include <vector>
 
+#include <fmt/format.h>
```
fengelniederhammer: Do we need this include?

fengelniederhammer (approved):
LGTM

Issue #1404, opened 2026-07-28
Allow querying of >100 positions for co-occurances


taepper:
It was brought up in the WisePulse meeting, that SILO should be able to handle large lists of position for a co-occurence analysis.

Example list:
https://genspectrum.org/collections/covid/4943

PR #1484, opened 2026-08-19
refactor(rhydb): change internal representation in `CopyOnWriteBitmap`

### LINKED ISSUE #1191: Filter operator interface improvement

taepper:
resolves #1191, also progress to #1404

### Summary

This changes the internal representation of "RowId sets" in `CopyOnWriteBitmap`. Instead of always storing a `roaring::Roaring` bitmap, we instead store individual container (views). This allows zero-copy filtering by just containing a view into existing column-containers. Also, more efficient algorithms can be implemented by having access to the containers directly

## PR Checklist

### DIFF
--- a/src/rhydb/query_engine/copy_on_write_bitmap.cpp
+++ b/src/rhydb/query_engine/copy_on_write_bitmap.cpp
@@ -1,95 +1,366 @@
 #include "rhydb/query_engine/copy_on_write_bitmap.h"
 
+#include <algorithm>
+#include <cstddef>
 #include <cstdint>
+#include <span>
+#include <type_traits>
 #include <utility>
+#include <variant>
 #include <vector>
 
 #include <roaring/roaring.hh>
 
 namespace rhydb::query_engine {
 
-CopyOnWriteBitmap::CopyOnWriteBitmap()
-    : mutable_bitmap(std::make_shared<roaring::Roaring>()),
-      immutable_bitmap(nullptr) {}
+using roaring_util::RoaringContainer;
+using roaring_util::RoaringContainerView;
 
-CopyOnWriteBitmap::CopyOnWriteBitmap(const roaring::Roaring* bitmap)
-    : mutable_bitmap(nullptr),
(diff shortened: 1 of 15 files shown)


src/rhydb/query_engine/copy_on_write_bitmap.cpp
```diff
@@ -1,95 +1,362 @@
...
+   while (idx < container_views.size()) {
+      const uint16_t key = container_views[idx].first;
+      size_t group_end = idx + 1;
+      while (group_end < container_views.size() && container_views[group_end].first == key) {
+         ++group_end;
+      }
+      if (group_end - idx == 1) {
+         // A single container for this key stays a zero-copy view
+         result.keys.push_back(key);
+         result.containers.emplace_back(container_views[idx].second);
```
fengelniederhammer, line 321: AI: nit: the single-container-per-key path pushes the view unconditionally, unlike the multi-container path which routes through `pushIfNonEmpty`. If given an empty view this breaks the "no container is ever empty" invariant: `cardinality() == 0` but `isEmpty() == false`, and `toRoaring()` would append an empty container. Current callers pass only non-empty views (asserted in `getMatchingContainerViews`), so this is latent, but the public factory doc states no non-empty precondition. Either skip an empty singleton (as `pushIfNonEmpty` does) or document the precondition.

fengelniederhammer (approved):
LGTM

PR #1493, opened 2026-08-20
feat(rhydb): add `transitiveClosure` table function

### LINKED ISSUE #1192: Add method to compute the count for every lineage

taepper:
resolves #1192

### Summary

This adds the table function `transitiveClosure` which computes the transitive closure of a set of edges. These edges are given as an input table.

The motivation for this change is issue #1192, which requires outputting the number of sublineages for every lineage. This can now be formulated as:
```
pango_lineage.transitiveClosure('parent', 'lineage', includeVertices:=true)
  .join(default, to = lineage_column)
  .groupBy({count := count()}, {from})
  .orderBy({from})
```

This produces the desired result:

for each lineage, `pango_lineage.transitiveClosure('parent', 'lineage', includeVertices:=true)` will contain one row for each descendent of this lineage. e.g. for lineage system `A -> A.1 -> A.1.1`, the transitive closure will be:
```
A | A
A | A.1
A | A.1.1
A.1 | A.1
A.1 | A.1.1
A.1.1 | A.1.1
```

The resulting join and group-by will for each lineage contain the count of all its descendents in the data

## PR Checklist

### DIFF
--- a/src/rhydb/query_engine/operators/transitive_closure_node.cpp
+++ b/src/rhydb/query_engine/operators/transitive_closure_node.cpp
@@ -1,0 +1,267 @@
+#include "rhydb/query_engine/operators/transitive_closure_node.h"
+
+#include <algorithm>
+#include <cstdint>
+#include <functional>
+#include <optional>
+#include <queue>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include <arrow/acero/exec_plan.h>
+#include <arrow/acero/options.h>
+#include <arrow/array/array_binary.h>
+#include <arrow/builder.h>
+#include <arrow/compute/exec.h>
+#include <arrow/util/async_generator.h>
+#include <nlohmann/json.hpp>
+
+#include "rhydb/query_engine/exec_node/arrow_util.h"
+#include "rhydb/query_engine/illegal_query_exception.h"
+#include "rhydb/schema/database_schema.h"
+
+namespace rhydb::query_engine::operators {
+
+namespace {
+
+/// The interned directed relation: `vertex_names[i]` is the name of vertex `i`, and
+/// `adjacency[i]` holds the vertices directly reachable from `i` via a single edge.
+struct Relation {
+   std::vector<std::string> vertex_names;
+   std::vector<std::vector<uint32_t>> adjacency;
+};
+
+/// Validates that `column_name` is a STRING column of the input and returns its position, which
+/// is also its index into the input's exec batches.
+uint32_t validateAndFindColumn(
+   const std::vector<schema::ColumnIdentifier>& child_schema,
+   const std::string& column_name
+) {
+   const auto found = std::ranges::find_if(child_schema, [&](const auto& column) {
+      return column.name == column_name;
+   });
+   CHECK_SILO_QUERY(
+      found != child_schema.end(),
+      "transitiveClosure() column '{}' is not present in the input's output schema",
+      column_name
+   );
+   CHECK_SILO_QUERY(
+      found->type == schema::ColumnType::STRING,
+      "transitiveClosure() can only be applied to STRING columns, but column '{}' has type {}",
+      column_name,
+      schema::columnTypeToString(found->type)
+   );
+   return static_cast<uint32_t>(std::distance(child_schema.begin(), found));
+}
+
+arrow::Status appendStringColumn(
+   const arrow::Datum& datum,
+   std::vector<std::optional<std::string>>& out
+) {
+   if (!datum.is_array()) {
+      return arrow::Status::Invalid("transitiveClosure() expected an array-typed input column");
+   }
+   const auto array = datum.make_array();
+   const auto& string_array = static_cast<const arrow::StringArray&>(*array);
+   for (int64_t row = 0; row < string_array.length(); ++row) {
+      if (string_array.IsNull(row)) {
+         out.emplace_back(std::nullopt);
+      } else {
+         out.emplace_back(string_array.GetString(row));
+      }
+   }
+   return arrow::Status::OK();
+}
+
+/// Builds the interned relation from the child's exec batches, reading the edge endpoints from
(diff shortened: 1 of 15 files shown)


src/rhydb/query_engine/operators/transitive_closure_node.cpp
```diff
@@ -0,0 +1,267 @@
...
+   return static_cast<uint32_t>(std::distance(child_schema.begin(), found));
+}
+
+arrow::Status appendStringColumn(
+   const arrow::Datum& datum,
+   std::vector<std::optional<std::string>>& out
+) {
+   if (!datum.is_array()) {
+      return arrow::Status::Invalid("transitiveClosure() expected an array-typed input column");
+   }
```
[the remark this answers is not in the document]
taepper: I think it is fine to be incomplete here for now

src/rhydb/query_engine/operators/transitive_closure_node.cpp
```diff
@@ -0,0 +1,301 @@
...
+   };
+
+   for (const auto& batch : batches) {
+      if (!batch.has_value()) {
+         continue;
+      }
+      ARROW_ASSIGN_OR_RAISE(const auto from_array, asStringArray(batch->values[from_index]));
+      ARROW_ASSIGN_OR_RAISE(const auto to_array, asStringArray(batch->values[to_index]));
+      for (int64_t row = 0; row < from_array->length(); ++row) {
+         if (from_array->IsNull(row) || to_array->IsNull(row)) {
```
fengelniederhammer: AI: risk: skipping the whole row drops the non-null endpoint from the vertex set. A lineage with a null `parent` AND no children (isolated root, or any flat/parentless lineage set) is never interned, so `includeVertices:=true` emits no `(v, v)` row for it and the motivating "count every lineage" query (#1192) omits it. Intern each non-null endpoint independently; only add the edge when both are present. (Copilot flagged this too.)

src/rhydb/query_engine/operators/transitive_closure_node.cpp
```diff
@@ -0,0 +1,301 @@
...
+}
+
+arrow::Result<arrow::acero::ExecNode*> TransitiveClosureNode::addToExecPlan(
+   arrow::acero::ExecPlan& plan,
+   const std::map<schema::TableName, std::shared_ptr<storage::Table>>& tables,
+   const config::QueryOptions& query_options
+) const {
+   const auto child_schema = child->getOutputSchema();
+   const uint32_t from_index = validateAndFindColumn(child_schema, from_column);
+   const uint32_t to_index = validateAndFindColumn(child_schema, to_column);
```
fengelniederhammer, line 240: AI: q: if `from` and `to` name the same column, `from_index == to_index` and every edge becomes a self-loop — valid but meaningless output. Intended, or should this be rejected with a clear error?

src/rhydb/query_engine/operators/transitive_closure_node.test.cpp
```diff
@@ -0,0 +1,207 @@
...
+      "transitiveClosure() column 'does_not_exist' is not present in the input's output schema"
+};
+
+}  // namespace
+
+QUERY_TEST(
+   TransitiveClosureTest,
+   TEST_DATA,
+   ::testing::Values(
+      TRANSITIVE_CLOSURE_SCENARIO,
```
fengelniederhammer, line 407: AI: info: test-coverage gaps worth adding: (1) cycle/self-loop input exercising the `include_vertices && !reached[source]` branch; (2) isolated/parentless-root lineage (the drop described above); (3) non-STRING column rejection (schema check + `asStringArray` runtime type error); (4) filter-on-output rejection from `filter_pushdown_pass`.

fengelniederhammer: (probably not all of those are actually worth testing)

fengelniederhammer (approved):
LGTM, AI found a few things

fengelniederhammer (changes requested):
After adding `lineageIndexType: both` to the db config,

```
pango_lineage.transitiveClosure('parent', 'lineage', includeVertices:=true)
  .join(default, to = pango_lineage)
  .groupBy({count := count()}, {from})
  .orderBy({from})
```

fails with `join() on-expression equality must reference equal column types from each input, but 'to' and 'pango_lineage' have mismatching types STRING and DICTIONARY_ENCODED at 2:21`

-> that sounds weird to me. Doesn't that fail the motivating case?

src/rhydb/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1487,7 +1489,8 @@ void collectJoinKeys(
       on_expression.location.toString()
    );
    CHECK_SILO_QUERY(
-      first.column.type == second.column.type,
+      exec_node::columnTypeToArrowType(first.column.type)
```
fengelniederhammer, line 1392: AI: q: comparing the materialized Arrow types (rather than `ColumnType`) fixes the intended dict-encoded ↔ STRING join, but it also silently makes `AMINO_ACID_SEQUENCE`, `NUCLEOTIDE_SEQUENCE` and `ZSTD_COMPRESSED_STRING` mutually joinable (all map to `arrow::binary()`), which the old `ColumnType`-equality check rejected. Intended? If not, consider restricting the relaxation to the string-like types.

taepper: That's okay

fengelniederhammer (approved):

PR #1494, opened 2026-08-20
refactor(storage): store horizontal coverage index as struct-of-arrays

taepper:
Another minor refactor for #1404. This changes the layout of coverage starts and ends to struct-of-arrays instead of array-of-structs. While having a little worse semantics (one struct represented one "value"), this change is required to implement hardware accelerated computations on the `starts` and `ends` vectors

Also, more summary statistics for the per-batch starts and ends are stored. The naming of these statistics also improved


fengelniederhammer (approved):
LGTM

PR #1496, opened 2026-08-20
build: remove default docker config files

taepper:
clean more of the top-level directory. These 2 files were quite unnecessary to keep at the top-level. This instead moves the three lines directly into the `Dockerfile` itself


fhennig (approved):
good!

PR #1500, opened 2026-08-21
refactor(query_engine): extract scalarToArrowExpression into a shared exec_node helper

taepper:
another minor refactor before #1404. This moves the `scalarToArrowExpression` function to its own class, so that classes other than `MapNode` can execute scalar expressions

### DIFF
--- a/src/rhydb/query_engine/exec_node/scalar_to_arrow_expression.h
+++ b/src/rhydb/query_engine/exec_node/scalar_to_arrow_expression.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <arrow/compute/expression.h>
+#include <arrow/result.h>
+
+#include "rhydb/query_engine/scalar_expressions/scalar_expression.h"
+
+namespace rhydb::query_engine::exec_node {
+
+/// Translates a scalar expression into an Arrow compute expression, e.g. for use in a projection or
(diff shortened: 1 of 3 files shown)


src/rhydb/query_engine/exec_node/scalar_to_arrow_expression.h
```diff
@@ -0,0 +1,19 @@
...
+#include <arrow/result.h>
+
+#include "rhydb/query_engine/scalar_expressions/scalar_expression.h"
+
+namespace rhydb::query_engine::exec_node {
+
+/// Translates a scalar expression into an Arrow compute expression, e.g. for use in a projection or
+/// for direct evaluation over a materialized batch. Field references become named `field_ref`s
+/// (resolved when the expression is bound to a schema); `at`, `isoWeek` and zstd-decompress map to
+/// the corresponding Arrow compute calls. Returns `NotImplemented` for expressions with no Arrow
```
fengelniederhammer:
```suggestion
/// (resolved when the expression is bound to a schema); function calls like `at`, `isoWeek` and zstd-decompress map to
/// the corresponding Arrow compute calls. Returns `NotImplemented` for expressions with no Arrow
```

so that we don't have to update this comment every time we add a new function.

fengelniederhammer (approved):
Looks good, other than the Copilot finding.

PR #1501, opened 2026-08-21
feat(query_engine): add `<>` operator

### LINKED ISSUE #1437: Add `<>` operator
The not-equals operator should be implemented for all column types

fengelniederhammer:
resolves #1437

### Summary

Adds the not-equals operator `<>` for every column type. Previously `<>` only existed as a negated equality; it now behaves like the other comparison operators, including excluding null cells.

- **`<>` for all column types** — string (plain + dictionary), int, int64, float, date, and bool. `a <> x` matches every row whose value differs from `x`, and never a null row (a null cell has no value to differ), consistent with the ordering operators.
- **Unified `=`/`<>`/ordering into one `Comparison` expression** — the standalone `Equals` expression is deleted; all six comparison operators now lower through `Comparison`, so the column may sit on either side of any of them.
- **Dictionary fast paths** — `=` compiles to a single index lookup; `<>` to the complement of one lookup (excluded value ∪ nulls) instead of unioning every other value's bitmap, which matters on high-cardinality columns.
- **Boolean `=`/`<>`** select the matching truth-value bitmap directly; ordering operators (`<`, `>`, `<=`, `>=`) remain rejected for booleans.

Note the deliberate divergence: `<>` excludes nulls, whereas `!(a = x)` is a set complement and still includes them. `<> null` is rejected (use `isNotNull()`), matching the `= null` ban from #1504.

## PR Checklist


src/rhydb/query_engine/saneql/ast_to_query.cpp
```diff
@@ -255,48 +275,20 @@ ScalarExpressionPtr convertBinaryExprToFilter(
...
+      case BinaryOp::EQUALS:
+      case BinaryOp::NOT_EQUALS:
       case BinaryOp::LESS_THAN:
       case BinaryOp::LESS_EQUAL:
       case BinaryOp::GREATER_THAN:
       case BinaryOp::GREATER_EQUAL: {
+         const Comparator comparator = toComparator(bin_expr.op);
+         if (auto null_test = convertNullTest(bin_expr, schema, comparator)) {
+            return std::move(null_test).value();
+         }
```
fengelniederhammer: @taepper do we want to keep this special case here? We could maybe also move it to the rewrite step of the expressions?

src/rhydb/query_engine/scalar_expressions/comparison.h
```diff
@@ -11,10 +11,18 @@
...
-/// Ordering comparison predicate: `left <op> right`. One side is a column
+/// Comparison predicate: `left <op> right`, for every comparison operator the
+/// query language offers (`=`, `<>`, `<`, `<=`, `>`, `>=`). One side is a column
 /// reference (FieldRef) and the other a literal value; compile() recognises this
 /// "column <op> constant" shape and lowers it to an efficient filter, dispatching
 /// on the literal's type.
+///
+/// Null cells never match, for any operator. `null` is not a comparable value at
+/// all: `column = null` and `column <> null` are rejected while converting the
+/// operand. Use the `isNull()` / `isNotNull()` functions to test for null.
```
taepper: Let's simplify to
```
`null` is not a comparable value
```

src/rhydb/query_engine/saneql/ast_to_query.cpp
```diff
@@ -251,48 +234,21 @@ ScalarExpressionPtr convertBinaryExprToFilter(
...
+      case BinaryOp::NOT_EQUALS:
       case BinaryOp::LESS_THAN:
       case BinaryOp::LESS_EQUAL:
       case BinaryOp::GREATER_THAN:
       case BinaryOp::GREATER_EQUAL: {
+         const Comparator comparator = toComparator(bin_expr.op);
+         // Both operands are converted in source order rather than inline as constructor
+         // arguments: the evaluation order of function arguments is unspecified, so if
+         // both operands are invalid the error the user sees would otherwise depend on
+         // the compiler. Left first means the leftmost invalid operand is reported.
```
taepper: I don't think we need this comment

src/rhydb/query_engine/scalar_expressions/comparison.cpp
```diff
@@ -115,24 +122,29 @@ std::unique_ptr<filter::operators::Operator> compileTypedComparison(
...
+   const Table& table,
    const std::string& column_name,
    Comparator comparator,
    const std::string& literal
 ) {
    if (table.columns.string_columns.contains(column_name)) {
+      // A non-indexed string column has no per-value index, so rewrite() turns equality on
+      // one into a (single-element) StringInSet that owns the scan (and that Or can merge
+      // with its siblings). Equality must therefore never reach this path; the remaining
+      // comparators have no such rewrite and are scanned here.
```
taepper: simplify to `non-indexed string columns are always converted to StringInSet`?

src/rhydb/query_engine/scalar_expressions/comparison.cpp
```diff
@@ -145,29 +157,87 @@ std::unique_ptr<filter::operators::Operator> compileStringComparison(
...
+      }
+      // `excluded` is a subset of the valid row universe, so covering every row means the
+      // complement is empty. Return Empty directly, mirroring the equality path, instead of
+      // a Complement that evaluates to nothing.
+      if (excluded.cardinality() == table.row_layout.numRows()) {
+         return std::make_unique<Empty>(table.row_layout);
+      }
+      return Operator::negate(
+         std::make_unique<IndexScan>(CopyOnWriteBitmap{std::move(excluded)}, table.row_layout)
+      );
```
taepper: 1. we can just add the individual cardinalities of `null` and `bitmap` (possibly saving the union)
2. we do not want to compute the union eagerly. Let's instead return `Complement(Or(1,2))` or better `And(Complement(1),Complement(2))`

taepper (commented):
Just a few comments

src/rhydb/query_engine/scalar_expressions/comparison.h
```diff
@@ -17,9 +17,7 @@ namespace rhydb::query_engine::scalar_expressions {
 /// "column <op> constant" shape and lowers it to an efficient filter, dispatching
 /// on the literal's type.
 ///
-/// Null cells never match, for any operator. `null` is not a comparable value at
-/// all: `column = null` and `column <> null` are rejected while converting the
-/// operand. Use the `isNull()` / `isNotNull()` functions to test for null.
+/// Null cells never match: `null` is not a comparable value.
```
taepper: We do not use the word `cells` elsewhere

src/rhydb/query_engine/scalar_expressions/comparison.cpp
```diff
@@ -115,24 +129,72 @@ std::unique_ptr<filter::operators::Operator> compileTypedComparison(
...
+      return Operator::negate(std::move(excluded.front()));
+   }
+
+   OperatorVector intersection_children;
+   intersection_children.push_back(Operator::negate(std::move(excluded[0])));
+   OperatorVector intersection_negated_children;
+   intersection_negated_children.push_back(std::move(excluded[1]));
+   return std::make_unique<Intersection>(
+      std::move(intersection_children), std::move(intersection_negated_children), table.row_layout
+   );
```
taepper: This seems like a common use-case to me, where one needs to split up into negated and non-negated ones, because of technical reasons (actually, both are negated, but we may not compile it as such). Can we move this logic to Intersection? (and reuse it elsewhere?)

taepper (commented):
Looks good to me, let's just look at one more thing

src/rhydb/query_engine/scalar_expressions/comparison.cpp
fengelniederhammer: Is that what you had in mind?

taepper (approved):
LGTM

PR #1502, opened 2026-08-24
feat(rhydb): add function `insertInto` and `/admin/query` endpoint

taepper:
This adds the functionality to add values to existing tables. This SaneQL interface should also be used by `createTable`, `dropTable`, `delete` and `update` statements in the future

### DIFF
--- a/app/src/admin_query_handler.cpp
+++ b/app/src/admin_query_handler.cpp
@@ -19,8 +19,12 @@
 
 namespace rhydb_app {
 
-AdminQueryHandler::AdminQueryHandler(std::shared_ptr<ActiveDatabase> database_handle)
-    : database_handle(std::move(database_handle)) {}
+AdminQueryHandler::AdminQueryHandler(
+   std::shared_ptr<ActiveDatabase> database_handle,
+   rhydb::config::QueryOptions query_options
+)
+    : query_options(query_options),
+      database_handle(std::move(database_handle)) {}
 
 void AdminQueryHandler::post(
    Poco::Net::HTTPServerRequest& request,
@@ -40,7 +44,7 @@ void AdminQueryHandler::post(
    SPDLOG_INFO("Request Id [{}] - received admin query: {}", request_id, query_string);
 
    try {
-      const nlohmann::json result = database->executeWrite(query_string);
+      const nlohmann::json result = database->executeWrite(query_string, query_options);
 
       SPDLOG_INFO(
          "Request Id [{}] - admin write statement applied: {}; new data version {}",
--- a/app/src/admin_query_handler.cpp
+++ b/app/src/admin_query_handler.cpp
@@ -14,7 +14,7 @@
 #include <nlohmann/json.hpp>
 
 #include <rhydb/append/append_exception.h>
-#include <rhydb/common/silo_directory.h>
+#include <rhydb/common/rhydb_directory.h>
 #include <rhydb/database.h>
 #include <rhydb/query_engine/illegal_query_exception.h>
 #include <rhydb/query_engine/saneql/parse_exception.h>
--- a/app/src/admin_query_handler.cpp
+++ b/app/src/admin_query_handler.cpp
@@ -1,16 +1,21 @@
 #include "admin_query_handler.h"
 
+#include <filesystem>
 #include <mutex>
+#include <stdexcept>
 #include <string>
 #include <utility>
 
 #include <Poco/Net/HTTPServerRequest.h>
 #include <Poco/Net/HTTPServerResponse.h>
 #include <Poco/StreamCopier.h>
+#include <fmt/format.h>
 #include <spdlog/spdlog.h>
 #include <nlohmann/json.hpp>
 
 #include <rhydb/append/append_exception.h>
+#include <rhydb/common/silo_directory.h>
+#include <rhydb/database.h>
 #include <rhydb/query_engine/illegal_query_exception.h>
 #include <rhydb/query_engine/saneql/parse_exception.h>
 #include <evobench/evobench.hpp>
@@ -23,20 +28,53 @@ namespace rhydb_app {
 AdminQueryHandler::AdminQueryHandler(
    std::shared_ptr<ActiveDatabase> database_handle,
    rhydb::config::QueryOptions query_options,
+   std::filesystem::path data_directory,
    std::shared_ptr<std::mutex> write_mutex
 )
     : query_options(query_options),
+      data_directory(std::move(data_directory)),
       database_handle(std::move(database_handle)),
       write_mutex(std::move(write_mutex)) {}
+
+rhydb::Database AdminQueryHandler::loadDatabaseToWriteTo() const {
+   EVOBENCH_SCOPE("AdminQueryHandler", "loadDatabaseToWriteTo");
+
+   const auto data_source = rhydb::RhyDBDirectory{data_directory}.getMostRecentDataDirectory();
+   if (!data_source.has_value()) {
+      throw std::runtime_error(fmt::format(
+         "the data directory '{}' holds no loadable database state to apply the write to",
+         data_directory.string()
+      ));
+   }
+
+   const auto served_version = database_handle->getActiveDatabase()->getDataVersionTimestamp();
+   const auto version_to_write_to = data_source->data_version.getTimestamp();
+   if (version_to_write_to < served_version) {
+      throw std::runtime_error(fmt::format(
+         "the most recent state in the data directory '{}' has data version {}, which is older "
+         "than the data version {} that is being served. Please resolve this conflict manually",
+         data_directory.string(),
+         version_to_write_to.value,
+         served_version.value
+      ));
+   }
+   if (served_version < version_to_write_to) {
+      SPDLOG_WARN(
+         "Admin write applies to data version {} from the data directory, which is newer than the "
+         "data version {} currently being served",
+         version_to_write_to.value,
+         served_version.value
+      );
+   }
+
+   return rhydb::Database::loadDatabaseState(data_source.value());
+}
 
 void AdminQueryHandler::post(
    Poco::Net::HTTPServerRequest& request,
    Poco::Net::HTTPServerResponse& response
 ) {
    EVOBENCH_SCOPE("AdminQueryHandler", "post");
-
-   // Pin the database so it outlives the append, mirroring QueryHandler.
-   const auto database = database_handle->getActiveDatabase();
 
    const auto request_id = response.get("X-Request-Id");
 
@@ -47,19 +85,25 @@ void AdminQueryHandler::post(
    SPDLOG_INFO("Request Id [{}] - received admin query: {}", request_id, query_string);
 
    try {
-      // One write at a time: an in-place mutation is not safe to run next to another one.
+      // One write at a time
       const std::lock_guard<std::mutex> write_lock{*write_mutex};
 
-      const nlohmann::json result = database->executeWrite(query_string, query_options);
+      rhydb::Database staged_database = loadDatabaseToWriteTo();
+
+      const nlohmann::json result = staged_database.executeWrite(query_string, query_options);
+
+      staged_database.saveDatabaseState(data_directory);
+
+      const auto data_version = staged_database.getDataVersionTimestamp();
 
       SPDLOG_INFO(
          "Request Id [{}] - admin write statement applied: {}; new data version {}",
          request_id,
          result.dump(),
-         database->getDataVersionTimestamp().value
+         data_version.value
       );
 
-      response.set("data-version", database->getDataVersionTimestamp().value);
+      response.set("data-version", data_version.value);
       response.setContentType("application/json");
       std::ostream& output_stream = response.send();
       output_stream << result.dump();
--- a/app/src/admin_query_handler.cpp
+++ b/app/src/admin_query_handler.cpp
@@ -1,0 +1,65 @@
+#include "admin_query_handler.h"
+
+#include <string>
+#include <utility>
+
+#include <Poco/Net/HTTPServerRequest.h>
+#include <Poco/Net/HTTPServerResponse.h>
+#include <Poco/StreamCopier.h>
+#include <spdlog/spdlog.h>
+#include <nlohmann/json.hpp>
+
+#include <rhydb/append/append_exception.h>
+#include <rhydb/query_engine/illegal_query_exception.h>
+#include <rhydb/query_engine/saneql/parse_exception.h>
+#include <evobench/evobench.hpp>
+
+#include "active_database.h"
+#include "bad_request.h"
+
+namespace rhydb_app {
+
+AdminQueryHandler::AdminQueryHandler(std::shared_ptr<ActiveDatabase> database_handle)
+    : database_handle(std::move(database_handle)) {}
+
+void AdminQueryHandler::post(
+   Poco::Net::HTTPServerRequest& request,
+   Poco::Net::HTTPServerResponse& response
+) {
+   EVOBENCH_SCOPE("AdminQueryHandler", "post");
+
+   // Pin the database so it outlives the append, mirroring QueryHandler.
+   const auto database = database_handle->getActiveDatabase();
+
+   const auto request_id = response.get("X-Request-Id");
+
+   std::string query_string;
+   std::istream& istream = request.stream();
+   Poco::StreamCopier::copyToString(istream, query_string);
+
+   SPDLOG_INFO("Request Id [{}] - received admin query: {}", request_id, query_string);
+
+   try {
+      const nlohmann::json result = database->executeWrite(query_string);
+
+      SPDLOG_INFO(
+         "Request Id [{}] - admin write statement applied: {}; new data version {}",
+         request_id,
+         result.dump(),
+         database->getDataVersionTimestamp().value
+      );
+
+      response.set("data-version", database->getDataVersionTimestamp().value);
+      response.setContentType("application/json");
+      std::ostream& output_stream = response.send();
+      output_stream << result.dump();
+   } catch (const rhydb::query_engine::saneql::ParseException& ex) {
+      throw BadRequest(ex.what());
+   } catch (const rhydb::query_engine::IllegalQueryException& ex) {
+      throw BadRequest(ex.what());
+   } catch (const rhydb::append::AppendException& ex) {
+      throw BadRequest(ex.what());
+   }
+}
+
+}  // namespace rhydb_app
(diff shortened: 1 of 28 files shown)


app/src/admin_query_handler.cpp
fengelniederhammer, line 1: Let's mention the new endpoint in the PR description?

app/src/request_handler_factory.cpp
```diff
@@ -56,6 +57,9 @@ std::unique_ptr<Poco::Net::HTTPRequestHandler> RhyDBRequestHandlerFactory::route
          database_handle, runtime_config.query_options
       );
    }
+   if (path == "/admin/query") {
+      return std::make_unique<rhydb_app::AdminQueryHandler>(database_handle);
```
fengelniederhammer: Nit:

```suggestion
      return std::make_unique<AdminQueryHandler>(database_handle);
```

src/rhydb/database.cpp
```diff
@@ -475,6 +478,17 @@ std::string Database::executeQueryAsArrowIpc(const std::string& query_string) co
    return output_stream.str();
 }
 
+nlohmann::json Database::executeWrite(const std::string& query_string) {
+   auto request = query_engine::command::parseRequest(query_string, tables);
+   auto* command = std::get_if<query_engine::command::WriteCommandPtr>(&request);
```
fengelniederhammer:
```suggestion
   const auto request = query_engine::command::parseRequest(query_string, tables);
   const auto* command = std::get_if<query_engine::command::WriteCommandPtr>(&request);
```

app/src/query_handler.cpp
```diff
@@ -55,8 +57,21 @@ void QueryHandler::post(
...
+
+      // The read query endpoint must not mutate the database. A write statement (e.g. `insertInto`)
+      // parses to a WriteCommand and has to go through the dedicated `POST /admin/query` endpoint.
+      auto* query_node = std::get_if<rhydb::query_engine::operators::QueryNodePtr>(&parsed_request);
+      if (query_node == nullptr) {
+         throw BadRequest(
+            "this query writes to the database; use the 'POST /admin/query' endpoint for write "
+            "statements"
+         );
+      }
```
fengelniederhammer, line 71: Let's add a test for this (I think I would prefer an e2e test)?

app/src/admin_query_handler.cpp
fengelniederhammer, line 1: Let's add an e2e test for this endpoint? Can be relatively simple, but I think it would be good to have some coverage of the actual API

src/rhydb/query_engine/command/write_command.h
```diff
@@ -0,0 +1,68 @@
...
+   [[nodiscard]] static WriteStatementRegistry& instance();
+
+  private:
+   std::map<std::string, Entry> entries_;
+};
+
+/// Parses `query_string` and classifies it: if its root names a registered write statement, returns
+/// the built WriteCommand; otherwise converts it to a read QueryNode tree. Throws
+/// IllegalQueryException / ParseException for malformed input
+Request parseRequest(std::string_view query_string, const saneql::Tables& tables);
```
fengelniederhammer, line 73: Naming nitpick: It doen't really fit that this function is in the file "write command", since it can also return a read query.

src/rhydb/query_engine/command/write_command.h
```diff
@@ -0,0 +1,68 @@
...
+)>;
+
+class WriteStatementRegistry {
+  public:
+   struct Entry {
+      saneql::FunctionSignature signature;
+      WriteStatementHandler handler;
+   };
+
+   WriteStatementRegistry();
```
fengelniederhammer, line 54: Does the constructor need to be public?

src/rhydb/query_engine/command/write_command.cpp
```diff
@@ -0,0 +1,97 @@
...
+Request parseRequest(std::string_view query_string, const Tables& tables) {
+   saneql::Parser parser(query_string);
+   auto ast = parser.parse();
+
+   // A write statement is a root function call whose name is registered as a write statement;
+   // anything else is an ordinary read query and goes through the unchanged read conversion path.
+   if (std::holds_alternative<saneql::ast::FunctionCall>(ast->value)) {
+      const auto& call = std::get<saneql::ast::FunctionCall>(ast->value);
+      if (const auto* entry =
+             WriteStatementRegistry::instance().findStatement(call.function_name)) {
```
fengelniederhammer, line 86: Would it be possible to write something like `insertInto().filter()` to circumvent this check?

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,43 @@
...
+   // sides, so every column type that a query can emit and the append path can ingest round-trips
+   // without a bespoke Arrow-to-column bridge.
+   auto query_plan = Planner::planQuery(
+      std::move(source_query_), database.tables, config::QueryOptions{}, "insertInto"
+   );
+
+   std::stringstream ndjson_buffer;
+   exec_node::NdjsonSink output_sink{&ndjson_buffer, query_plan.results_schema};
+   constexpr uint64_t DEFAULT_TIMEOUT_SECONDS = 120;
+   query_plan.executeAndWrite(output_sink, DEFAULT_TIMEOUT_SECONDS);
```
fengelniederhammer: Does this materialize the whole table (at least potentially - depending on the filter) in memory?

documentation/query_documentation.md
```diff
@@ -484,6 +474,43 @@ so nucleotide and amino acid sequences cannot be distinguished from ordinary str
...
+SaneQL construct that writes: it mutates the target table rather than returning rows to the caller.
+
+```
+source.filter(country='CH').insertInto(archive)
+source.filter(country='CH').project({primaryKey, country, age}).insertInto(archive)
+source.insertInto('archive')
+```
+
+The target may be written as a bare identifier (`archive`) or a string literal (`'archive'`). It
+must be an existing table in the database; `insertInto` never creates a table.
```
fengelniederhammer, line 510: Is it possible to write into the source table though?

On staging, RhyDB is publicly available. We should probably protect the instance there somehow against writes (even though the data there doesn't matter so much, we still use it for the dashboards for testing).

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+   auto& target_table = *database.tables.at(target_table_);
+   const size_t rows_before = target_table.row_layout.numRows();
+   database.appendData(target_table_, ndjson_buffer);
```
fengelniederhammer: AI: risk: append is row-by-row (TableInserter flushes in chunks), so a conversion error on a later row leaves earlier rows in the target. The docs claim "no rows are inserted" on error — untrue for mid-stream failures. Either stage/validate the full result before committing, or fix the docs to say a partial insert is possible.

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,43 @@
...
+    : source_query_(std::move(source_query)),
+      target_table_(std::move(target_table)) {}
+
+nlohmann::json InsertCommand::execute(Database& database) {
+   // Plan and run the source query, streaming its result rows out as NDJSON. Reusing the NDJSON
+   // sink and the existing NDJSON append path keeps a single JSON representation shared by both
+   // sides, so every column type that a query can emit and the append path can ingest round-trips
+   // without a bespoke Arrow-to-column bridge.
+   auto query_plan = Planner::planQuery(
+      std::move(source_query_), database.tables, config::QueryOptions{}, "insertInto"
```
fengelniederhammer: AI: risk: hardcodes `config::QueryOptions{}` and a 120s timeout, ignoring the runtime-configured query options. The read `/query` path uses `runtime_config.query_options`, so inserts silently bypass configured limits. Thread the runtime `QueryOptions` through `AdminQueryHandler` → `executeWrite` → `InsertCommand`.

app/src/admin_query_handler.cpp
```diff
@@ -0,0 +1,65 @@
...
+    : database_handle(std::move(database_handle)) {}
+
+void AdminQueryHandler::post(
+   Poco::Net::HTTPServerRequest& request,
+   Poco::Net::HTTPServerResponse& response
+) {
+   EVOBENCH_SCOPE("AdminQueryHandler", "post");
+
+   // Pin the database so it outlives the append, mirroring QueryHandler.
+   const auto database = database_handle->getActiveDatabase();
```
fengelniederhammer: AI: risk: no serialization. A concurrent admin write + read (or two concurrent writes) races the shared `Database`. Docs state in-place mutation is unsafe concurrently, but nothing enforces it. Is there any server-level lock guarding `/admin/query`? If not, consider a mutex around write execution.

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,43 @@
...
+   exec_node::NdjsonSink output_sink{&ndjson_buffer, query_plan.results_schema};
+   constexpr uint64_t DEFAULT_TIMEOUT_SECONDS = 120;
+   query_plan.executeAndWrite(output_sink, DEFAULT_TIMEOUT_SECONDS);
+
+   auto& target_table = *database.tables.at(target_table_);
+   const size_t rows_before = target_table.row_layout.numRows();
+   database.appendData(target_table_, ndjson_buffer);
+   const size_t rows_after = target_table.row_layout.numRows();
+
+   return {{"insertedRows", rows_after - rows_before}};
```
fengelniederhammer, line 70: AI: q: `insertedRows = rows_after - rows_before` assumes no concurrent mutation between the two reads. Fine under the current single-threaded assumption, but would report wrong counts if concurrency is ever added. Worth a note tying it to the concurrency assumption above.

src/rhydb/database.test.cpp
```diff
@@ -298,3 +298,137 @@ TEST(DatabaseTest, canCreateMultipleTablesAndAddData) {
...
+   // insertInto needs both a source and a target.
+   EXPECT_THAT(
+      [&]() { database.executeWrite("insertInto(source)"); },
+      ThrowsMessage<rhydb::query_engine::IllegalQueryException>(
+         ::testing::HasSubstr("insertInto() requires argument 'target'")
+      )
+   );
+
+   // If the query does not produce every column of the target, the append is rejected.
+   EXPECT_ANY_THROW(database.executeWrite("source.project({key}).insertInto(archive)"));
```
fengelniederhammer: AI: test gap: missing/mismatched column only checked via `EXPECT_ANY_THROW`. Assert the error type/message and that no partial rows landed in the target (row count unchanged) — otherwise the "no rows inserted" contract stays unverified.

app/src/admin_query_handler.test.cpp
```diff
@@ -0,0 +1,151 @@
...
+TEST(AdminQueryHandler, insertsQueryResultAndReportsRowCount) {
+   auto handle = makeActiveDatabaseWithSourceData();
+
+   rhydb_app::test::MockResponse response;
+   postAdminQuery(handle, "source.filter(country='CH').insertInto(archive)", response);
+
+   EXPECT_EQ(response.getStatus(), Poco::Net::HTTPResponse::HTTP_OK);
+
+   const auto body = nlohmann::json::parse(response.out_stream.str());
+   EXPECT_EQ(body.at("insertedRows").get<size_t>(), 2);
```
fengelniederhammer, line 162: AI: test gap: happy path does not assert the `data-version` response header that the handler sets on write success. Add an assertion so header regressions are caught.

documentation/query_documentation.md
fengelniederhammer: (Currently it's not possible to create a new table via the API, is it? I.e. the only way to execute write commands on staging would be by writing into an existing table)

src/rhydb/query_engine/command/write_command.h
taepper: I think this is fine for now

src/rhydb/query_engine/command/write_command.h
taepper: Currently not, but we would want to allow the creation of different registries so this is fine IMO

src/rhydb/query_engine/command/insert_command.cpp
taepper: Yes..

documentation/query_documentation.md
taepper: I forgot that I wanted to guard this behind an api-option, which is now done

src/rhydb/query_engine/command/insert_command.cpp
taepper: This is now changed to use a streaming insert instead

src/rhydb/query_engine/command/write_command.cpp
taepper: No, the ast will still look like `filter(insertInto)` and fail

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,72 @@
...
+      target_table_.getName()
+   );
+
+   auto query_plan =
+      Planner::planQuery(std::move(source_query_), database.tables, query_options, "insertInto");
+
+   auto target_table = database.tables.at(target_table_);
+   const size_t rows_before = target_table->row_layout.numRows();
+
+   TableInsertSink output_sink{target_table, query_plan.results_schema};
```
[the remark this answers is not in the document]
taepper: Fine for now

documentation/api.md
```diff
@@ -192,6 +193,50 @@ print(table.to_pandas())
 
 ---
 
+### `POST /admin/query`
+
+Executes a SaneQL **write statement** against the database:
+- `insertInto(query: expression, table: symbol)`. Runs `query` and inserts its result rows into
+`table` — see [`insertInto`](query_documentation.md#`insertInto`) for semantics and
```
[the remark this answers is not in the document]
taepper: Link works for me

documentation/api.md
```diff
@@ -192,6 +193,50 @@ print(table.to_pandas())
...
+limitations.
+
+The write goes through the [data directory](#runtime-configuration): the most recent state there is
+loaded into a database of its own, the statement is applied to that one, and the result is saved
+back as a new data version.
+
+The endpoint is **opt-in**: it is only served when
+[`api.allowAdminEndpoint`](#runtime-configuration) is set to `true`. While it is disabled the path
+does not exist and requests to it get a 404, so an instance that does not enable it stays read-only.
```
fengelniederhammer, line 209: And when you enable it, it's still up to you to secure the endpoint (e.g. with a basic auth password in the proxy around RhyDB)?

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,72 @@
...
+   ScannedTableCollector scanned_tables;
+   scanned_tables.collectFrom(source_query_);
+   CHECK_RHYDB_QUERY(
+      !scanned_tables.scanned_tables.contains(target_table_),
+      "insertInto() cannot write into table '{}' while the query reads from it",
+      target_table_.getName()
+   );
+
+   auto query_plan =
+      Planner::planQuery(std::move(source_query_), database.tables, query_options, "insertInto");
```
fengelniederhammer: Don't we have a proper request id? (We would need to wire it from the admin endpoint to here though)

src/rhydb/query_engine/command/insert_command.cpp
```diff
@@ -0,0 +1,72 @@
...
+   Database& database,
+   const config::QueryOptions& query_options
+) {
+   // Inserting streams into the target table while the source query is still producing, so the
+   // source must not read the table that is being written
+   ScannedTableCollector scanned_tables;
+   scanned_tables.collectFrom(source_query_);
+   CHECK_RHYDB_QUERY(
+      !scanned_tables.scanned_tables.contains(target_table_),
+      "insertInto() cannot write into table '{}' while the query reads from it",
```
fengelniederhammer, line 53: AI: 🔴 bug: this self-read guard can be bypassed, so an `insertInto` *can* read the same table it streams writes into.

`ScannedTableCollector` only sees a `TableScanNode` if the tree walk actually reaches it. It reuses `PipelinePassBase`'s defaults for recursion, but that base only recurses into the node types it explicitly handles (Filter/Project/Map/Aggregate/OrderBy/Fetch/Join/UnionAll/Unresolved*). Any other node hits the generic `template <typename T> operator()(T&)` fallback, which returns `nullptr` **without** propagating into its child. `SchemaNode` (built by `handleSchema` at `ast_to_query.cpp:1039`, i.e. the `.schema` operator) is exactly such an uncovered single-child node, so anything below it is never collected and `scanned_tables` comes back empty.

Reproduced against a running instance (single `default` table):

```
default.insertInto(default)          -> guard fires: "cannot write into table 'default' while the query reads from it"  ✓
default.schema.insertInto(default)   -> guard does NOT fire; planning/execution proceeds
```

The `.schema` case then failed later only incidentally (its `{fieldName,type}` output doesn't match `default`'s columns). Reshape it to match — e.g. `default.schema.map({primary_key := fieldName, ...})` — and you get exactly what this guard exists to prevent: rows streamed into `default` while the `SchemaNode` scan is still reading `default`'s column chunks.

`SchemaNode` is the only convert-time offender today, but the design is fragile: every future single-child `QueryNode` that isn't added to `PipelinePassBase` silently re-opens this hole.

Fix: don't rely on the optimizer base for this traversal. Either give the fallback `operator()(T&)` a generic child-walk (visit all children of any node), or collect scanned tables with a dedicated visitor that recurses unconditionally. A regression test `default.schema.insertInto(default)` (and a nested variant) would lock it down.

fengelniederhammer: My AI was able to execute that on a dummy instance without sequences, not sure how much that should trouble us though. `schema` doesn't really "read" the table.

fengelniederhammer: and it didn't work for any of the other schema changing functions (mutations, group by, etc.)

fengelniederhammer (approved):

src/rhydb/query_engine/command/insert_command.cpp
fengelniederhammer: As discussed: Won't fix now, as the `schema` output doesn't really harm here. But good to keep an eye on it and maybe fix later.

taepper: I added #1537 instead

Issue #1462, opened 2026-08-05
Use efficient sequence filters in `seq.at(<pos>)` cases


taepper:
When the user asks whether a sequence position has a value, it can be achieved using the following two methods:

```
x.filter(seq.at(123) = 'A')
x.filter(nucleotideEquals(position := 123, symbol := 'A', sequenceName := 'main'))
```

but only the second construct will use efficient sequence filters. (Until #1372 the former actually throws an error)

Especially after #1432 the former will be much more succinct (`seq[123] = 'A'`)

PR #1506, opened 2026-08-26
feat(query_engine): allow filters on subexpressions

### LINKED ISSUE #1372: Allow filters on subexpression

fengelniederhammer:
resolves #1372

### Summary
Filters can now run at any point in a query plan, not only when they can be pushed into the table scan. A predicate that can't be pushed down is kept as an executable `FilterNode` (an Arrow filter) above the operator it applies to — so e.g. `map(...).filter(<derived column>)`, `mutations(...).filter(...)`, `schema().filter(...)`, `transitiveClosure(...).filter(...)`, and filters above `join`/`limit`/`groupBy` now work.

- **Fail-closed pushdown** — `FilterPushdownPass` now breaks pushdown by default: every node retains the filter above it (as an Arrow filter) unless explicitly classified as transparent (`project`, deterministic `orderBy`) or a consumer (table scan). Fixes filters previously being silently pushed below row-count/schema-changing nodes (`limit`, `groupBy`, …).
- **Randomized `orderBy` is a barrier** — its sort key depends on input ordinals, so a filter above it is retained rather than pushed below (which would change the order for a given seed); deterministic `orderBy` stays transparent.
- **Negation not yet supported in subexpression filters** — `!(...)` has no Arrow translation yet and is rejected there (see #1525); the native bitmap path is unaffected.

Bitmap-only predicates (e.g. `hasMutation`, `seq.at(n) = symbol`) still only work when pushed into the scan; #1462 tracks `seq.at()`.

## PR Checklist


src/rhydb/query_engine/scalar_expressions/negation.cpp
```diff
@@ -24,6 +25,23 @@ std::vector<schema::ColumnIdentifier> Negation::freeIUs() const {
...
+   // native path (and inherit that inconsistency) or match SQL three-valued logic (and diverge from
+   // the native path). Rather than lock in a behaviour that must change once the native
+   // inconsistency is resolved, we bail out so subexpression filters containing a negation are
+   // rejected. Once #1525 is fixed this should be implemented to match the corrected native
+   // semantics.
+   return arrow::Status::NotImplemented(
+      "negation ('!') is not yet supported in filters on subexpressions (see GitHub issue #1525); "
+      "apply the negation in a filter that is pushed into the table scan instead"
+   );
+}
```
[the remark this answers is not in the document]
fengelniederhammer: I think that is ok for now. The "not implemented" is not supposed to be permanent anyway.

src/rhydb/query_engine/optimizer/filter_pushdown_pass.cpp
```diff
@@ -45,14 +39,56 @@ bool isFieldRef(const operators::MapNode::Assignment& assignment) {
...
+   propagateToNode(node.child);
+   return nullptr;
+}
+
+// Filter-transparent: ordering does not change which rows exist
+// NOLINTNEXTLINE(misc-no-recursion)
+operators::QueryNodePtr FilterPushdownPass::operator()(operators::OrderByNode& node) {
+   propagateToNode(node.child);
+   return nullptr;
+}
```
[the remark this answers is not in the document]
fengelniederhammer: @taepper what do you think?

src/rhydb/query_engine/operators/map_node.test.cpp
```diff
@@ -113,54 +113,6 @@ const QueryTestScenario MAP_FIELD_REF_SCENARIO = {
...
-};
-
-const QueryTestScenario MAP_AT_SEQUENCE_INNER_SCENARIO = {
-   .name = "MAP_AT_SEQUENCE_INNER",
-   .query = "default.map({base := segment1.at(3)}).project({primaryKey, base})",
-   .expected_query_result = nlohmann::json(
-      {{{"primaryKey", "id_0"}, {"base", "G"}}, {{"primaryKey", "id_1"}, {"base", nullptr}}}
-   )
-};
-
```
taepper, line 163: Why are these removed?

src/rhydb/query_engine/optimizer/filter_pushdown_pass.cpp
```diff
@@ -45,14 +39,56 @@ bool isFieldRef(const operators::MapNode::Assignment& assignment) {
 
 }  // namespace
 
+// NOLINTNEXTLINE(misc-no-recursion)
+void FilterPushdownPass::propagateToNode(operators::QueryNodePtr& node) {
+   if (auto replacement = operators::visit(*node, *this)) {
+      node = std::move(replacement);
+   }
+   // Fail-closed default: whatever filters the node did not push into its child or consume itself
```
taepper, line 47: This "Fail-closed" comment is repeated 4 times at definition, declaration, and at multiple usages, maybe we can consolidate a bit?

src/rhydb/query_engine/optimizer/filter_pushdown_pass.cpp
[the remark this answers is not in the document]
taepper: Ugh, that is unfortunately true.. and I would add it the suggested code

(Long-term, I think we want to remove the built-in `seed` functionality at some point. The user should provide its own hashing constructs if desired...)

src/rhydb/query_engine/optimizer/filter_pushdown_pass.cpp
```diff
@@ -93,103 +129,26 @@ operators::QueryNodePtr FilterPushdownPass::operator()(operators::MapNode& node)
...
-      current_filters.empty(),
-      "filter() cannot be applied to the output of schema(); schema() is a source operator "
-      "and its result cannot be filtered. Apply filter() before schema() instead."
-   );
-   // Push filters down *within* the child subtree using a fresh pass, so that filters inside
-   // the child (e.g. `default.filter(...).mutations().schema()`) are pushed into the scan.
-   // NodeResolutionPass requires this: it expects a bare table scan beneath
-   // mutations()/insertions(). A separate instance is used so the filters from above schema()
-   // cannot leak into the child.
+void FilterPushdownPass::barrier(operators::QueryNodePtr& child) {
```
taepper, line 145: Nice addition!

src/rhydb/query_engine/operators/map_node.test.cpp
fengelniederhammer: They are moved to a dedicated `at` test.

taepper (approved):
LGTM

PR #1511, opened 2026-09-01
feat: rename env variables and CLI arguments (SILO -> RhyDb)

### LINKED ISSUE #1470: Rename env-variables to `RHYDB_` prefix and rename CLI arguments
Currently, all environment variables start with `SILO_`. This should be changed to `RHYDB_`

fengelniederhammer:
resolves #1470

### Summary

Renames RhyDB's option interface away from the old `SILO` naming: environment variables now use the `RHYDB_` prefix instead of `SILO_`, and the `rhydb append` directory options are renamed.

- **Env var prefix `SILO_` → `RHYDB_`** — applies to all config env vars, plus the standalone `RHYDB_PANIC` and `RHYDB_KEEP_SERIALIZED_STATE` vars.
- **`rhydb append` options** — `--silo-directory` → `--data-directory` (`RHYDB_DATA_DIRECTORY`, YAML `dataDirectory`) and `--silo-data-source` → `--data-source` (`RHYDB_DATA_SOURCE`, YAML `dataSource`), aligning with `rhydb api` which already uses `dataDirectory`.
- **Deprecation warning** — any env var still using the `SILO_` prefix is logged at WARN and ignored, so old configs at least log something rather than silently falling back to defaults.

No backward-compatible aliases: old `SILO_*` vars and `--silo-*` flags stop being recognized.

## PR Checklist

BREAKING CHANGE: environment variables now use the `RHYDB_` prefix instead of `SILO_`, and the `rhydb append` options `--silo-directory`/`--silo-data-source` are renamed to `--data-directory`/`--data-source` (YAML keys `dataDirectory`/`dataSource`). Update env vars, CLI flags, YAML config keys, and Docker/compose setups accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,9 +51,9 @@ EXPOSE 8081
 
 ENTRYPOINT ["./rhydb"]
 
-ENV SILO_PREPROCESSING_CONFIG="/app/preprocessing_config.yaml"
-ENV SILO_DEFAULT_PREPROCESSING_CONFIG="/app/default_preprocessing_config.yaml"
-ENV SILO_DEFAULT_RUNTIME_CONFIG="/app/default_runtime_config.yaml"
+ENV RHYDB_PREPROCESSING_CONFIG="/app/preprocessing_config.yaml"
+ENV RHYDB_DEFAULT_PREPROCESSING_CONFIG="/app/default_preprocessing_config.yaml"
+ENV RHYDB_DEFAULT_RUNTIME_CONFIG="/app/default_runtime_config.yaml"
 
 LABEL org.opencontainers.image.source="https://github.com/GenSpectrum/LAPIS-SILO"
 LABEL org.opencontainers.image.description="High-performance analytical database for sequence alignment data"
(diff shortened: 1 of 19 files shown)


taepper:
We should mark this as breaking, right?

taepper (approved):
LGTM, we should mark this as breaking. The breaking change footer, which will land in the change-log, should list _all_ variable renames

fengelniederhammer:
> LGTM, we should mark this as breaking. The breaking change footer, which will land in the change-log, should list _all_ variable renames

Really all variable names? IMO that would be overkill. All env variable names are renamed using the same pattern (changing the prefix). That should be sufficiently covered. `--silo-directory` and `--silo-data-source` change outside of that pattern -> they are listed explicitly.

taepper:
> Really all variable names? IMO that would be overkill. All env variable names are renamed using the same pattern (changing the prefix). That should be sufficiently covered. --silo-directory and --silo-data-source change outside of that pattern -> they are listed explicitly.

True, I saw the `silo-data-source` and directory changes and was wondering whether it might be confusing. Yes, for the env-vars it is sufficient to just mention it once, if the `data-directory` and `data-source` changes are mentioned explicitly separately

PR #1517, opened 2026-09-02
feat(query_engine): allow a boolean field ref as filter

### LINKED ISSUE #1378: Allow a boolean `FieldRef` as filter
The current scalar expression `FieldRef` can only be used in map expressions. It should also be possible to use that expression as a filter, if the referenced field is of type `boolean`

fengelniederhammer:
resolves #1378

### Summary

A boolean column can now be used directly as a filter predicate: `default.filter(isHuman)` (equivalent to `isHuman = true`) and `default.filter(!isHuman)`. Previously a bare column reference was only valid in `map()`/scalar contexts.

- **`filter(boolColumn)`** — a bare boolean `FieldRef` compiles to the column's `true` bitmap. `!boolColumn` is a set complement, so null rows are kept (matching `!(boolColumn = true)`), not SQL three-valued logic.
- **Non-boolean bare reference is rejected** at `FieldRef::compile()` (e.g. `filter(intField)`); use an explicit comparison instead.

## PR Checklist

### DIFF
--- a/src/rhydb/query_engine/scalar_expressions/field_ref.h
+++ b/src/rhydb/query_engine/scalar_expressions/field_ref.h
@@ -12,8 +12,9 @@ namespace rhydb::query_engine::scalar_expressions {
 
 /// References an existing column by name. As a scalar expression it evaluates to
 /// that column's value per row, e.g. `y := age`; its type() is the referenced
-/// column's type. It is not a filter predicate, so it cannot be compile()d into a
-/// filter operator.
+/// column's type. As a filter predicate it is only valid for boolean columns
+/// (e.g. `filter(isHuman)`), where compile() selects the rows whose value is
+/// `true`; compiling a non-boolean reference is rejected.
 class FieldRef : public ScalarExpression {
   public:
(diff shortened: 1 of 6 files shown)


src/rhydb/query_engine/scalar_expressions/field_ref.h
```diff
@@ -12,8 +12,9 @@ namespace rhydb::query_engine::scalar_expressions {
 
 /// References an existing column by name. As a scalar expression it evaluates to
 /// that column's value per row, e.g. `y := age`; its type() is the referenced
-/// column's type. It is not a filter predicate, so it cannot be compile()d into a
-/// filter operator.
+/// column's type. As a filter predicate it is only valid for boolean columns
+/// (e.g. `filter(isHuman)`), where compile() selects the rows whose value is
+/// `true`; compiling a non-boolean reference is rejected.
```
taepper: I do not think this comments adds any value. Let's remove any mention of filter predicates here

taepper (approved):
LGTM

PR #1535, opened 2026-09-04
fix(query_engine): orderby produces internal server error when result is empty

### LINKED ISSUE #1530: Orderby proportion produces internal server error when result is empty
For example, when sending this request
```
https://lapis.cov-spectrum.org/open/v2/sample/aminoAcidMutations?dateFrom=2027-01-01&orderBy=proportion&limit=5
```
I get: 
```
{
  "error": {
    "detail": "Could not connect to silo: class java.io.IOException HTTP/1.1 header parser received no bytes",
    "status": 500,
    "type": "about:blank"
  },
  "info": {
    "dataVersion": null,
    "requestId": "59e1dc66-0090-4589-9e67-bd86a96af662",
    "requestInfo": "sars_cov-2_nextstrain_open on lapis.cov-spectrum.org at 2026-09-03T14:10:12.199166780",
    "lapisVersion": "0.8.6",
    "siloVersion": "0.13.3"
  }
}
```
The problem seems to be orderBy=`proportion` when the result is empty. As normal behaviour, I would expect just returning empty data as it does without the orderBy argument.

fengelniederhammer:
resolves #1530

### AI Summary

`orderBy(...).limit(n)` queries that produced an empty result crashed RhyDB with an internal server error (e.g. `aminoAcidMutations` with a filter matching no rows). The fused order-by-with-limit node used Arrow's `select_k`, whose empty-input handling returns an unset `Datum` that the following `Take` dereferences, terminating the process.

- **Replace `select_k` with a full sort + fetch** — `OrderByWithLimitNode` now emits `order_by_sink` followed by a `FetchNode` for the `offset + limit` window. Both Arrow sinks fully materialize and `Take`, so there is no memory regression, and `order_by_sink` handles empty input correctly (like `OrderByNode`).
- **Refactor first (separate commit)** — extracted sort-key and plan-node construction into helpers to keep `addToExecPlan()` under the cognitive-complexity limit; no behavior change.

Key fact that made the chosen fix clearly best: `select_k_sink` and `order_by_sink` are the **same Acero node** — both accumulate everything then `Take`. So `select_k` was never memory-bounded top-k here; it only saved a `log n / log k` factor on one CPU phase, while shipping a `std::terminate`.

<details>
<summary>Other fix options considered</summary>

**1. Guard node before the `select_k` sink** — insert something between child and `select_k` so it never sees 0 rows. *Impossible:* `DoFinish()` always runs; a 0-row batch (or zero batches) still yields a 0-row table → same unset `Datum` → `Take` crash. Dropping empty batches doesn't help.

**2. Custom buffering top-k sink** — keep `select_k` semantics via our own sink→buffer→source: accumulate child batches, call `SelectKUnstable` only when `rows > 0`, else emit empty. *Viable but costly:* ~60–100 lines of concurrency-adjacent code reimplementing Arrow's `FromRecordBatches`/`SelectKUnstable`/`Take` (Future composition, memory pool, threading) plus edge-case tests. Only justified if profiling later proves the top-k CPU win matters.

**3. Full sort + fetch — chosen.** Swap `select_k_sink` → `order_by_sink` + `FetchNode`. No memory regression, handles empty correctly.

**4. Wrap/patch Arrow's `SelectKUnstable`** — intercept the meta-function to return empty indices on 0 rows. *Can't reach it:* the crash is inside `OrderBySinkNode::DoFinish` (anonymous namespace, not overridable); `SinkNode` isn't subclassable either. We can't patch vendored Arrow.

**5. Drop the rewrite entirely (max-YAGNI)** — delete `SelectKRewritePass` + `OrderByWithLimitNode` and let `orderBy` + `limit` run as two nodes (that non-fused path already works on empty). Behaviorally identical to option 3's result, but throws away the node as a seam and its tests. We kept the node so a *correct* custom top-k (option 2) can be reintroduced later behind a benchmark.

**6. Upgrade / patch Arrow upstream** — newer Arrow or a conan patch fixing the empty-input `Datum`. Out of scope; higher risk/coupling than a local fix.

</details>

## PR Checklist

### DIFF
--- a/src/rhydb/query_engine/operators/mutations_node.test.cpp
+++ b/src/rhydb/query_engine/operators/mutations_node.test.cpp
@@ -77,6 +77,19 @@ const QueryTestScenario MUTATIONS_MIN_PROPORTION_EXCLUDES_ALL = {
    .expected_query_result = nlohmann::json::array(),
 };
 
+// Regression for #1530: orderBy on an empty mutations result must return empty, not crash.
+const QueryTestScenario MUTATIONS_EMPTY_ORDER_BY_PROPORTION = {
+   .name = "MUTATIONS_EMPTY_ORDER_BY_PROPORTION",
+   .query = "default.mutations(minProportion:=0.6).orderBy({proportion})",
+   .expected_query_result = nlohmann::json::array(),
+};
+
+const QueryTestScenario MUTATIONS_EMPTY_ORDER_BY_PROPORTION_WITH_LIMIT = {
+   .name = "MUTATIONS_EMPTY_ORDER_BY_PROPORTION_WITH_LIMIT",
+   .query = "default.mutations(minProportion:=0.6).orderBy({proportion}).limit(5)",
+   .expected_query_result = nlohmann::json::array(),
+};
+
 const QueryTestScenario MUTATIONS_SEQUENCE_NAMES_SELECTS = {
    .name = "MUTATIONS_SEQUENCE_NAMES_SELECTS",
    .query =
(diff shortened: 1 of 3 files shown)


fengelniederhammer:
Superseded by #1544

PR #1543, opened 2026-09-07
chore: finish renaming to RhyDB in the code

### LINKED ISSUE #1489: Check remaining `silo` mentions in code-base
As the last step of #1421, we should check the code-base again for any remaining mentions of the old name

fengelniederhammer:
resolves #1489

### Summary
I think that should be it. There are still some occurence of "SILO" in the codebase, but I think they are all expected for now.

## PR Checklist


taepper (approved):
LGTM

PR #1544, opened 2026-09-07
fix(rhydb): implement custom arrow::acero node for selectK

### LINKED ISSUE #1530: Orderby proportion produces internal server error when result is empty

taepper:
resolves #1530

### Summary

This implements a custom selectK node implementation, which does not have the same bug as `arrow::acero`'s provided node.

## PR Checklist

### DIFF
--- a/src/rhydb/query_engine/exec_node/select_k.cpp
+++ b/src/rhydb/query_engine/exec_node/select_k.cpp
@@ -1,0 +1,238 @@
+#include "rhydb/query_engine/exec_node/select_k.h"
+
+#include <algorithm>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <sstream>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include <arrow/acero/exec_plan.h>
+#include <arrow/acero/options.h>
+#include <arrow/acero/query_context.h>
+#include <arrow/acero/util.h>
+#include <arrow/compute/api_vector.h>
+#include <arrow/compute/exec.h>
+#include <arrow/compute/ordering.h>
+#include <arrow/datum.h>
+#include <arrow/record_batch.h>
+#include <arrow/result.h>
+#include <arrow/status.h>
+#include <arrow/table.h>
+
+#include "rhydb/common/panic.h"
+
+namespace rhydb::query_engine::exec_node {
+
+namespace {
+
+constexpr std::string_view SELECT_K_FACTORY_NAME = "rhydb_select_k";
+
+// Carries the sort order and output window from `addSelectKNode` into the node factory.
+struct SelectKNodeOptions : public arrow::acero::ExecNodeOptions {
+   SelectKNodeOptions(arrow::Ordering ordering, int64_t offset, int64_t limit)
+       : ordering(std::move(ordering)),
+         offset(offset),
+         limit(limit) {}
+
+   arrow::Ordering ordering;
+   int64_t offset;
+   int64_t limit;
+};
+
+// A select-k Acero node
+// Closely following the implementation of the acero-native SelectK-node:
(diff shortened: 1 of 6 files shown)


src/rhydb/query_engine/exec_node/select_k.h
```diff
@@ -0,0 +1,24 @@
...
+#include <arrow/result.h>
+
+namespace rhydb::query_engine::exec_node {
+
+/// Appends a select-k Acero node on top of `input_node` and returns it.
+///
+/// The node (a custom `arrow::acero::ExecNode` registered as `rhydb_select_k`) buffers the whole
+/// input stream, selects the `offset + limit` smallest rows (by `ordering`) with a heap-based
+/// `select_k`, and emits the `[offset, offset + limit)` window in sorted order.
+arrow::Result<arrow::acero::ExecNode*> addSelectKNode(
```
fengelniederhammer, line 16: Should we maybe mention here that we can get rid of this once the bug in arrow is fixed?

src/rhydb/query_engine/exec_node/select_k.cpp
```diff
@@ -0,0 +1,236 @@
...
+      ARROW_ASSIGN_OR_RAISE(auto table, arrow::Table::FromRecordBatches(output_schema_, batches));
+
+      arrow::compute::ExecContext* ctx = plan_->query_context()->exec_context();
+      std::shared_ptr<arrow::Table> selected;
+      // Manually check for num_rows == 0 until https://github.com/apache/arrow/issues/51210 is
+      // fixed
+      if (table->num_rows() == 0) {
+         selected = std::move(table);
+      } else {
+         const arrow::compute::SelectKOptions options{k_, ordering_.sort_keys()};
```
fengelniederhammer, line 151: AI: 🟡 risk: this drops the `k > 0` validation that Arrow's `select_k_sink` enforced (`ValidateSelectKOptions`: `` `k` must be > 0 ``). The `SelectKRewritePass` fires whenever `count.has_value()` (`select_k_rewrite_pass.cpp:18`), so `.limit(0)` yields `k_ = offset + 0 = 0`, and `SelectKUnstable(table, {k: 0})` gets called on a non-empty table — the `num_rows() == 0` guard above doesn't cover it. Old path returned a clean `Status::Invalid`; new path passes `k=0` straight to Arrow. Suggest guarding `k_ == 0` (emit empty window) or validating in `make()`.

fengelniederhammer (approved):
Looks good, but let's add some comments to clarify where that code came from and why.

src/rhydb/query_engine/exec_node/select_k.h
taepper: I am not sure whether we want to get rid of it. Maybe we will at some point only use _arrow compute_, but the query-plan structure might move to our code base at some point. See

> Acero is experimental and a stable API is not yet guaranteed.
https://arrow.apache.org/docs/cpp/acero.html

src/rhydb/query_engine/exec_node/select_k.cpp
taepper: I don't think this holds. We are correctly executing in the `k=0` case

Issue #1114, opened 2026-01-20
Reference sequences can contain ambiguous symbols


Taepper:
For now, we had the assumption that reference sequences will not contain ambiguous symbols, most importantly not the symbol `N` for nucleotides or the symbol `X` for amino acids, which signifies that the symbol is "unknown".

We should check that SILO handles these cases correctly:
- Filtering for a mutation at that position
- These positions correctly appear in mutation distributions
- adapting local reference sequences for silo partitions behaves correctly in the case where the local reference initially had the "unknown" symbol

PR #1556, opened 2026-09-14
feat(query_engine): implement more efficient bitmap-aggregation

### LINKED ISSUE #1404: Allow querying of >100 positions for co-occurances

taepper:
resolves #1404

### Summary

This PR provides a more efficient implementation for computing group-by expressions where the group-by-fields have bitmaps available. These bitmaps can be used to compute group combinations efficiently without needing to materializing the input.

Importantly, this PR changes the way this speed-up is implemented:

First, it makes these efficient computations per-chunk: For each chunk the bitmaps are leveraged to efficiently compute all groups that do appear in the data and providing the count for each group.

Second, the individual (per-chunk) computations have been sped up, by avoiding copies of roaring containers when possible and providing a simd-accelerated scan of the row-wise coverage data.

## PR Checklist


src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
+                  except_reference_and_missing |= view;
+               }
+            }
+         }
+      }
+      return ChunkMutationContainers{
+         .views = std::move(mutation_views),
+         .except_missing = std::move(except_missing),
+         .except_reference_and_missing = std::move(except_reference_and_missing)
+      };
```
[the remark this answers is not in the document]
taepper: Only linter has authority.

src/rhydb/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
```diff
@@ -140,22 +141,111 @@ std::optional<operators::GroupingDimension> matchSequencePositionDimension(
...
+      return std::nullopt;  // a plain scan column, not a map-computed expression
+   }
+   const auto& expression = *assignment->get().expression;
+   if (expression.freeIUs().empty() || !isGroupableScalarType(expression.type()) ||
+       !isEvaluableGroupableExpression(expression, *source.scan.table->schema)) {
+      return std::nullopt;
+   }
+   return operators::ScalarExpressionDimension{
+      expression.clone(), expression.type(), group_by_field.name
+   };
```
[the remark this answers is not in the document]
taepper: Not part of lintingOnly linter has authority.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
+      }
+      roaring::internal::container_free(intersection, result_typecode);
    }
 }
 
-/// Recursively partitions `filter_bitmap` by the per-dimension group bitmaps, pruning empty
-/// branches, and records one combination of values with its row count per surviving leaf. Only
-/// non-empty combinations are visited (their number is bounded by the count of matching rows), so
-/// this scales to many dimensions without the exponential blow-up of a full Cartesian product.
+/// Computes the co-occurrence counts one 2^16 chunk at a time. It enumerates the filter's
```
fengelniederhammer, line 720: is "co-occurence" really the correct word here? The issue is about them, but this implementation seems about aggregation in general.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
+   std::vector<size_t>& chosen_indices,
+   CombinationCounts& counts
 ) {
-   if (depth == group_bitmaps_per_dimension.size()) {
-      combinations.push_back(
-         GroupCombination{.group_indices = accumulated_indices, .count = current.cardinality()}
-      );
+   const size_t last_dimension = groups_by_dimension.size() - 1;
+
+   if (const size_t* whole_chunk_label = std::get_if<size_t>(groups_by_dimension[depth])) {
```
fengelniederhammer, line 673:
```suggestion
   if (const size_t* whole_chunk_label = std::get_if<SingletonKeyGroup>(groups_by_dimension[depth])) {
```

fengelniederhammer, line 673: Why is this called `whole_chunk_label`?

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
-/// column per dimension (holding that dimension's group value, or null) plus the int64 count
-/// column.
+/// Materializes the combinations for `combinations[begin, end)` into a single ExecBatch: one column
+/// per dimension (holding that dimension's group value, or null) plus the int64 count column. Each
+/// dimension's output column is gathered by `Take`-ing its value array (`values_per_dimension[i]`,
+/// element = group value) at the combinations' group indices, so the output column has the value
+/// array's type -- utf8 for the string dimensions, int64 for a numeric column, etc.
 // The cognitive-complexity count comes entirely from the ARROW_RETURN_NOT_OK/ARROW_ASSIGN_OR_RAISE
-// error-check macros, not from real branching; the logic is a straight-line append loop.
+// error-check macros, not from real branching; the logic is a straight-line gather loop.
```
fengelniederhammer, line 781: We can probably shorten this comment?

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
 // NOLINTNEXTLINE(readability-function-cognitive-complexity)
```
fengelniederhammer, line 782: Do we still need this? The new implementation has quite a bit less indentation.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
+/// column's vertical mutation index and horizontal coverage index. Plain mutation groups are handed
+/// out as zero-copy views into the vertical index's stored containers; only the reference, missing
+/// and null groups -- which are defined by set arithmetic against the coverage and the filter --
+/// are computed per chunk. The reference / missing (no coverage) / null handling mirrors
+/// `SymbolInSet` exactly so the grouping matches the generic `at()`/groupBy path:
+///   * a plain mutation symbol -> the rows carrying that mutation at the position,
+///   * the local reference symbol -> the covered rows carrying no other mutation,
+///   * the missing symbol -> the not-covered rows (plus any explicit missing mutation), minus
+///   nulls,
+///   * a null sequence -> its own null group (it carries no symbol at any position).
```
fengelniederhammer, line 119: Maybe we should soften this comment a bit? "mirrors `SymbolInSet` exactly" sounds quite strong, and it's not a 1-to-1 copy of the code there, is it?

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
+         data =
+            buildScalarGroups<BoolValueTraits>(table, referenced, bound_expression, output_type);
+         break;
+      case schema::ColumnType::DATE32:
+         data =
+            buildScalarGroups<Date32ValueTraits>(table, referenced, bound_expression, output_type);
+         break;
+      default:
+         // The rewrite pass only routes groupable scalar output types here; anything else is a bug.
+         throw rhydb::query_engine::IllegalQueryException(
```
fengelniederhammer, line 647:
```suggestion
         throw IllegalQueryException(
```

src/rhydb/storage/column/horizontal_coverage_index.cpp
```diff
@@ -49,6 +57,197 @@ void HorizontalCoverageIndex::insertCoverage(RowId row_id, const Coverage& cover
...
+      }
+   }
+
+   // The scan builds a dense bitset container; compact it (to a run/array) so the downstream set
+   // algebra on the covered set stays cheap.
+   result.runOptimize();
+
+   // Every row added above lies in `chunk_id`, so `result` holds at most this one 2^16 container.
+   // Steal it out of the roaring array (no clone) and hand it back on its own; an empty result
+   // becomes an empty container.
```
fengelniederhammer, line 199: Reads clunky IMO
```suggestion
   // This is constructed such that `result` holds at most this one 2^16 container.
   // Steal it out of the roaring array (no clone) and hand it back on its own.
```

src/rhydb/storage/column/horizontal_coverage_index.cpp
```diff
@@ -49,6 +57,197 @@ void HorizontalCoverageIndex::insertCoverage(RowId row_id, const Coverage& cover
...
+   uint32_t position,
+   uint16_t chunk_id
+) const {
+   roaring::Roaring result;
+   if (chunk_id >= starts.size()) {
+      return roaring_util::RoaringContainer::withCapacity(1);
+   }
+
+   // No row in this chunk can cover the position.
+   if (batch_max_end.at(chunk_id) <= position || batch_min_start.at(chunk_id) > position) {
```
fengelniederhammer, line 148: Let's use `noRowCoversPositionInChunk`?

src/rhydb/storage/column/horizontal_coverage_index.cpp
```diff
@@ -49,6 +57,197 @@ void HorizontalCoverageIndex::insertCoverage(RowId row_id, const Coverage& cover
...
+   }
+
+   const uint32_t base_row_id = static_cast<uint32_t>(chunk_id) << 16U;
+   const auto& chunk_starts = starts[chunk_id];
+   const auto& chunk_ends = ends[chunk_id];
+
+   // Fast path: if the position lies within the chunk's intersection envelope
+   // `[batch_max_start, batch_min_end)`, every row in the chunk covers it, so add the whole chunk
+   // in one range operation.
+   if (batch_max_start.at(chunk_id) <= position && position < batch_min_end.at(chunk_id)) {
```
fengelniederhammer, line 159: Isn't this essentially `positionCoveredByWholeChunk`?

src/rhydb/query_engine/optimizer/bitmap_aggregation_rewrite_pass.cpp
```diff
@@ -168,6 +258,9 @@ std::optional<operators::GroupingDimension> resolveDimension(
    if (auto dimension = matchIndexedColumnDimension(source, group_by_field)) {
       return dimension;
    }
+   if (auto dimension = matchScalarExpressionDimension(source, group_by_field)) {
+      return dimension;
+   }
```
fengelniederhammer, line 263: Does this PR mangle two features into one PR? Extending aggregation by grouping by scalar expressions and the general optimization? It would have probably been easier to review them separately.

src/rhydb/storage/column/horizontal_coverage_index.cpp
```diff
@@ -49,6 +57,197 @@ void HorizontalCoverageIndex::insertCoverage(RowId row_id, const Coverage& cover
...
+      cardinality += static_cast<uint32_t>(__builtin_popcount(covered));
+   }
+   for (; row < num_rows; ++row) {
+      if (starts[row] <= position && position < ends[row]) {
+         bitset[row >> 6U] |= (uint64_t{1} << (row & 63U));
+         ++cardinality;
+      }
+   }
+   return cardinality;
+}
```
fengelniederhammer, line 109: AI: The hand-written AVX-512 `coverageScan` has no test asserting it produces the same result as `coverageScanScalar`. Since it only runs on AVX-512-capable CPUs (gated by `cpuHasAvx512()`), CI on other hardware never exercises it, so a mask/shift bug could ship undetected. Suggest a unit test that runs both implementations on randomized `starts`/`ends`/`position` inputs and asserts equal bitsets + cardinality (guarded by `cpuHasAvx512()` so it's a no-op where unsupported).

src/rhydb/storage/column/horizontal_coverage_index.cpp
```diff
@@ -49,6 +57,197 @@ void HorizontalCoverageIndex::insertCoverage(RowId row_id, const Coverage& cover
...
+   RHYDB_ASSERT_EQ(roaring_array.size, 1);
+   const auto cardinality = static_cast<uint32_t>(roaring::internal::container_get_cardinality(
+      roaring_array.containers[0], roaring_array.typecodes[0]
+   ));
+   roaring_util::RoaringContainer container{
+      roaring_array.containers[0], cardinality, roaring_array.typecodes[0]
+   };
+   roaring::internal::ra_clear_without_containers(&roaring_array);
+   return container;
+}
```
fengelniederhammer, line 213: AI: The new public methods `coveredRowsInChunk`, `noRowCoversPositionInChunk` and `positionCoveredByWholeChunk` have no direct unit tests — they're only covered indirectly through query tests. The logic here is subtle (envelope fast paths, in-region-N removal, and the `ra_clear_without_containers` container-steal), so a bug would surface only as an opaque wrong query count. Suggest focused `HorizontalCoverageIndex` unit tests covering: out-of-range chunk_id, whole-chunk-covered fast path, partial coverage (SIMD/scalar scan), in-region N removal, and empty result.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,821 @@
...
+            buildScalarGroups<DoubleValueTraits>(table, referenced, bound_expression, output_type);
+         break;
+      case schema::ColumnType::BOOL:
+         data =
+            buildScalarGroups<BoolValueTraits>(table, referenced, bound_expression, output_type);
+         break;
+      case schema::ColumnType::DATE32:
+         data =
+            buildScalarGroups<Date32ValueTraits>(table, referenced, bound_expression, output_type);
+         break;
```
fengelniederhammer, line 644: AI: Only the STRING and DATE32 scalar-expression output types are exercised by tests. The INT32 / INT64 / FLOAT / BOOL `*ValueTraits` paths (key extraction, bucketing, value-array build) are uncovered. A query test grouping on a numeric/bool map expression for each would guard these traits instantiations.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,823 @@
...
+      );
+   }
+
    std::vector<GroupCombination> combinations;
-   std::vector<size_t> accumulated_indices;
-   partition(filter_bitmap, 0, group_bitmaps_per_dimension, accumulated_indices, combinations);
+   combinations.reserve(counts.size());
+   for (const auto& [indices, count] : counts) {
+      combinations.push_back(GroupCombination{.group_indices = indices, .count = count});
+   }
```
fengelniederhammer, line 778: AI: nit (peak memory): these key vectors are copied out of `counts` while the map is still alive, so peak memory briefly holds two copies of every group-index tuple. For the #1404 case (100+ positions → up to ~1M distinct combinations, each key `num_dimensions * sizeof(size_t)` bytes) that doubles the largest allocation. Consider moving keys out (`counts.extract(it)`) instead of copying, and/or narrower index types (`uint16_t`/`uint32_t` — symbol and value counts fit easily).

src/rhydb/storage/column/horizontal_coverage_index.cpp
```diff
@@ -49,6 +57,205 @@ void HorizontalCoverageIndex::insertCoverage(RowId row_id, const Coverage& cover
...
+   const uint64_t chunk_end_key = static_cast<uint64_t>(base_row_id) + chunk_starts.size();
+   const auto chunk_rows_end =
+      chunk_end_key > UINT32_MAX
+         ? horizontal_bitmaps.end()
+         : horizontal_bitmaps.lower_bound(static_cast<uint32_t>(chunk_end_key));
+   for (auto iter = chunk_rows_begin; iter != chunk_rows_end; ++iter) {
+      if (iter->second.contains(position)) {
+         result.remove(iter->first);
+      }
+   }
```
fengelniederhammer, line 200: AI: nit (perf): this in-region-N removal loop runs unconditionally, but after the `positionCoveredByWholeChunk` fast path it can never remove anything — that function already scanned the same `horizontal_bitmaps` range for an N at `position` and would have returned false if it found one. So for a fully-covered chunk this is a second full map scan per (position, chunk) that does nothing. Move the loop into the `else` (scan) branch. The `lower_bound(base_row_id)` / `chunk_end_key` range computation is also duplicated with `positionCoveredByWholeChunk` (L244-250) — a small helper would remove the copy.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,823 @@
...
+   )
+       : column(column),
+         position_idx(position_idx) {
+      CHECK_RHYDB_QUERY(
+         position_idx < column.metadata->reference_sequence.size(),
+         "SymbolInSet<{}> position is out of bounds {} > {}",
+         SymbolType::SYMBOL_NAME,
+         position_idx + 1,
+         column.metadata->reference_sequence.size()
+      );
```
fengelniederhammer, line 141: AI: nit: this user-facing message says `SymbolInSet<...> position is out of bounds`, but this path no longer goes through `SymbolInSet`. Reword to refer to the sequence position / `at()` so it matches what the user actually wrote. Note `CO_OCCURRENCE_VIA_MAP_POSITION_OUT_OF_RANGE` pins the old text and would need updating too.

src/rhydb/query_engine/operators/bitmap_aggregation_node.h
```diff
@@ -74,9 +73,11 @@ struct IndexedColumnDimension {
 /// One grouping dimension of a `BitmapAggregationNode`: a rule for partitioning a filtered row-set
 /// into disjoint, value-keyed groups directly from roaring bitmaps, plus the STRING output column
 /// it contributes. A variant over the supported kinds lets a single query group on a mix of them;
-/// add an alternative to support another kind. Every alternative offers `buildGroups`,
-/// `outputColumn` and `toJson`, so a generic `std::visit` dispatches over them.
-using GroupingDimension = std::variant<SequencePositionDimension, IndexedColumnDimension>;
+/// add an alternative to support another kind. Every alternative offers `outputColumn` and
+/// `toJson` (so a generic `std::visit` dispatches over them) and a `makeGrouper` overload in the
+/// implementation file that resolves it against the table into a per-chunk grouping strategy.
```
fengelniederhammer, line 78: AI: nit: this doc still says the dimension contributes "the STRING output column", but `ScalarExpressionDimension` now emits its own typed column (INT32/INT64/FLOAT/BOOL/DATE32/STRING). Drop the STRING claim.

src/rhydb/query_engine/operators/bitmap_aggregation_node.test.cpp
```diff
@@ -297,6 +391,132 @@ const QueryTestScenario CO_OCCURRENCE_AMBIGUOUS_CODES = {
...
+   ])")
+};
+
+// ---------------------------------------------------------------------------
+// One grouping key per remaining groupable scalar type
+//
+// The bitmap aggregation node instantiates a separate ValueTraits per output type -- key
+// extraction, bucketing and the typed value array. Only the string and date ones were reached by a
+// test; these cover int32, int64, float and bool. A map field reference over a non-indexed column
+// is the only shape that gets a numeric or boolean expression to that path (`at` and `isoWeek`
```
fengelniederhammer, line 421: AI: nit + test gap: this says only the string and date traits were reached by a test, but `isoWeek()` yields a STRING, so `Date32ValueTraits` (DATE32 output) is still never instantiated by any test. Add a scenario grouping directly on a date column, e.g. `default.map({d := date}).groupBy({count:=count()}, {d})`, and fix the comment.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,823 @@
...
+
+         CopyOnWriteContainer group;
+
+         if (is_reference && is_missing) {
+            // The local reference symbol is itself the missing symbol: every filtered row without
+            // an explicit other mutation carries it, minus the sequence-less rows.
+            group = CopyOnWriteContainer{filter_view - mutations.except_missing.view()};
+            if (null_view.has_value()) {
+               group -= *null_view;
+            }
```
fengelniederhammer, line 249: AI: test gap: the branch where the local reference symbol *is* the missing symbol (`is_reference && is_missing`) is not exercised. The `segment1` reference is `ATGCN`, so position 5 has reference N — a scenario grouping on `segment1.at(5)` would cover this branch (and its null / `except_missing` subtraction), which otherwise differs only subtly from the plain missing and reference paths.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,823 @@
...
    const Bitmap& filter_bitmap
 ) {
-   GroupBitmaps result;
-   for (const auto symbol : SymbolType::SYMBOLS) {
-      auto compiled = scalar_expressions::compileSymbolInSet<SymbolType>(
-         column, position_idx, std::vector<typename SymbolType::Symbol>{symbol}, row_layout
+   arrow::compute::ExecContext exec_context;
+   std::map<typename Traits::KeyType, roaring::Roaring> bitmap_by_value;
+   roaring::Roaring null_bitmap;
+   for (const auto& [chunk_id, filter_view] : filter_bitmap) {
```
fengelniederhammer, line 540: AI: test gap: scalar-expression grouping is never tested together with a filter. This loop deliberately evaluates only the filtered rows; a scenario like `default.filter(...).map({c := country}).groupBy({count:=count()}, {c})` would confirm excluded rows don't leak into groups and that counts stay filter-bounded.

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
```diff
@@ -1,175 +1,823 @@
...
+      for (size_t group_index = 0; group_index < values.size(); ++group_index) {
+         for (const auto& [chunk_id, view] : Bitmap{values[group_index].second}) {
+            groups_by_chunk[chunk_id].emplace_back(group_index, CopyOnWriteContainer{view});
+         }
+         key_values.emplace_back(std::move(values[group_index].first));
+      }
+      const size_t null_index = values.size();
+      for (const auto& [chunk_id, view] : Bitmap{&column.null_bitmap}) {
+         groups_by_chunk[chunk_id].emplace_back(null_index, CopyOnWriteContainer{view});
+      }
```
fengelniederhammer, line 333: AI: test gap: the null group of an indexed column isn't covered — `region` is always present in the test data, so this trailing null-index group is never populated or emitted. A row with a null indexed value would pin that the null group gets the right index and count.

fengelniederhammer (commented):
Some of these might be slop and some not worth fixing

src/rhydb/query_engine/operators/bitmap_aggregation_node.cpp
taepper: To do in #1114

fengelniederhammer (approved):

PR #1562, opened 2026-09-14
test(performance): add a benchmark for lineage() filter

### LINKED ISSUE #1558: Create benchmark for `lineage` filter
Before removing the `lineage_index`, we should implement a benchmark for the lineage filter to detect possible performance regressions

taepper:
resolves #1558

### Summary

This adds a benchmark for `lineage()` filters, to track possible regressions

## PR Checklist


performance/lineage_filter_benchmark.cpp
```diff
@@ -0,0 +1,226 @@
...
+   const std::map<std::filesystem::path, LineageTreeAndIdMap> lineage_trees{
+      {LINEAGE_TREE_NAME, lineage_tree}
+   };
+   const std::vector<std::string> names = loadLineageNames();
+   SPDLOG_INFO("Loaded {} lineages from {}", names.size(), LINEAGE_DEFINITION_PATH.string());
+
+   // Target lineages of varying descendant-set size: a near-root high-fanout lineage, a mid-level
+   // one, and a leaf. Sorting makes the pick deterministic across runs.
+   std::vector<std::string> sorted_names = names;
+   std::ranges::sort(sorted_names);
```
fengelniederhammer: Why does sorting guarantee here that we get a leaf if we take the last element?

performance/lineage_filter_benchmark.cpp
```diff
@@ -0,0 +1,226 @@
...
+   const std::vector<std::pair<std::string, std::string>> targets{
+      {"near-root", sorted_names.front()},
+      {"mid-level", sorted_names.at(sorted_names.size() / 2)},
+      {"leaf", sorted_names.back()},
+   };
```
fengelniederhammer: AI: The comment claims these targets vary by "descendant-set size" (near-root/mid-level/leaf), but the selection is a plain alphabetical `sort` picking front/middle/back. Alphabetical order doesn't correlate with tree depth or fanout, so the labels are misleading and the benchmark doesn't deterministically exercise varying subtree sizes (only `front()`=="A"==root happens to match by coincidence). Consider sorting by actual descendant count computed from `lineage_tree.getChildToParentRelation()`, or drop the size claim from the labels.

performance/lineage_filter_benchmark.cpp
```diff
@@ -0,0 +1,226 @@
...
+      const auto build_ms =
+         std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::high_resolution_clock::now() - start_build
+         )
+            .count();
+
+      SPDLOG_INFO("build: {} ms", build_ms);
+
+      for (const auto& [target_label, target] : targets) {
+         for (const std::string_view mode : {"doNotFollow", "alwaysFollow"}) {
```
fengelniederhammer: AI: Only `doNotFollow` and `alwaysFollow` are benchmarked. Is omitting `followIfFullyContainedInClade` intentional? It's the more computationally interesting recombinant case and may be worth tracking for regressions too.

fengelniederhammer (commented):
Could you please rebase / merge main into this PR? https://github.com/GenSpectrum/LAPIS-SILO/pull/1551 caused conflicts.

fengelniederhammer (approved):

PR #1563, opened 2026-09-14
feat(query_engine): add startingFrom to transitiveClosure()

### LINKED ISSUE #1561: Allow `startingSet` in `transitiveClosure`
Currently, transitive-closure computes the closure for the entire edge relation's `from` set. Now, we add the option to start building the relation set from a given set of vertices, which can save a lot of computation time

taepper:
resolves #1561

### Summary

Restricts the closure to the given source vertices: only pairs whose `from` is one of them are emitted, and only the part of the graph those vertices reach is walked. Asking for the descendants of one node then costs one search instead of one per vertex in the relation.

Each per-source search now also costs what it reaches rather than the size of the graph: a generation-stamped visited array replaces clearing the bitmap per source, and pairs are buffered as they are discovered instead of scanning every vertex afterwards.

## PR Checklist


documentation/query_documentation.md
```diff
@@ -541,9 +541,25 @@ Here `lineage_column` is a `STRING` column of `default` holding each sequence's
...
+
+```
+pango_lineage.transitiveClosure('parent', 'lineage', includeVertices:=true, startingFrom:={'B.1.1.7'})
+  .join(default, to = lineage_column)
+  .groupBy({count := count()}, {from})
+```
+
+A vertex named in `startingFrom` that does not occur in the relation contributes no rows, not even
+its reflexive pair. Since `filter(...)` cannot be applied to the closure's output, this is also the
+way to narrow the result by source vertex.
```
fengelniederhammer: A filter can now be applied there.

documentation/query_documentation.md
```diff
@@ -541,9 +541,25 @@ Here `lineage_column` is a `STRING` column of `default` holding each sequence's
 the joined-against column must itself be `STRING` (a lineage column configured with an index is
 dictionary-encoded and cannot be used as the join key directly).
 
+**Starting from selected vertices.** `startingFrom` restricts the closure to the given source
+vertices: only pairs whose `from` is one of them are emitted, and only the part of the graph those
+vertices reach is ever walked. Asking for the descendants of one lineage therefore costs one search
+instead of one per vertex in the relation:
```
fengelniederhammer, line 566: Let's maybe add when a user should use it? It reads like a performance optimization for queries, but it's not really clear for which cases.

src/rhydb/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1556,9 +1556,22 @@ operators::QueryNodePtr handleTransitiveClosure(
...
    }
+   std::optional<std::vector<std::string>> starting_from;
+   if (const auto* expr = args.get("startingFrom")) {
+      const auto& set = extractSetLiteral(*expr);
+      starting_from.emplace();
+      starting_from->reserve(set.elements.size());
+      for (const auto& element : set.elements) {
+         starting_from->push_back(extractStringLiteral(*element));
+      }
+   }
```
fengelniederhammer, line 1576: AI: Duplicate vertices in `startingFrom` produce duplicate output rows. Confirmed on a live instance: `startingFrom:={'A.2','A.2'}` emits every pair twice. `handleIn` (L322) dedupes set elements via `std::unordered_set`; here the elements go into a plain vector, so each duplicate becomes a separate source and re-emits its whole reachable set. Suggest deduplicating while resolving (e.g. skip names already seen, or dedup `sources` in `buildRelation`).

src/rhydb/query_engine/saneql/ast_to_query.cpp
```diff
@@ -1556,9 +1556,22 @@ operators::QueryNodePtr handleTransitiveClosure(
    if (const auto* expr = args.get("includeVertices")) {
       include_vertices = extractBoolLiteral(*expr);
    }
+   std::optional<std::vector<std::string>> starting_from;
+   if (const auto* expr = args.get("startingFrom")) {
```
fengelniederhammer, line 1569: AI: An empty set `startingFrom:={}` silently yields an empty result (confirmed 200, no rows). Is that the intended contract, or should an empty starting set be rejected as a likely user error? Worth a test either way.

src/rhydb/query_engine/saneql/ast_to_query.cpp
taepper: That is expected? We have multi-set semantics

src/rhydb/query_engine/saneql/ast_to_query.cpp
taepper: This is the intended contract

Issue #1541, opened 2026-09-06
Support non-string values in `in()`


chaoran-chen:
If I'm not mistaken, `.in()` only supports string values at the moment and I think that it would make sense to support other value types as well.

taepper:
Should be resolved by #1565, but will be confirmed and covered by tests separately

PR #1565, opened 2026-09-14
feat(query_engine): resolve lineage() from the lineage relation table

### LINKED ISSUE #1559: Also compute `lineage` filter from lineage table
Currently, the `lineage` filter is only available when a `lineage_index` has been created for the given column. We should also add the possibility to compute the `lineage` filter if we instead only store the lineage table

taepper:
resolves #1559

### Summary

This allows the computation of `lineage()` filters on lineage relation tables (no `lineage_index` on the column). This uses an `.in(<subquery>)` predicate, where the subquery computes the sublineages using the `transitiveClosure` function with the queried lineage as the starting set.

## PR Checklist


taepper:
Also resolves #1541, but will be merged and covered by tests separately

PR #1570, opened 2026-09-21
refactor(storage): take row sets as rhydb::Bitmap, not roaring::Roaring

taepper:
### Summary

`overwriteSymbolsInSequences` and `overwriteCoverageInSequence` now take a
`Bitmap` and walk it block by block. This both avoid data copies in some cases due to the internal per-container abstraction,
but also exposes a cleaner API that can be reused more easily if we have no
`roaring::Roaring` available

## PR Checklist


fengelniederhammer (approved):
LGTM

